PR #23, opened 2022-04-14
Add warp.torch to __init__

mmarcinkiewicz:
Allows to run

`import warp as wp`
`wp.torch.xx`

instead of

`import warp as wp`
`import warp.torch`
`wp.torch.xx`

### DIFF
--- a/warp/__init__.py
+++ b/warp/__init__.py
@@ -11,3 +11,4 @@ from warp.context import *
 from warp.builtins import *
 from warp.tape import *
 from warp.utils import *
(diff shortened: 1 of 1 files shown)


mmacklin:
Thanks @mmarcinkiewicz , the reason we don't currently import torch by default is to avoid putting a hard dependency on Torch in the lib. (since many users don't need it).

Perhaps we could wrap the `import torch` with a `try/except` block? Would that work?

mmacklin:
Merged in latest.

PR #25, opened 2022-05-04
fix the loading-dll-error under Windows with Python 3.8

### LINKED ISSUE #24: FileNotFoundError: Could not find module 'warp.dll'
After building from source, I encounter the following error when running tests. 

```
FileNotFoundError: Could not find module 'warp.dll' (or one of its dependencies). Try using the full path with constructor syntax.
```

After spending some time on the problem, I figured out that this is due to a bug of Python 3.8 on Windows. 

This stackoverflow post pointed out the issue.
https://stackoverflow.com/questions/59330863/cant-import-dll-module-in-python/64472088#64472088

The author of the above post has reported this bug and it has been fixed for future python versions. 
https://github.com/python/cpython/issues/86280

In short, the solution is to explicitly pass `winmode=0` to `ctypes.CDLL`. In Python 3.8, the default value of `winmode` is set to `None`, which is unintended as it is not consistent with the documentation. This argument only affects the loading logic under Windows. 

Besides this change, I also need to pass the full path of the file `warp.dll` instead of just the filename in order to successfully load the file. This is suggested in the error message above.

I'll be creating a pull request to fix this issue soon. I think python 3.8 users on Windows will encounter this problem and will benefit from the fix. 

------
**My environment**
- OS: Windows 10
- Python version: 3.8.10
- MSVC Compiler Version (cl.exe version) 19.26.28806 for x64
- Warp version: 0.1.25

DesmondZhong:
This pull request fixes #24 and makes dll loading working correctly under Windows with python 3.8.

1. By directly passing `winmode=0` to `ctypes.CDLL`, we achieve the correct default behavior. See the following posts for more details. 

> This stackoverflow post pointed out the issue.
https://stackoverflow.com/questions/59330863/cant-import-dll-module-in-python/64472088#64472088
> 
> The author of the above post has reported this bug and it has been fixed for future python versions.
https://github.com/python/cpython/issues/86280

2. The full path of `warp.dll` instead of the file name is used (L459 in `context.py`). This is now consistent with the logic under Mac (L464) and Linux (L469). Note that passing only the file name will still raise the `FileNotFound` error in my environment, so this change is necessary for me. I haven't tested it on Windows with Python 3.7 or lower yet, but I'm happy to do it if you are interested in merging this pull request!

----
**My environment**
- OS: Windows 10
- Python version: 3.8.10
- MSVC Compiler Version (cl.exe version) 19.26.28806 for x64
- Warp version: 0.1.25


shi-eric:
Looks like related changes were merged in a810b842fe82d3e0aa99871ca4f75df9ba892b4e a long time ago. Thanks for the suggestions!

PR #59, opened 2022-08-27
Marginally improve pytorch interface

half-potato:
I want to use from_pytorch to allow pytorch to handle memory allocation but I don't want to allocate unecessary memory for the gradient. This also handles the two dtypes allowed logically.

### DIFF
--- a/warp/torch.py
+++ b/warp/torch.py
@@ -11,13 +11,15 @@ import numpy
 
 
 # wrap a torch tensor to a wp array, data is not copied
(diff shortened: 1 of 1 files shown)


mmacklin:
Hi @half-potato, we recently addressed this issue in `0.8.0` so that you can override the gradient allocation in `wp.from_torch()` by setting `requires_grad=False`.

Hope that addresses your problem - if not feel free to open an issue.

Best,
Miles

PR #66, opened 2022-10-26
Fix the error when there are variables in a struct with .grad=None

PingchuanMa:
When a variable in a struct has .grad=None, it complains about the NoneType when assigning it to the struct. Bug fixed by replacing the grad with a same-dtype same-device same-shape zero array.


shi-eric:
Hey @PingchuanMa, is this still a bug? The code base has changed so much that GitHub is confused about the changes in the pull request. Please reopen a new pull request if you're interested in getting this fixed still, thanks.

PR #86, opened 2023-04-05
Add abs and sign functions for vec types and fix neg operator

kmarchais:
## Description
This pull request proposes the following changes that I personally needed:
- Implementation of the `abs` and `sign` functions for `vec` types
- Fix for a bug with the `neg` operator

## Changes Made
- `vec.h` &rarr; Added abs and sign functions for `vec` types
- `builtins.py` &rarr; Added built-in `abs` and `sign` functions for `vec` types
- `types.py` &rarr; Corrected the `neg` operator, which was expecting an extra `y` parameter.  

## Testing
I didn't change any test because I don't understand the following issues
- For the `abs` and `sign` functions, tests will probably be similar to those for `min` and `max`. However, `min` and `max` tests are not run due to slow compile time.
- While the test for the `neg` operator appeared to pass, an error was encountered when the operator was used.


c0d1f1ed:
Thanks for this proposal and the accompanying code changes!

We're going to need a different license and have contributors sign an agreement, before we can merge public MRs. We're working towards that but it might take a while.

Also, the proposed functionality in this MR all sounds very useful, but we have to balance it against the run-time compilation impact of larger headers. I'm currently working on changes that should give us more headroom for that (using pre-compiled headers), and we'll put benchmarks in place to ensure the tradeoffs are well balanced.

Thanks for your patience. Note that in the meantime you could use `@func` to write your own helper functions that implement this functionality, instead of making them part of the `wp.*` builtins.

christophercrouzet:
Hi @kmarchais, this is just to let you know that the `abs()` and `sign()` built-ins for vectors have been merged into the `main` branch, and the issue with the `neg` operator was fixed a while ago already. Thanks for your pull request!

PR #87, opened 2023-04-11
Update annotation checks to make it work with 'self' keyword

kmarchais:
## Description
With 0.8, it seems that it is not possible to define kernel functions that are members of a class because of the annotation checks to make sure that the number of arguments matches the number of annotations.
I propose this change to be able to use the `self` keyword for kernel functions except if this is not the expected behaviour.

## Changes
- `codegen.py` &rarr; do not raise the error if the number of annotations is equal to the number of arguments - 1 and the first argument is `self`

### DIFF
--- a/warp/codegen.py
+++ b/warp/codegen.py
@@ -312,7 +312,11 @@ class Adjoint:
         if overload_annotations is None:
             # use source-level argument annotations
             if len(argspec.annotations) < len(argspec.args):
-                raise RuntimeError(f"Incomplete argument annotations on function {adj.fun_name}")
+                if not (
+                    len(argspec.annotations) == len(argspec.args) - 1
+                    and argspec.args[0] == "self"
+                ):
+                    raise RuntimeError(f"Incomplete argument annotations on function {adj.fun_name}")
             adj.arg_types = argspec.annotations
         else:
             # use overload argument annotations


mmacklin:
Thank you @kmarchais! Let me ask @nvlukasz  to review this and make the change internally if he agrees.

Best,
Miles

nvlukasz:
Nice catch @kmarchais, I will take a look.  Thanks!

nvlukasz:
Hi @kmarchais, it is possible to declare kernels as methods in a class, but you can omit the `self` parameter.  A Warp kernel cannot take a Python object as an argument, so a class instance passed as `self` could not be used in the kernel code.

This works fine currently:
```python
class Foo:

    @wp.kernel
    def hello(x: int):
        print(x)

# launch with instance
foo = Foo
wp.launch(foo.hello, dim=1, inputs=[42])

# launch with class
wp.launch(Foo.hello, dim=1, inputs=[42])
```

We _could_ allow passing an undecorated first argument to kernels declared in a class, but they still could not be used in the kernel body.  A Warp kernel declared in a class is not really an instance method - it's more like a static method.

Possibly I'm missing something?  Could you share a bit more info about your use case?

kmarchais:
My bad! I confused, I'm too used to use the self keyword, indeed it is not necessary in this case. Thank you!

Issue #90, opened 2023-04-23
fileno


mingno:
### in Maya

Error: AttributeError: file C:\Users\mingn\AppData\Roaming\Python\Python37\site-packages\warp\__init__.py line 9: 'maya.Output' object has no attribute 'fileno'

### in Blender

Traceback (most recent call last):
  File "<blender_console>", line 1, in <module>
  File "C:\Program Files\Blender Foundation\Blender 3.5\3.5\python\lib\site-packages\warp\__init__.py", line 9, in <module>
    faulthandler.enable()
io.UnsupportedOperation: fileno

mmacklin:
Hmm, thanks @mingno - this could be due to the recent addition of the `faulthandler`. I will look to disable this or at least handle these exceptions.

Best,
Miles

c0d1f1ed:
Thanks @mingno for reporting this! Could you give https://github.com/NVIDIA/warp/pull/92 a try?

mmacklin:
We have disabled the `faulthandler` on init so I will close this for now.

PR #92, opened 2023-04-25
Conditionally restore faulthandler support

c0d1f1ed:
Some environments may override `sys.stderr` with a stream object which does not have the `fileno` attribute, which faulthandler relies on.

Issue: #90


c0d1f1ed:
Re-created as an internal MR.

PR #163, opened 2023-11-30
Support overriding the LLVM project source directory path

c0d1f1ed:
If `--build_llvm` is specified we previously cloned the LLVM project
Git repository from https://github.com/llvm/llvm-project.git into
`external/llvm-project`.

This change adds an additional `--llvm_source_path` build argument for
specifying a path to an existing LLVM project repo. Note that currently
it needs to be the `llvmorg-15.0.7` tag revision.


c0d1f1ed:
This was merged as 18aff4d89a94d48ae2ddc976bbd2341e28ca4bc7.

PR #195, opened 2024-03-19
[XPBD] Project tetrahedra constraints at once

jjcasmar:
According to "Parallel block Neo-Hookean XPBD using graph clustering", projecting tetrahedra constraints at once instead of first the deviatoric term and then the hidrostatic term may lead to a significant convergence improvement. 

This patch adds that.

I am not sure if you accept external contributions or if you are event interested in something like this, so the submitted code is not tested, and most probably doesn't really work. If you are interested, I can review it so it passes tests


shi-eric:
Closing this pull request as `warp.sim` has been removed from Warp, with work continuing in the Newton project. Sorry we weren't able to get this merged.

PR #221, opened 2024-05-22
Add extras for tests and the usd features to pyproject.toml

cadop:
Using extras is convenient in setup.py for users, but also allows for specific versioning requirements.


shi-eric:
Can you stick these into `[project.optional-dependencies]` of [pyproject.toml](https://github.com/NVIDIA/warp/blob/main/pyproject.toml) instead?

cadop:
I moved to `pyproject.toml`. I see a `dev` extra, should that be included in the readme as an optional pip install arg?

shi-eric:
> I moved to `pyproject.toml`. I see a `dev` extra, should that be included in the readme as an optional pip install arg?

Thanks! I think we will document the `[dev]` option separately when we more specifically document how other developers outside NVIDIA can contribute to Warp as we've only recently received clarification from legal that our existing license allows for such contributions.

I briefly discussed your PR with @mmacklin last week, and I think it is better if you just add everything under `extras`, e.g. 

```
extras = ['usd-core', 'matplotlib', 'pyglet']
```

It seems that having a `warp-lang[extras]` that only installs `usd-core` isn't much more convenient than just doing a `pip instal usd-core`, and the name `warp-lang[examples]` might be misleading as it is the example dependencies that are installed, not the examples themselves.

Anyway, I want to use your MR to develop a process for getting external contributions back into the Warp codebase, but it might take a few days. As you might have noticed we do our development on an internal repo and then push those changes to GitHub. Eventually we want to migrate development to be GitHub-first, but there are many things that have to happen first before we can do this.

So this PR likely won't be merged directly on GitHub when approved, but someone will probably grab your commits and merge them to our internal repo first, and then close the PR when everything is merged. You should still get credit for the contribution in the commit log. Just giving you a heads up about what's going on behind the scenes!

cadop:
Updated to just the extras arg. 

Thanks for the heads up on the gitlab mirror. Happy to help test future PR processes :)

shi-eric:
Hey @cadop, your PR is now in Warp! https://github.com/NVIDIA/warp/commits?author=cadop

There was a hiccup when doing the merge initially on GitLab. We normally have squash merge turned on, so when all your merges were squashed into a single commit, I became the author. I needed to revert the merge and then merge your changes again, this time turning the squash merge option off. This means that we will have to have the author clean up their PR commit history in the future rather than relying on the system to squash-merge. I've made a note of this.

Please close this PR if you are satisfied with the resolution. Thanks!

PR #306, opened 2024-09-03
3D Volume initialization from numpy

steinraf:
## Category

- [ ] New feature
- [X] Bugfix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please explain)

## Description

The current Volume.load_from_numpy method mistakenly initializes a 1D numpy array with values `(target_shape[0], target_shape[1], target_shape[2], 3)` instead of that shape.

## Changelog

- In case bg_value is multidimensional, correctly initialize an array filled by the bg_value
- Adding a test which is not passed by the original version
- Running ruff format 

- [X] Do you agree to the terms under which contributions are accepted as described in Section 9 the Warp [License](https://github.com/NVIDIA/warp/blob/main/LICENSE.md)?
- [X] Have you read the [Contributor Guidelines](https://github.com/NVIDIA/warp/blob/main/CONTRIBUTING.md)?
- [X] Have you written any new necessary tests?
- [ ] Have you added or updated any necessary documentation?
- [ ] Have you added any files modified by compiling Warp and building the documentation to this PR (.e.g. `stubs.py`, `functions.rst`)?
- [X] Does your code pass `ruff check` and `ruff format --check`?

### DIFF
--- a/warp/tests/test_volume.py
+++ b/warp/tests/test_volume.py
@@ -63,11 +63,7 @@ def test_volume_sample_linear_f(volume: wp.uint64, points: wp.array(dtype=wp.vec
         return  # not testing against background values
 
     expect_near(wp.volume_sample_f(volume, p, wp.Volume.LINEAR), expected, 2.0e-4)
-    expect_near(
-        wp.volume_sample(volume, p, wp.Volume.LINEAR, dtype=wp.float32),
-        expected,
-        2.0e-4,
-    )
+    expect_near(wp.volume_sample(volume, p, wp.Volume.LINEAR, dtype=wp.float32), expected, 2.0e-4)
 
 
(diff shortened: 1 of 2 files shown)


shi-eric:
Thanks @steinraf. It seems like your pull request makes a lot of unrelated formatting changes, perhaps from a code editor setting that adds trailing commas everywhere, or maybe a max-line-length setting. Can you please update your pull request so that it only has the changes relevant for your bugfix?

gdaviet:
Thanks @steinraf, this is indeed a bug and the fix looks good to me!
Ideally we would also verify the sampled values in the test, but it looks like the exsiting `test_volume_from_numpy` was not doing that either, so that's something we can add on our side

PR #327, opened 2024-10-10
Fixes a bug when assigning a function to an intermediate variable

mehdiataei:
## Category

- [ ] New feature
- [x] Bugfix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please explain)

## Description

Fixes a bug when assigning a function to an intermediate variable. Warp's code generator currently doesn't handle this pattern correctly. When a function is returned, the code generator replaces the expression with a special AST node (`ast.Name`) with an identifier `__warp_func__` and attaches the actual function object to it using an attribute `warp_func`.

In the example repro below from the doc, when the code generator later encounters the name `func` in `output[tid] = func(a, b)`, it tries to resolve `func` but fails because it doesn't find `__warp_func__` in its symbol table, leading to the error.

**NOTE: All tests pass, including the newly added test. However, please review carefully. Fingers crossed my tweaks don't unleash chaos.**

```python
import warp as wp

@wp.func
def do_add(a: float, b: float):
    return a + b

@wp.func
def do_sub(a: float, b: float):
    return a - b

@wp.func
def do_mul(a: float, b: float):
    return a * b

op_handlers = {
    "add": do_add,
    "sub": do_sub,
    "mul": do_mul,
}

inputs = wp.array([[1, 2], [3, 0]], dtype=wp.float32)
outputs = wp.empty(2, dtype=wp.float32)

for op in op_handlers.keys():

    @wp.kernel
    def operate(input: wp.array(dtype=inputs.dtype, ndim=2), output: wp.array(dtype=wp.float32)):
        tid = wp.tid()
        a, b = input[tid, 0], input[tid, 1]
        # retrieve the right function to use for the captured dtype variable
        output[tid] = wp.static(op_handlers[op])(a, b) # this works (as per the docs)
        # ERROR: But below does not work unexpectedly (even though it should be equivalent)
        # func = wp.static(op_handlers[op])
        # output[tid] = func(a, b) # this does not work

    wp.launch(operate, dim=2, inputs=[inputs], outputs=[outputs])

print(outputs.numpy())
```

## Changelog

- Allow functions to be correctly assigned to variables in Warp kernels.

- [x] Do you agree to the terms under which contributions are accepted as described in Section 9 the Warp [License](https://github.com/NVIDIA/warp/blob/main/LICENSE.md)?
- [x] Have you read the [Contributor Guidelines](https://github.com/NVIDIA/warp/blob/main/CONTRIBUTING.md)?
- [x] Have you written any new necessary tests?
- [ ] Have you added or updated any necessary documentation?
- [x] Have you added any files modified by compiling Warp and building the documentation to this PR (.e.g. `stubs.py`, `functions.rst`)?
- [x] Does your code pass `ruff check` and `ruff format --check`?

### DIFF
--- a/warp/codegen.py
+++ b/warp/codegen.py
@@ -1659,6 +1659,9 @@ class Adjoint:
         # lookup symbol, if it has already been assigned to a variable then return the existing mapping
         if node.id in adj.symbols:
             return adj.symbols[node.id]
+        # Check if the node has a warp_func attribute
+        if hasattr(node, 'warp_func'):
+            return node.warp_func
 
         obj = adj.resolve_external_reference(node.id)
 
@@ -2323,6 +2326,18 @@ class Adjoint:
                 raise WarpCodegenError(
                     "Tuple constructs are not supported in kernels. Use vectors like `wp.vec3()` for small collections instead."
                 )
+            elif isinstance(lhs, ast.Name):
+                # symbol name
+                name = lhs.id
+
+                # evaluate rhs
+                rhs = adj.eval(node.value)
+
+                # Check if rhs is a function object
+                if isinstance(rhs, warp.context.Function):
+                    # Assign the function directly to the symbol table
+                    adj.symbols[name] = rhs
+                    return
 
         # handle the case where we are assigning multiple output variables
         if isinstance(lhs, ast.Tuple):
--- a/warp/codegen.py
+++ b/warp/codegen.py
@@ -1660,7 +1660,7 @@ class Adjoint:
         if node.id in adj.symbols:
             return adj.symbols[node.id]
         # Check if the node has a warp_func attribute
-        if hasattr(node, 'warp_func'):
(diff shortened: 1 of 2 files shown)


shi-eric:
This is addressed by 8204312f165495aeda3352d2cf11b991eca79c4f.

The landed change covers the function-local binding bug described here and extends the fix beyond this PR: direct `@wp.func` locals, `wp.static(...)` returning a function, `wp.grad(f)` where `f` is function-valued, clear errors for unsupported rebinding, and module dependency tracking for functions bound through locals.

Test coverage was added for:
- `warp/tests/test_codegen.py`: direct function locals, static function locals, gradients through function locals, and unsupported rebinding cases.
- `warp/tests/test_reload.py`: module references through local-bound functions, including tuple binding.

Since the fix is now on `main` and this PR is stale/dirty against the current code layout, closing this as redundant.

PR #560, opened 2025-03-03
Fixed unrolling bug with static expressions

mehdiataei:
## Category

- [ ] New feature
- [x] Bugfix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please explain)

## Description

Fixes 
https://github.com/Autodesk/XLB/issues/114

Forced loop unrolling when wp.static expressions are detected prevents runtime evaluation errors caused by undefined loop variables.

## Changelog

- Add specific line-by-line info of user-facing changes in this PR.

- [x] Do you agree to the terms under which contributions are accepted as described in Section 9 the [Warp License](https://github.com/NVIDIA/warp/blob/main/LICENSE.md)?
- [ ] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `stubs.py`, `functions.rst`)
- [x] Code passes `ruff check` and `ruff format --check`?

### DIFF
--- a/warp/codegen.py
+++ b/warp/codegen.py
@@ -2041,6 +2041,37 @@ class Adjoint:
 
             ok_to_unroll = True
 
+            # Check if the loop body contains any wp.static expressions
+            contains_static = False
+
+            # Helper function to check for wp.static expressions
+            def check_for_static(node):
+                if isinstance(node, ast.Call):
+                    if isinstance(node.func, ast.Attribute):
+                        if (
+                            isinstance(node.func.value, ast.Name)
+                            and node.func.value.id == "wp"
+                            and node.func.attr == "static"
+                        ):
+                            return True
+                    return any(check_for_static(arg) for arg in node.args) or any(
+                        check_for_static(kw.value) for kw in node.keywords
+                    )
+                return False
+
+            # Check if loop body contains static expressions
+            contains_static = any(check_for_static(node) for node in ast.walk(loop))
+
+            # Always unroll if the loop contains static expressions
+            if contains_static:
+                # Forced unrolling for loops with static expressions regardless of max_unroll
+                if warp.config.verbose and max_iters > max_unroll:
+                    print(
+                        f"Notice: Forcing unroll of loop with {max_iters} iterations because it contains wp.static expressions."
+                    )
+                return range(start, end, step)
+
+            # Apply max_unroll check only for regular loops (no static expressions)
             if max_iters > max_unroll:
                 if warp.config.verbose:
                     print(


shi-eric:
What's the status of this pull request? Does the underlying bug still exist? If so, can you update the pull request and add unit tests for the issue?

mehdiataei:
@shi-eric Just rebased and added test for it.

warp/_src/codegen.py
```diff
@@ -2291,6 +2291,37 @@ def get_unroll_range(adj, loop):
 
             ok_to_unroll = True
 
+            # Check if the loop body contains any wp.static expressions
+            contains_static = False
+
+            # Helper function to check for wp.static expressions
+            def check_for_static(node):
+                if isinstance(node, ast.Call):
```
eric-heiden: Can you use `adj.is_static_expression(func)` here? That check should be more reliable than this custom code. See `Adjoint.replace_static_expressions()` for example.

warp/tests/test_static.py
```diff
@@ -346,6 +346,33 @@ def static_loop_variable(results: wp.array(dtype=int)):
...
+
+        test.assertEqual(results[0], sum(range(8)), "Static for loop must compute the correct solution")
+
+        if hasattr(static_loop_force_unroll.adj, "blocks"):
+            counts = count_ssa_occurrences(static_loop_force_unroll, ["add", "for"])
+            test.assertEqual(counts["add"], 8, "Static for loop must be unrolled")
+            test.assertGreaterEqual(counts["for"], 1, "Static for loop must be unrolled")
+            test.assertLessEqual(counts["for"], 2, "Static for loop must be unrolled")
+    finally:
+        wp.set_module_options(original_options)
```
eric-heiden, line 373: Thanks for this test! I think it would be good to also test the scenario where a static expression has no relation to the loop iteration variable `i`. In that case the loop should not be force to unroll, so there should be a for loop in the generated code. This should already happen with what you have now because static expressions that don't depend on loop variables will be replaced at the time of declaration. It would just be good to have a test for this.

warp/_src/codegen.py
mehdiataei: Done

warp/tests/test_static.py
mehdiataei: Done.

eric-heiden:
pre-commit.ci autofix

eric-heiden:
Thanks @mehdiataei, this looks good to me! We need you to sign the DCO to be able to merge your PR.

mehdiataei:
> Thanks @mehdiataei, this looks good to me! We need you to sign the DCO to be able to merge your PR.

Thanks. Signed DCO

shi-eric:
Sorry, one more request: Can you squash this down to one commit?

mehdiataei:
Done

shi-eric:
Thanks! I'll get it merged soon.

PR #592, opened 2025-03-23
Add external resource interoperability

ramenguy99:
## Category

- [x] New feature
- [ ] Bugfix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please explain)

## Description

Hi, this PR adds support for the [External Resource Interoperability API](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html) for interoperability between Cuda and other graphics APIs such as Vulkan, D3D12 and D3D11.

PR is functional and tested on windows with Vulkan, some docs are missing and some cleanup might be needed. If there is interest in merging something like this I am more than happy to complete this.

~ Dario

## Changelog

- Expose External Resource Interoperability API through `ExternalSemaphore` and `ExternalMemoryBuffer`

- [x] All commits are [signed-off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) to indicate that your contribution
adheres to the [Developer Certificate of Origin](https://developercertificate.org/) requirements
- [ ] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `stubs.py`, `functions.rst`)

### DIFF
--- a/warp/__init__.py
+++ b/warp/__init__.py
@@ -82,6 +82,7 @@ from warp.context import capture_begin, capture_end, capture_launch
 from warp.context import Kernel, Function, Launch
 from warp.context import Stream, get_stream, set_stream, wait_stream, synchronize_stream
 from warp.context import Event, record_event, wait_event, synchronize_event, get_event_elapsed_time
+from warp.context import ExternalMemoryBuffer, ExternalSemaphore, signal_external_semaphore, wait_external_semaphore
 from warp.context import RegisteredGLBuffer
 from warp.context import is_mempool_supported, is_mempool_enabled, set_mempool_enabled
 from warp.context import (
--- a/warp/context.py
+++ b/warp/context.py
@@ -2669,6 +2669,20 @@ class Stream:
         """
         runtime.core.cuda_stream_wait_event(self.cuda_stream, event.cuda_event)
 
+    def signal_external_semaphore(self, semaphore: ExternalSemaphore, value: int = 0):
+        """TODO: docs"""
+
+        runtime.core.cuda_signal_external_semaphore_async(
+            semaphore.device.context, semaphore.external_semaphore, value, self.cuda_stream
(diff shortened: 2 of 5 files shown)


shi-eric:
I think @nvlukasz will be back next week. Will remind him to take a look.

nvlukasz:
Thanks @ramenguy99, this looks like a potentially useful contribution.

My main concern is that I'm not aware of any immediate use for it, so we don't have good test cases or usage examples.

Can you share more information about how you are using this functionality?

ManosAgelidis:
@nvlukasz If I may add a point to the discussion, we are also looking into integrating warp into UE5. Currently there is a lot of overhead due to costly UE5 rendering thread->UE CPU->warp python->warp CUDA->warp python->UE5 CPU->UE5 rendering thread data transfers taking place. This is due to the fact that we want to update the vertices of meshes in UE5 (which already live on the GPU) through warp and there is currently no interop between UE5's rendering thread and warp. Not sure what this PR does but the feature itself could be quite useful.

shi-eric:
I agree with the earlier discussion that external resource interop is a potentially useful direction for Warp, especially for rendering and engine integrations where the data already lives on the GPU. That said, I do not think this PR is in a state where we can move it forward as-is. It is stale against the current codebase, does not include tests, docs, or an example, and the API surface needs more design work before becoming part of Warp’s public API.

I’m going to close this PR for now, but I do not want to close the door on the feature. Since there is no GitHub issue tracking the request or proposed API, the best next step would be to open a fresh issue or design discussion focused on a concrete integration path and testable subset of the API, for example Vulkan/D3D external memory buffers plus the semaphore types needed for that workflow. From there we can agree on scope, ownership/lifetime semantics, platform support, and what examples or tests would make the feature maintainable.

PR #777, opened 2025-06-08
Fix shared tile memory managment and select/where ops

RSchwan:
## Description

This PR fixes memory issues with shared tiles and fixes compilation issues when using `wp.select()` or `wp.where()` with tiles and fixes their adj propagation.

Memory bugs were arising when copying `tile_shared_t` instances in cpp. The default copy constructor would just copy the `data` and `grad` pointers, as well as ownership. Hence, at deconstruction, the newly copied `tile_shared_t` would also de-allocate it's data leading to a double-free. Deleting the default constructor then revealed all places in the code with shared tile copies, which are now handled manually to avoid the mentioned issues and have correct ownership.

In a similar spirit, the assignment operator `operator=()` of `tile_shared_t` was causing similar memory issues. The previous implementation would just copy the `data` and `grad` pointers, even when it was owning. This also leads to double-frees and memory leaks. Now, if `tile_shared_t` is owning, it copies the data instead.

With these changes in mind, the `wp.copy()` and `wp.select() / wp.where()` have also been slightly modify. In particular, if it receives a tile argument, they now make sure that the resulting tile has the right type and ownership. For example, if one argument is not owning (e.g. after `wp.transpose()` operation), then it's now ensured that the result is owning. This is important for correct adj propagation.

Tests have been added.

PS: `wp.select()` is deprecated. When are you planning to remove it? This would clean up the code quite a bit and remove duplication. On this note, the architecure of `tile_shared_t` is rather strange. Why store the `data` and `grad` in the same struct? As far as I can see, are the adjs for other types are all stored in separate instances. This leads to strange patterns where the adj vars are references to the normal var:
```cpp
wp::tile_shared_t<wp::float32,wp::tile_layout_strided_t<wp::tile_shape_t<8>, wp::tile_stride_t<1>>, true> var_26 = wp::tile_alloc_empty<wp::float32,wp::tile_shape_t<8>,wp::tile_shape_t<1>,true>();
...
wp::tile_shared_t<wp::float32,wp::tile_layout_strided_t<wp::tile_shape_t<8>, wp::tile_stride_t<1>>, true>& adj_26 = var_26;
```
Down the line, this then leads to strange patterns where the [`operator+=` is used to accumulate gradients](https://github.com/NVIDIA/warp/blob/0f9764d23a487b71e35a2921e64d700480cbdd6e/warp/native/tile.h#L532). I would suggest detaching the grad from the `tile_shared_t` struct and just have two instances. This then would also align how it's done in `tile_register_t` and all other types. I would be happy to help, but before, I wanted to understand the rationale behind it and if there are any things I missed. But this is for another PR. :)

- [x] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `stubs.py`, `functions.rst`)

### DIFF
--- a/warp/builtins.py
+++ b/warp/builtins.py
@@ -6074,14 +6074,33 @@ add_builtin(
 )
 
 
+def copy_value_func(arg_types: Mapping[str, type], arg_values: Mapping[str, Any]):
+    a = arg_types["a"]
+
+    # if the input is a shared tile, we force a copy
+    if is_tile(a) and a.storage == "shared":
+        return tile(
+            dtype=a.dtype,
+            shape=a.shape,
+            storage=a.storage,
+            strides=a.strides,
+            layout=a.layout,
+            owner=True,
+        )
+
+    return a
+
+
 add_builtin(
     "copy",
     input_types={"a": Any},
-    value_func=lambda arg_types, arg_values: arg_types["a"],
+    value_func=copy_value_func,
     hidden=True,
     export=False,
     group="Utility",
 )
+
+
 add_builtin(
     "assign",
     input_types={"dest": Any, "src": Any},
@@ -6089,6 +6108,37 @@ add_builtin(
     export=False,
     group="Utility",
 )
+
+
+def select_value_func(arg_types: Mapping[str, type], arg_values: Mapping[str, Any]):
+    if arg_types is None:
+        return Any
+
+    v_true = arg_types["value_if_true"]
+    v_false = arg_types["value_if_false"]
+
+    if not types_equal(v_true, v_false):
+        raise RuntimeError(
+            f"select() true value type ({v_true}) must be of the same type as the false type ({v_false})"
+        )
+
+    if is_tile(v_false):
+        if v_true.storage == "register":
+            return v_true
+        if v_false.storage == "register":
+            return v_false
+
+        # both v_true and v_false are shared
+        return tile(
+            dtype=v_true.dtype,
+            shape=v_true.shape,
+            storage=v_true.storage,
+            strides=v_true.strides,
+            layout=v_true.layout,
+            owner=True,
+        )
+
+    return v_true
 
 
 def select_dispatch_func(input_types: Mapping[str, type], return_type: Any, args: Mapping[str, Var]):
@@ -6107,7 +6157,7 @@ def select_dispatch_func(input_types: Mapping[str, type], return_type: Any, args
 add_builtin(
     "select",
     input_types={"cond": builtins.bool, "value_if_false": Any, "value_if_true": Any},
-    value_func=lambda arg_types, arg_values: Any if arg_types is None else arg_types["value_if_false"],
+    value_func=select_value_func,
     dispatch_func=select_dispatch_func,
     doc="""Select between two arguments, if ``cond`` is ``False`` then return ``value_if_false``, otherwise return ``value_if_true``.
 
@@ -6120,7 +6170,7 @@ for t in int_types:
     add_builtin(
         "select",
         input_types={"cond": t, "value_if_false": Any, "value_if_true": Any},
-        value_func=lambda arg_types, arg_values: Any if arg_types is None else arg_types["value_if_false"],
+        value_func=select_value_func,
         dispatch_func=select_dispatch_func,
         doc="""Select between two arguments, if ``cond`` is ``False`` then return ``value_if_false``, otherwise return ``value_if_true``.
 
@@ -6132,7 +6182,7 @@ for t in int_types:
 add_builtin(
     "select",
     input_types={"arr": array(dtype=Any), "value_if_false": Any, "value_if_true": Any},
-    value_func=lambda arg_types, arg_values: Any if arg_types is None else arg_types["value_if_false"],
+    value_func=select_value_func,
     dispatch_func=select_dispatch_func,
     doc="""Select between two arguments, if ``arr`` is null then return ``value_if_false``, otherwise return ``value_if_true``.
 
@@ -6142,10 +6192,40 @@ add_builtin(
     group="Utility",
 )
 
+
+def where_value_func(arg_types: Mapping[str, type], arg_values: Mapping[str, Any]):
+    if arg_types is None:
+        return Any
+
+    v_true = arg_types["value_if_true"]
+    v_false = arg_types["value_if_false"]
+
+    if not types_equal(v_true, v_false):
+        raise RuntimeError(f"where() true value type ({v_true}) must be of the same type as the false type ({v_false})")
+
+    if is_tile(v_false):
+        if v_true.storage == "register":
+            return v_true
+        if v_false.storage == "register":
+            return v_false
+
+        # both v_true and v_false are shared
+        return tile(
+            dtype=v_true.dtype,
+            shape=v_true.shape,
+            storage=v_true.storage,
+            strides=v_true.strides,
+            layout=v_true.layout,
+            owner=True,
+        )
+
+    return v_true
+
+
 add_builtin(
     "where",
     input_types={"cond": builtins.bool, "value_if_true": Any, "value_if_false": Any},
-    value_func=lambda arg_types, arg_values: Any if arg_types is None else arg_types["value_if_false"],
+    value_func=where_value_func,
     doc="Select between two arguments, if ``cond`` is ``True`` then return ``value_if_true``, otherwise return ``value_if_false``.",
     group="Utility",
 )
@@ -6153,14 +6233,14 @@ for t in int_types:
     add_builtin(
         "where",
         input_types={"cond": t, "value_if_true": Any, "value_if_false": Any},
-        value_func=lambda arg_types, arg_values: Any if arg_types is None else arg_types["value_if_false"],
+        value_func=where_value_func,
         doc="Select between two arguments, if ``cond`` is ``True`` then return ``value_if_true``, otherwise return ``value_if_false``.",
         group="Utility",
     )
 add_builtin(
     "where",
     input_types={"arr": array(dtype=Any), "value_if_true": Any, "value_if_false": Any},
-    value_func=lambda arg_types, arg_values: Any if arg_types is None else arg_types["value_if_false"],
+    value_func=where_value_func,
     doc="Select between two arguments, if ``arr`` is not null then return ``value_if_true``, otherwise return ``value_if_false``.",
     group="Utility",
 )
--- a/warp/native/tile.h
+++ b/warp/native/tile.h
@@ -542,7 +542,7 @@ struct tile_register_t
 
     // define the += operator which is used during backward pass codegen
     // when returning a register tile from a user defined function
-    inline CUDA_CALLABLE auto& operator += (tile_register_t<T, Layout>& rhs) 
+    inline CUDA_CALLABLE auto& operator += (const tile_register_t<T, Layout>& rhs) 
     {
         grad_add(rhs);
         return *this;
@@ -658,7 +658,7 @@ struct tile_register_t
             data[i] += tile.data[i];
     }
 
-    CUDA_CALLABLE void grad_add(const tile_global_t<T, typename Layout::Shape>& global) 
+    inline CUDA_CALLABLE void grad_add(const tile_global_t<T, typename Layout::Shape>& global) 
     {
         apply([&](int reg, auto c) {data[reg] += global.load_grad(c);});
     }
@@ -758,6 +758,7 @@ inline CUDA_CALLABLE void* tile_alloc_shared(int num_bytes, bool init=false, boo
         
         // one entry per-thread so no need for synchronization
         smem_base[WP_TILE_THREAD_IDX] += tile_align(num_bytes);
+        assert(smem_base[WP_TILE_THREAD_IDX] >= 0);
 
 #ifdef __CUDA_ARCH__
         extern __shared__ char dynamic_smem_base[];
@@ -905,6 +906,28 @@ struct tile_shared_t
     {
     }
 
+    // we delete the copy constructor because in the case the shared tile is owning,
+    // this leads to a double deallocation. 
+    // this also forces one to handle copies explicitly
+    inline CUDA_CALLABLE tile_shared_t(const tile_shared_t& other) : data(other.data), grad(other.grad), initialized(other.initialized)
+    {
+        static_assert(!Owner, "Copy constructor is only supported for non-owning tiles.");
+    }
+
+    // move constructor
+    inline CUDA_CALLABLE tile_shared_t(tile_shared_t&& other) : data(other.data), grad(other.grad), initialized(other.initialized)
+    {
+        other.data.ptr = nullptr;
+        other.grad.ptr = nullptr;
+    }
+
+    template <typename OtherT, typename OtherLayout, bool OtherOwner>
+    inline CUDA_CALLABLE tile_shared_t(const tile_shared_t<OtherT, OtherLayout, OtherOwner>& other) : data(other.data.ptr), grad(other.grad.ptr), initialized(other.initialized)
+    {
+        static_assert(!Owner, "Copy constructor is only supported for non-owning tiles.");
+        static_assert(Layout::Size == OtherLayout::Size, "Expected Size == OtherLayout::Size");
+    }
+
     // initialize from an existing tile's memory
     inline CUDA_CALLABLE tile_shared_t(T* data, T* grad=nullptr, bool initialized=true) : data(data), grad(grad), initialized(initialized)
     {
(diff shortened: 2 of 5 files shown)


shi-eric:
> `wp.select()` is deprecated. When are you planning to remove it?

According to my notes, we deprecated it in v1.7. The earliest we can remove it is v1.9.

We'll need @mmacklin to look at this PR, but he's in Paris this week for GTC.

RSchwan:
Ok, after playing around a bit more with this PR, I ran into other issues. For example, the following will not compile:
```python
@wp.func
def tile_chol(x: wp.tile(dtype=wp.float64, shape=(2, 2), storage='shared')):
    return wp.tile_cholesky(x)

@wp.kernel
def test_multi_valued_kernel(L: wp.array2d(dtype=wp.float64)):
    L_tile = wp.tile_load(L, shape=(2, 2), offset=(0, 0), storage='shared')
    L_tile = tile_chol(L_tile)
    wp.tile_store(L, L_tile, offset=(0, 0))

a = wp.ones((2, 2), dtype=wp.float64, device='cuda')
wp.launch_tiled(test_multi_valued_kernel, dim=[1], inputs=[a], block_dim=32)
```

The generated code for the forward pass
```cpp
static CUDA_CALLABLE wp::tile_shared_t<wp::float64,wp::tile_layout_strided_t<wp::tile_shape_t<2,2>, wp::tile_stride_t<2,1>>, true> tile_chol_0(
    wp::tile_shared_t<wp::float64,wp::tile_layout_strided_t<wp::tile_shape_t<2,2>, wp::tile_stride_t<2,1>>, true> var_x)
{
    //---------
    // primal vars
    wp::tile_shared_t<wp::float64,wp::tile_layout_strided_t<wp::tile_shape_t<2,2>, wp::tile_stride_t<2,1>>, true> var_0 = wp::tile_alloc_empty<wp::float64,wp::tile_shape_t<2,2>,wp::tile_shape_t<2,1>,false>();
    //---------
    // forward
    // def tile_chol(x: wp.tile(dtype=wp.float64, shape=(2, 2), storage='shared')):           <L 2>
    // return wp.tile_cholesky(x)                                                             <L 3>
    var_0 = tile_cholesky(potrf_2_2_1_86_32_1_1_6_x_x_1, var_x, var_0);
    return var_0;
}

extern "C" __global__ void test_multi_valued_kernel_1185c4f0_cuda_kernel_forward(
    wp::launch_bounds_t dim,
    wp::array_t<wp::float64> var_L)
{
    for (size_t _idx = static_cast<size_t>(blockDim.x) * static_cast<size_t>(blockIdx.x) + static_cast<size_t>(threadIdx.x);
         _idx < dim.size;
         _idx += static_cast<size_t>(blockDim.x) * static_cast<size_t>(gridDim.x))
    {
        // reset shared memory allocator
        wp::tile_alloc_shared(0, true);

        //---------
        // primal vars
        const wp::int32 var_0 = 2;
        const wp::int32 var_1 = 2;
        wp::tuple_t<wp::int32, wp::int32> var_2;
        const wp::int32 var_3 = 0;
        const wp::int32 var_4 = 0;
        wp::tuple_t<wp::int32, wp::int32> var_5;
        const wp::str var_6 = "shared";
        wp::tile_shared_t<wp::float64,wp::tile_layout_strided_t<wp::tile_shape_t<2,2>, wp::tile_stride_t<2,1>>, true> var_7 = wp::tile_alloc_empty<wp::float64,wp::tile_shape_t<2,2>,wp::tile_shape_t<2,1>,false>();
        const wp::int32 var_8 = 0;
        const wp::int32 var_9 = 0;
        wp::tile_shared_t<wp::float64,wp::tile_layout_strided_t<wp::tile_shape_t<2,2>, wp::tile_stride_t<2,1>>, true> var_10 = wp::tile_alloc_empty<wp::float64,wp::tile_shape_t<2,2>,wp::tile_shape_t<2,1>,false>();
        const wp::int32 var_11 = 0;
        const wp::int32 var_12 = 0;
        wp::tuple_t<wp::int32, wp::int32> var_13;
        const wp::int32 var_14 = 0;
        const wp::int32 var_15 = 0;
        //---------
        // forward
        // def test_multi_valued_kernel(L: wp.array2d(dtype=wp.float64)):                         <L 6>
        // L_tile = wp.tile_load(L, shape=(2, 2), offset=(0, 0), storage='shared')                <L 7>
        var_2 = wp::tuple(var_0, var_1);
        var_5 = wp::tuple(var_3, var_4);
        var_7 = wp::tile_load<2, 2>(var_L, var_8, var_9);
        // L_tile = tile_chol(L_tile)                                                             <L 8>
        var_10 = tile_chol_0(var_7);
        // wp.tile_store(L, L_tile, offset=(0, 0))                                                <L 9>
        var_13 = wp::tuple(var_11, var_12);
        wp::tile_store(var_L, var_14, var_15, var_10);
    }
}
```
The issue now is that I deleted the copy constructor of `tile_shared_t`. One easy fit fix would be to pass arguments as reference to the generation warp functions. Is there any downside in doing so? As far as I've seen in the code gen, everything is treated as immutable anyway hence, we can avoid copies to function calls.

mmacklin:
Thanks @RSchwan, regarding the shared tiles storing both the `data` and `grad` pointers. Originally these were separated, however the reason for combining them was due to slicing and the backward pass. Specifically when we slice in the forward pass we need generate a new tile that references the original data, however we also need to slice the gradient tile, which turns out to be somewhat complicated to do with SSA forward/reverse code generation. Let me sketch out the problem in psuedo-code:

```cpp
// allocate forward vars
var_0 = wp.tile_alloc()
var_1 = None

// allocate gradient vars
adj_0 = wp.tile_alloc()
adj_1 = None

// forward pass
var_0 = wp.tile_load(src)
var_1 = wp.tile_slice(var_0, index)   // create a reference to the loaded tile
wp.tile_store(dest, var_1)

// backward pass
wp.adj_tile_store(dest, var_1, adj_1)              // want to write to adj_1 but it hasn't been initialized yet
wp.adj_tile_slice(var_0, index, adj_0, adj_index, adj_1)   // could initialize adj_1 here, but it's too late
wp.adj_tile_load(src)
```

If we track data/grad together then we can essentially force the dims to match and initialize them together during the forward pass. It's a subtle issue, but hope that helps explain the decision. I would be happy to revisit this, since I agree that it greatly complicates the code.

One solution would be to say that gradients never reference other memory, and always allocate a copy of the gradient instead. That would likely increase peak shared memory usage, but may not be too bad depending on how often things like slicing are used.

RSchwan:
I see the problem. And what if one would modify the code generation instead? For example, for all slicing/reshaping operations, one could already prime the adj in the forward pass. Off course, this just moves the complexity somewhere else, but at least all it would make the tiles consistent with all other data structures.
For example, something along these lines:
```
// allocate forward vars
var_0 = wp.tile_alloc()
var_1 = None

// allocate gradient vars
adj_0 = wp.tile_alloc()
adj_1 = None

// forward pass
var_0 = wp.tile_load(src)
var_1 = wp.tile_slice(var_0, index)   // create a reference to the loaded tile
adj_1 = wp.tile_slice(adj_0, index)   // generate reference in forward pass explicitly for adj 
wp.tile_store(dest, var_1)

// backward pass
wp.adj_tile_store(dest, var_1, adj_1)
wp.adj_tile_slice(var_0, index, adj_0, adj_index, adj_1)
wp.adj_tile_load(src)
```
Always allocating a copy of the gradient also would require a change of the code gen. And I'm not a big fan of throwing more memory at the problem, i.e., warp should be as close as possible in performance (and memory usage) to writing cuda by hand. I'm currently writing specialized factorization routines in warp, and I'm already running a bit into shared memory limitations due to not being able to reuse shared memory efficiently (only forward pass though).

mmacklin:
Yes, modifying code-gen is certainly an option, we'd need to be careful to do it where ever these types of reference operations occur so hopefully could find a general purpose mechanism rather than on a per-operation basis.

Btw the wp.array data type works the same way as shared tiles (tracking data/grad together) for much the same reason. Just want to point out that this isn't the only data type following this approach, but agree in general that code is much simpler if you can treat these things as separate.

Regarding shared memory allocation, you probably noticed that right now Warp uses a stack allocator for shared memory, meaning you should be able to get some re-use by hoisting work into separate functions. This could also be improved more sophisticated (possibly compile-time) methods.

RSchwan:
I wanted to reactivate this PR as it still contains important memory fixes for shared tiles. For example, the provided new test case fails without the fixes in this PR.

I rebased it on the main branch, and all tests are passing on my machine.

daedalus5:
Thanks @RSchwan. I was holding off because we might refactor some of the tile API, but these are good changes that would help at present. I'll take a closer look and review soon.

warp/builtins.py
```diff
@@ -5975,14 +5975,33 @@ def printf_dispatch_func(input_types: Mapping[str, type], return_type: Any, args
 )
 
 
+def copy_value_func(arg_types: Mapping[str, type], arg_values: Mapping[str, Any]):
+    a = arg_types["a"]
+
+    # if the intput is a shared tile without ownership, we force a copy
+    if is_tile(a) and a.storage == "shared":
```
daedalus5, line 6081: With `wp.copy()` we always want to deep copy for shared tiles, which is what this does. Comment is slightly ambiguous though, so maybe
`# if the input is a shared tile, we force a copy`

warp/builtins.py
```diff
@@ -5992,6 +6011,36 @@ def printf_dispatch_func(input_types: Mapping[str, type], return_type: Any, args
...
+    v_true = arg_types["value_if_true"]
+    v_false = arg_types["value_if_false"]
+
+    if not types_equal(v_true, v_false):
+        raise RuntimeError(
+            f"select() true value type ({v_true}) must be of the same type as the false type ({v_false})"
+        )
+
+    # if one of the tiles is a register, use also register storage for result
+    if is_tile(v_false) and v_false.storage == "register":
```
daedalus5: Need some clarity here. Can we do

```python
if is_tile(v_false):
    if v_true.storage == "register":
        return v_true
    if v_false.storage == "register":
        return v_false
    return tile (...)  # both v_true and v_false are shared
```

warp/native/tile.h
```diff
@@ -1053,6 +1119,26 @@ struct tile_shared_t
         WP_TILE_SYNC();
     }
 
+    template <typename OtherT, typename OtherLayout, bool OtherOwner>
```
daedalus5, line 1163: comment:
// shared tile deep copy

warp/native/tile.h
```diff
@@ -1502,13 +1608,81 @@ inline CUDA_CALLABLE void adj_len(const tile_register_t<T,L>& t, const AdjTile&
 {
 }
 
+// select specialization for shared tiles
+template <typename C, typename T, typename LRegister, typename LShared, bool Owner>
+inline CUDA_CALLABLE auto select(const C& cond, const tile_register_t<T, LRegister>& a, const tile_shared_t<T, LShared, Owner>& b)
+{
+    // The double NOT operator !! casts to bool without compiler warnings.
+    return (!!cond) ? b : a.copy_to_register();
```
daedalus5: `b.copy_to_register()` instead of `a`

warp/native/tile.h
```diff
@@ -1502,13 +1608,81 @@ inline CUDA_CALLABLE void adj_len(const tile_register_t<T,L>& t, const AdjTile&
...
+}
 
-template <typename T, typename L>
-inline CUDA_CALLABLE void adj_print(const tile_register_t<T, L>& t, const tile_register_t<T, L>& a) { a.print(); }
-template <typename T, typename L, bool Owner>
-inline CUDA_CALLABLE void adj_print(const tile_shared_t<T, L, Owner>& t, const tile_shared_t<T, L, Owner>& a) { a.print(true); }
+template <typename C, typename T, typename LRegister, typename LShared, bool Owner>
+inline CUDA_CALLABLE auto select(const C& cond, const tile_shared_t<T, LShared, Owner>& a, const tile_register_t<T, LRegister>& b)
```
daedalus5, line 1661: `a.copy_to_register()` instead of `b`

warp/native/tile.h
```diff
@@ -1502,13 +1608,81 @@ inline CUDA_CALLABLE void adj_len(const tile_register_t<T,L>& t, const AdjTile&
...
+{
+    // The double NOT operator !! casts to bool without compiler warnings.
+    return (!!cond) ? b.copy_to_register() : a;
+}
+
+template <typename C, typename T, typename L, bool Owner>
+inline CUDA_CALLABLE auto select(const C& cond, const tile_shared_t<T, L, Owner>& a, const tile_shared_t<T, L, Owner>& b)
```
daedalus5, line 1668: Redundant, given the more general template that follows

warp/native/tile.h
```diff
@@ -1502,13 +1608,81 @@ inline CUDA_CALLABLE void adj_len(const tile_register_t<T,L>& t, const AdjTile&
...
+
+template <typename C, typename T, typename LRegister, typename LShared, bool Owner>
+inline CUDA_CALLABLE auto where(const C& cond, const tile_shared_t<T, LShared, Owner>& a, const tile_register_t<T, LRegister>& b)
+{
+    // The double NOT operator !! casts to bool without compiler warnings.
+    return (!!cond) ? a.copy_to_register() : b;
+}
+
+template <typename C, typename T, typename L, bool Owner>
+inline CUDA_CALLABLE auto where(const C& cond, const tile_shared_t<T, L, Owner>& a, const tile_shared_t<T, L, Owner>& b)
```
daedalus5, line 1699: Also redundant.

warp/native/tile.h
```diff
@@ -1502,13 +1608,81 @@ inline CUDA_CALLABLE void adj_len(const tile_register_t<T,L>& t, const AdjTile&
...
+template <typename C, typename T, typename L, bool LOwner, bool ROwner>
+inline CUDA_CALLABLE auto where(const C& cond, const tile_shared_t<T, L, LOwner>& a, const tile_shared_t<T, L, ROwner>& b)
+{
+    // The double NOT operator !! casts to bool without compiler warnings.
+    return (!!cond) ? tile_shared_t<T, L, false>(a.data.ptr, a.grad.ptr) : tile_shared_t<T, L, false>(b.data.ptr, b.grad.ptr);
+}
+
+// adj_where same as in builtin.h
+
+// copy specialization for shared tiles
```
daedalus5: Could you add some additional commentary here? ie, lvalue is owning, this invokes the copy assign path.

warp/native/tile.h
```diff
@@ -1502,13 +1608,81 @@ inline CUDA_CALLABLE void adj_len(const tile_register_t<T,L>& t, const AdjTile&
...
+inline CUDA_CALLABLE auto copy(const tile_shared_t<T, L, Owner>& t)
+{
+    return tile_shared_t<T, L, false>(t.data.ptr, t.grad.ptr);
+}
+
+template <typename T, typename L, bool Owner>
+inline CUDA_CALLABLE void adj_copy(const tile_shared_t<T, L, Owner>& src, tile_shared_t<T, L, Owner>& adj_src, tile_shared_t<T, L, Owner>& adj_dest)
+{
+    adj_src += adj_dest;
+    adj_dest.grad_zero();
```
daedalus5, line 1725: I'm not sure the `grad_zero()` is necessary here. The dest variable gradient might be needed elsewhere no?

warp/tests/tile/test_tile.py
```diff
@@ -837,6 +837,66 @@ def test_tile_assign(test, device):
...
+        np.full(TILE_M, 0.0, dtype=np.float32),
+        np.full(TILE_M, 1.0, dtype=np.float32),
+    ]
+
+    for i in range(4):
+        tape = wp.Tape()
+        with tape:
+            wp.launch_tiled(test_tile_where_kernel, dim=[1], inputs=[i, x, y], outputs=[z], block_dim=32, device=device)
+
+        x.grad.fill_(0.0)
```
daedalus5: Gradient arrays are automatically initialized to zero

daedalus5 (changes requested):
Thanks for the PR, this is a nice fix. Just a few small things to look at.

Could you also add a changelog entry to CHANGELOG.md? You'll see the format we use.

warp/native/tile.h
RSchwan: This is needed due to ambiguity with the even more general definition in `builtin.h`:
```cpp
template <typename C, typename T>
CUDA_CALLABLE inline T select(const C& cond, const T& a, const T& b)
```
Here a small scale reproduction: https://godbolt.org/z/xzPjbe7jz

warp/native/tile.h
RSchwan: Same as previous comment

warp/native/tile.h
RSchwan: I followed the implementation in `builtin.h`:
```cpp
template <typename T>
CUDA_CALLABLE inline void adj_copy(const T& src, T& adj_src, T& adj_dest)
{
    adj_src += adj_dest;
    adj_dest = T{};
}
```
which also `zeros` out the destination variable

warp/tests/tile/test_tile.py
RSchwan: Are you sure? If I remove the lines
```python
x.grad.fill_(0.0)
y.grad.fill_(0.0)
```
the tests fail.

daedalus5: Ah because of the loop. The gradients need to be reset. What you have works, but normally we'd have:

```python
    z.grad = wp.ones_like(z)

    tape.backward()

    assert_np_equal(z.numpy(), z_expected[i])
    assert_np_equal(x.grad.numpy(), x_grad_expected[i])
    assert_np_equal(y.grad.numpy(), y_grad_expected[i])

    tape.zero()
```

warp/native/tile.h
daedalus5: Okay thinking on it some more this is safe, even if it might not be necessary. Let's keep as is.

RSchwan:
I squashed all commits into a single one and slightly updated the wording in the changelog.

PR #912, opened 2025-08-14
added JAX interop adjoint support using FFI

mehdiataei:
## Description

Added JAX interop adjoint support using FFI. Also added support for static inputs (with auto-detection), vmap, multi-output with AD. Fixed some bugs related to vmap for the forward pass. 

This is a new area for me, so there may be some oddities or poor decisions, but all tests pass. We may need to discuss any missing features and whether this is the way the user intends to use it.

This work is inspired by and thanks to J. Sevcik
https://gist.github.com/jaro-sevcik/893ffb891564dfc7617c8f12f3c2d1d2

### DIFF
--- a/warp/jax_experimental/ffi.py
+++ b/warp/jax_experimental/ffi.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import ctypes
+import inspect
 import threading
 import traceback
 from enum import IntEnum
@@ -173,14 +174,14 @@ class FfiKernel:
                     raise TypeError(
                         f"Invalid data type for array argument '{input_arg.name}', expected {input_arg.jax_scalar_type}, got {input_value.dtype}"
                     )
-                # check ndim
-                if input_value.ndim != input_arg.jax_ndim:
+                # allow leading batch dims under vmap; require trailing dtype dims to match
+                if input_value.ndim < input_arg.jax_ndim:
                     raise TypeError(
-                        f"Invalid dimensionality for array argument '{input_arg.name}', expected {input_arg.jax_ndim} dimensions, got {input_value.ndim}"
+                        f"Invalid dimensionality for array argument '{input_arg.name}', expected at least {input_arg.jax_ndim} dimensions, got {input_value.ndim}"
                     )
-                # check inner dims
+                # check inner (dtype) dims at the end
                 for d in range(input_arg.dtype_ndim):
-                    if input_value.shape[input_arg.type.ndim + d] != input_arg.dtype_shape[d]:
+                    if input_value.shape[-input_arg.dtype_ndim + d] != input_arg.dtype_shape[d]:
                         raise TypeError(
                             f"Invalid inner dimensions for array argument '{input_arg.name}', expected {input_arg.dtype_shape}, got {input_value.shape[-input_arg.dtype_ndim :]}"
                         )
@@ -193,7 +194,7 @@ class FfiKernel:
 
             # append in-out arg to output types
             if input_arg.in_out:
-                out_types.append(get_jax_output_type(input_arg, input_value.shape))
+                out_types.append(get_jax_output_type(input_arg, get_warp_shape(input_arg, input_value.shape)))
 
         # launch dimensions
         if launch_dims is None:
@@ -455,14 +456,14 @@ class FfiCallable:
                     raise TypeError(
                         f"Invalid data type for array argument '{input_arg.name}', expected {input_arg.jax_scalar_type}, got {input_value.dtype}"
                     )
-                # check ndim
-                if input_value.ndim != input_arg.jax_ndim:
+                # allow leading batch dims under vmap; require trailing dtype dims to match
+                if input_value.ndim < input_arg.jax_ndim:
                     raise TypeError(
-                        f"Invalid dimensionality for array argument '{input_arg.name}', expected {input_arg.jax_ndim} dimensions, got {input_value.ndim}"
+                        f"Invalid dimensionality for array argument '{input_arg.name}', expected at least {input_arg.jax_ndim} dimensions, got {input_value.ndim}"
                     )
-                # check inner dims
+                # check inner (dtype) dims at the end
                 for d in range(input_arg.dtype_ndim):
-                    if input_value.shape[input_arg.type.ndim + d] != input_arg.dtype_shape[d]:
+                    if input_value.shape[-input_arg.dtype_ndim + d] != input_arg.dtype_shape[d]:
                         raise TypeError(
                             f"Invalid inner dimensions for array argument '{input_arg.name}', expected {input_arg.dtype_shape}, got {input_value.shape[-input_arg.dtype_ndim :]}"
                         )
@@ -475,7 +476,7 @@ class FfiCallable:
 
             # append in-out arg to output types
             if input_arg.in_out:
-                out_types.append(get_jax_output_type(input_arg, input_value.shape))
+                out_types.append(get_jax_output_type(input_arg, get_warp_shape(input_arg, input_value.shape)))
 
         # output shapes
         if isinstance(output_dims, dict):
@@ -591,8 +592,25 @@ class FfiCallable:
             for i, arg in enumerate(self.input_args):
                 if arg.is_array:
                     buffer = inputs[i].contents
-                    shape = buffer.dims[: buffer.rank - arg.dtype_ndim]
-                    arr = wp.array(ptr=buffer.data, dtype=arg.type.dtype, shape=shape, device=device)
+                    # Collapse leading batch dimensions so Warp sees only the declared warp_ndim
+                    jax_rank = buffer.rank - arg.dtype_ndim
+                    dims_jax = buffer.dims[:jax_rank]
+                    warp_ndim = arg.type.ndim
+                    if jax_rank < warp_ndim:
+                        raise RuntimeError(
+                            f"Argument '{arg.name}' expects {warp_ndim} dimension(s) but the passed array has {jax_rank} dimension(s)."
+                        )
+                    if jax_rank == warp_ndim:
+                        collapsed_shape = dims_jax
+                    else:
+                        # collapse all leading (batch) dims into the first warp dimension
+                        batch_prod = 1
+                        for d in dims_jax[: jax_rank - warp_ndim]:
+                            batch_prod *= d
+                        warp_dims = list(dims_jax[jax_rank - warp_ndim :])
+                        warp_dims[0] = warp_dims[0] * batch_prod
+                        collapsed_shape = tuple(warp_dims)
+                    arr = wp.array(ptr=buffer.data, dtype=arg.type.dtype, shape=collapsed_shape, device=device)
                     arg_list.append(arr)
                 else:
                     # scalar argument, get stashed value
@@ -602,8 +620,24 @@ class FfiCallable:
             # pure output args (skip in-out FFI buffers)
             for i, arg in enumerate(self.output_args):
                 buffer = outputs[i + self.num_in_out].contents
-                shape = buffer.dims[: buffer.rank - arg.dtype_ndim]
-                arr = wp.array(ptr=buffer.data, dtype=arg.type.dtype, shape=shape, device=device)
+                # Collapse leading batch dimensions so Warp sees only the declared warp_ndim
+                jax_rank = buffer.rank - arg.dtype_ndim
+                dims_jax = buffer.dims[:jax_rank]
+                warp_ndim = arg.type.ndim
+                if jax_rank < warp_ndim:
+                    raise RuntimeError(
+                        f"Output argument '{arg.name}' expects {warp_ndim} dimension(s) but the passed array has {jax_rank} dimension(s)."
+                    )
+                if jax_rank == warp_ndim:
+                    collapsed_shape = dims_jax
+                else:
+                    batch_prod = 1
+                    for d in dims_jax[: jax_rank - warp_ndim]:
+                        batch_prod *= d
+                    warp_dims = list(dims_jax[jax_rank - warp_ndim :])
+                    warp_dims[0] = warp_dims[0] * batch_prod
+                    collapsed_shape = tuple(warp_dims)
+                arr = wp.array(ptr=buffer.data, dtype=arg.type.dtype, shape=collapsed_shape, device=device)
                 arg_list.append(arr)
 
             # call the Python function with reconstructed arguments
@@ -823,6 +857,229 @@ def register_ffi_callback(name: str, func: Callable, graph_compatible: bool = Tr
     jax.ffi.register_ffi_target(name, ffi_capsule, platform="CUDA")
 
 
+def jax_ad_kernel(kernel, num_outputs=1, static_argnames=None, vmap_method: Optional[str] = "broadcast_all"):
+    """Create a JAX-callable from a Warp kernel with a custom VJP.
+
+    Args:
+        kernel: The Warp kernel to wrap.
+        num_outputs: Number of output arrays produced by the kernel.
+        static_argnames: Optional iterable of argument names that should be treated
+            as static (non-differentiable) in JAX. If None, scalar (non-array) inputs
+            are treated as static by default.
+        vmap_method: How the callback transforms under jax.vmap.
+
+    Returns:
+        A JAX-callable function that can be used inside jax.jit and differentiated with jax.grad.
+    """
+
+    # Infer the original kernel signature (names and annotations)
+    signature = inspect.signature(kernel.func)
+
+    # Positional-or-keyword parameters only
+    parameters = [p for p in signature.parameters.values() if p.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD]
+    parameter_count = len(parameters)
+    num_inputs = parameter_count - num_outputs
+
+    # Determine static argument indices
+    static_args: list[int] = []
+    if static_argnames is not None:
+        name_set = set(static_argnames)
+        for i, p in enumerate(parameters[:num_inputs]):
+            if p.name in name_set:
+                static_args.append(i)
+    else:
+        # Default heuristic: non-array inputs (e.g., scalars) are static
+        for i, p in enumerate(parameters[:num_inputs]):
+            ann = p.annotation
+            try:
+                is_array = isinstance(ann, wp.array)
+            except Exception:
+                is_array = False
+            if not is_array:
+                static_args.append(i)
+
+    static_args = sorted(set(static_args))
+
+    # Forward kernel wrapper: simply launches the kernel
+    def fwd_kernel_wrapper(*args):
+        # args are Warp arrays/scalars reconstructed via FFI in jax_callable
+        wp.launch(kernel, dim=args[0].shape, inputs=args[:-num_outputs], outputs=args[-num_outputs:])
+
+    # Expose the kernel signature to the wrapper so type annotations flow through
+    fwd_kernel_wrapper.__signature__ = signature
+
+    # JAX forward callable using FFI
+    jax_fwd_kernel = jax_callable(fwd_kernel_wrapper, num_outputs=num_outputs, vmap_method=vmap_method)
+
+    # Backward wrapper: launches adjoint with provided output gradients
+    def bwd_kernel_wrapper(*args):
+        # Args: inputs ++ outputs ++ out-grads ++ in-grads-without-statics
+        assert len(args) == 2 * parameter_count - len(static_args)
+
+        inputs = list(args[:num_inputs])
+        outputs = list(args[num_inputs:parameter_count])
+        grad_out = list(args[parameter_count : parameter_count + num_outputs])
+        grad_in = list(args[parameter_count + num_outputs :])
+
+        # Insert placeholders for static arg grads to satisfy Warp's adjoint signature
+        for i in static_args:
+            grad_in.insert(i, inputs[i])
+
+        # Ensure gradient input buffers are zero-initialized before accumulation
+        try:
+            for gi in grad_in:
+                if isinstance(gi, wp.array):
+                    gi.zero_()
+        except Exception:
+            pass
+
+        # Launch adjoint
+        wp.launch(
+            kernel,
+            dim=inputs[0].shape,
+            inputs=inputs,
+            outputs=outputs,
+            adj_inputs=grad_in,
+            adj_outputs=grad_out,
+            adjoint=True,
+        )
+
+    # Build the backward wrapper signature expected by jax_callable
+    # Inputs to the backward function are: inputs, outputs, output grads
+    bwd_input_params = parameters[:num_inputs]
+    bwd_output_params = parameters[num_inputs:parameter_count]
+    bwd_grad_output_params = [
+        inspect.Parameter(
+            p.name + "__vjp",
+            inspect.Parameter.POSITIONAL_OR_KEYWORD,
+            default=p.default,
+            annotation=p.annotation,
+        )
+        for p in bwd_output_params
+    ]
+
+    # Outputs of backward are gradients for differentiable inputs (exclude statics)
+    bwd_grad_input_params = [
+        inspect.Parameter(
+            p.name + "__vjp",
+            inspect.Parameter.POSITIONAL_OR_KEYWORD,
+            default=p.default,
+            annotation=p.annotation,
+        )
+        for p in bwd_input_params
+    ]
+    for i in reversed(static_args):
+        del bwd_grad_input_params[i]
+
+    bwd_signature = bwd_input_params + bwd_output_params + bwd_grad_output_params + bwd_grad_input_params
+    bwd_kernel_wrapper.__signature__ = inspect.Signature(bwd_signature)
+
+    jax_bwd_kernel = jax_callable(
+        bwd_kernel_wrapper,
+        num_outputs=len(bwd_input_params) - len(static_args),
+        vmap_method=vmap_method,
+    )
+
+    # Names of gradient outputs corresponding to differentiable inputs
+    differentiable_input_indices = [i for i in range(num_inputs) if i not in static_args]
+    differentiable_input_names = [parameters[i].name for i in differentiable_input_indices]
+
+    # Define custom VJP wrappers
+    def fwd_function(*args):
+        outputs = jax_fwd_kernel(*args)
+        non_static_inputs = list(args)
+        for i in reversed(static_args):
+            del non_static_inputs[i]
+        if num_outputs == 1:
+            if isinstance(outputs, (list, tuple)):
+                outputs_tuple = (outputs[0],)
+            else:
+                outputs_tuple = (outputs,)
+        else:
+            outputs_tuple = tuple(outputs) if isinstance(outputs, (list, tuple)) else (outputs,)
+        return outputs, (tuple(non_static_inputs), outputs_tuple)
+
+    def bwd_function(*bwd_args):
+        # Layout: (*nondiff_values, residuals, *grad_out)
+        nondiff_vals = list(bwd_args[: len(static_args)])
+        residuals = bwd_args[len(static_args)]
+        grad_out_args = bwd_args[len(static_args) + 1 :]
+
+        non_static_inputs, output_vals_tuple = residuals
+
+        # Reconstruct full input list including static values at correct indices
+        input_vals = list(non_static_inputs)
+        for i, v in zip(static_args, nondiff_vals):
+            input_vals.insert(i, v)
+
+        # Call the adjoint launcher
+        if num_outputs > 1:
+            if len(grad_out_args) == 1 and isinstance(grad_out_args[0], (list, tuple)):
+                grad_out_tuple = tuple(grad_out_args[0])
+            else:
+                grad_out_tuple = tuple(grad_out_args)
+        else:
+            go = grad_out_args[0]
+            if isinstance(go, (list, tuple)):
+                grad_out_tuple = (go[0],)
+            else:
+                grad_out_tuple = (go,)
+        bwd_call_args = list(input_vals) + list(output_vals_tuple) + list(grad_out_tuple)
+
+        # Provide output dims mapping in warp dims (exclude batch and dtype trailing dims)
+        # so that vmap can broadcast batch dims externally.
+        out_dims_map = {}
+        # Build a quick lookup of parameter annotations by name
+        param_ann = {p.name: p.annotation for p in parameters[:num_inputs]}
+        for name, val in zip(differentiable_input_names, non_static_inputs):
+            ann = param_ann.get(name)
+            if ann is None:
+                continue
+            try:
+                is_array_ann = isinstance(ann, wp.array)
+            except Exception:
+                is_array_ann = False
+            if not is_array_ann:
+                continue
+            # Determine dtype_ndim from annotation
+            dtype_ndim = 0
+            try:
+                if hasattr(ann.dtype, "_wp_scalar_type_"):
+                    dtype_ndim = len(ann.dtype._shape_)
+            except Exception:
+                pass
+            warp_ndim = getattr(ann, "ndim", 0)
+            vshape = tuple(val.shape)
+            if warp_ndim == 0:
+                # scalar value: no gradient array expected
+                continue
+            if dtype_ndim > 0:
+                # remove trailing dtype dims then take last warp_ndim dims
+                core_rank = max(0, len(vshape) - dtype_ndim)
+                warp_dims = vshape[max(0, core_rank - warp_ndim) : core_rank]
+            else:
+                warp_dims = vshape[-warp_ndim:]
+            out_dims_map[f"{name}__vjp"] = tuple(warp_dims)
+
+        non_static_input_grads = jax_bwd_kernel(*bwd_call_args, output_dims=out_dims_map)
+        return tuple(non_static_input_grads)
+
+    jax_func = jax.custom_vjp(jax_fwd_kernel, nondiff_argnums=tuple(static_args))
+    jax_func.defvjp(fwd_function, bwd_function)
+
+    # If static scalar args are present, wrap with jitted callable that marks them static
+    if static_args:
+        static_names = [parameters[i].name for i in static_args]
+
+        def _user_callable(*args):
+            return jax_func(*args)
+
+        _user_callable.__signature__ = signature
+        return jax.jit(_user_callable, static_argnames=tuple(static_names))
+
+    return jax_func
+
+
 ###############################################################################
 #
 # Utilities
@@ -859,7 +1116,7 @@ def get_jax_output_type(arg, dims):
         # vector/matrix array
         if ndim == arg.warp_ndim:
             return jax.ShapeDtypeStruct((*dims, *arg.dtype_shape), arg.jax_scalar_type)
-        elif ndim == arg.jax_ndim:
+        elif ndim >= arg.jax_ndim:
             # make sure inner dimensions match
             inner_dims = dims[-arg.dtype_ndim :]
             for i in range(arg.dtype_ndim):
@@ -870,6 +1127,6 @@ def get_jax_output_type(arg, dims):
             raise ValueError(f"Invalid output dimensions for argument '{arg.name}': {dims}")
     else:
         # scalar array
-        if ndim != arg.warp_ndim:
+        if ndim < arg.warp_ndim:
             raise ValueError(f"Invalid output dimensions for argument '{arg.name}': {dims}")
         return jax.ShapeDtypeStruct(dims, arg.jax_scalar_type)
--- a/warp/tests/interop/test_jax.py
+++ b/warp/tests/interop/test_jax.py
@@ -308,6 +308,533 @@ def test_jax_kernel_launch_dims(test, device):
     assert_np_equal(result_2d, expected_2d)
 
 
+@unittest.skipUnless(_jax_version() >= (0, 4, 31), "Jax version too old for FFI custom_vjp")
+def test_jax_ad_kernel_simple(test, device):
+    import jax
+    import jax.numpy as jp
+
+    from warp.jax_experimental.ffi import jax_ad_kernel
+
+    @wp.kernel
+    def scale_sum_square_kernel(a: wp.array(dtype=float), b: wp.array(dtype=float), s: float, c: wp.array(dtype=float)):
+        tid = wp.tid()
+        c[tid] = (a[tid] * s + b[tid]) ** 2.0
+
+    jax_func = jax_ad_kernel(scale_sum_square_kernel, num_outputs=1, static_argnames=("s",), vmap_method="sequential")
+
+    from functools import partial
+
+    @partial(jax.jit, static_argnames=["s"])
+    def loss(a, b, s):
+        out = jax_func(a, b, s)[0]
+        return jp.sum(out)
+
+    n = 16
+    a = jp.arange(n, dtype=jp.float32)
+    b = jp.ones(n, dtype=jp.float32)
+    s = 2.0
+
+    with jax.default_device(wp.device_to_jax(device)):
+        da, db = jax.grad(loss, argnums=(0, 1))(a, b, s)
+
+    # reference gradients
+    # d/da sum((a*s + b)^2) = sum(2*(a*s + b) * s)
+    # d/db sum((a*s + b)^2) = sum(2*(a*s + b))
+    a_np = np.arange(n, dtype=np.float32)
+    b_np = np.ones(n, dtype=np.float32)
+    ref_da = 2.0 * (a_np * s + b_np) * s
+    ref_db = 2.0 * (a_np * s + b_np)
+
+    assert_np_equal(np.asarray(da), ref_da)
+    assert_np_equal(np.asarray(db), ref_db)
+
+
+@unittest.skipUnless(_jax_version() >= (0, 4, 31), "Jax version too old for FFI custom_vjp")
+def test_jax_ad_kernel_multi_output(test, device):
+    import jax
+    import jax.numpy as jp
+
+    from warp.jax_experimental.ffi import jax_ad_kernel
+
+    @wp.kernel
+    def multi_output_kernel(
+        a: wp.array(dtype=float), b: wp.array(dtype=float), s: float, c: wp.array(dtype=float), d: wp.array(dtype=float)
+    ):
+        tid = wp.tid()
+        c[tid] = a[tid] ** 2.0
+        d[tid] = a[tid] * b[tid] * s
+
+    jax_func = jax_ad_kernel(multi_output_kernel, num_outputs=2, static_argnames=("s",))
+
+    def caller(fn, a, b, s):
+        c, d = fn(a, b, s)
+        return jp.sum(c + d)
+
+    @jax.jit
+    def grads(a, b, s):
+        # mark s as static in the inner call via partial to avoid hashing
+        def _inner(a, b, s):
+            return caller(jax_func, a, b, s)
+
+        return jax.grad(lambda a, b: _inner(a, b, 2.0), argnums=(0, 1))(a, b)
+
+    n = 16
+    a = jp.arange(n, dtype=jp.float32)
+    b = jp.ones(n, dtype=jp.float32)
+    s = 2.0
+
+    with jax.default_device(wp.device_to_jax(device)):
+        da, db = grads(a, b, s)
+
+    a_np = np.arange(n, dtype=np.float32)
+    b_np = np.ones(n, dtype=np.float32)
+    # d/da sum(c+d) = 2*a + b*s
+    ref_da = 2.0 * a_np + b_np * s
+    # d/db sum(c+d) = a*s
+    ref_db = a_np * s
+
+    assert_np_equal(np.asarray(da), ref_da)
+    assert_np_equal(np.asarray(db), ref_db)
+
+
+@unittest.skipUnless(_jax_version() >= (0, 4, 31), "Jax version too old for FFI custom_vjp")
+def test_jax_ad_kernel_vec2(test, device):
+    import jax
+    import jax.numpy as jp
+
+    from warp.jax_experimental.ffi import jax_ad_kernel
+
+    @wp.kernel
+    def scale_vec_kernel(a: wp.array(dtype=wp.vec2), s: float, out: wp.array(dtype=wp.vec2)):
+        tid = wp.tid()
+        out[tid] = a[tid] * s
+
+    jax_func = jax_ad_kernel(scale_vec_kernel, num_outputs=1, static_argnames=("s",))
+
+    from functools import partial
+
+    @partial(jax.jit, static_argnames=("s",))
+    def loss(a, s):
+        out = jax_func(a, s)[0]
+        return jp.sum(out)
+
+    n = 10
+    a = jp.arange(n, dtype=jp.float32).reshape((n // 2, 2))
+    s = 3.0
+
+    with jax.default_device(wp.device_to_jax(device)):
+        (da,) = jax.grad(loss, argnums=(0,))(a, s)
+
+    # d/da sum(a*s) = s
+    ref = np.full_like(np.asarray(a), s)
+    assert_np_equal(np.asarray(da), ref)
+
+
+@unittest.skipUnless(_jax_version() >= (0, 4, 31), "Jax version too old for FFI custom_vjp")
+def test_jax_ad_kernel_2d(test, device):
+    import jax
+    import jax.numpy as jp
+
+    from warp.jax_experimental.ffi import jax_ad_kernel
+
+    @wp.kernel
+    def add_one_2d(a: wp.array2d(dtype=float), out: wp.array2d(dtype=float)):
+        i, j = wp.tid()
+        out[i, j] = a[i, j] + 1.0
+
+    jax_func = jax_ad_kernel(add_one_2d, num_outputs=1)
+
+    @jax.jit
+    def loss(a):
+        out = jax_func(a)[0]
+        return jp.sum(out)
+
+    n, m = 8, 6
+    a = jp.arange(n * m, dtype=jp.float32).reshape((n, m))
+
+    with jax.default_device(wp.device_to_jax(device)):
+        (da,) = jax.grad(loss, argnums=(0,))(a)
+
+    ref = np.ones((n, m), dtype=np.float32)
+    assert_np_equal(np.asarray(da), ref)
+
+
+@unittest.skipUnless(_jax_version() >= (0, 4, 31), "Jax version too old for FFI custom_vjp")
+def test_jax_ad_kernel_mat22(test, device):
+    import jax
+    import jax.numpy as jp
+
+    from warp.jax_experimental.ffi import jax_ad_kernel
+
+    @wp.kernel
+    def scale_mat_kernel(a: wp.array(dtype=wp.mat22), s: float, out: wp.array(dtype=wp.mat22)):
+        tid = wp.tid()
+        out[tid] = a[tid] * s
+
+    jax_func = jax_ad_kernel(scale_mat_kernel, num_outputs=1, static_argnames=("s",))
+
+    from functools import partial
+
+    @partial(jax.jit, static_argnames=("s",))
+    def loss(a, s):
+        out = jax_func(a, s)[0]
+        return jp.sum(out)
+
+    n = 12  # must be divisible by 4 for 2x2 matrices
+    a = jp.arange(n, dtype=jp.float32).reshape((n // 4, 2, 2))
+    s = 2.5
+
+    with jax.default_device(wp.device_to_jax(device)):
+        (da,) = jax.grad(loss, argnums=(0,))(a, s)
+
+    ref = np.full((n // 4, 2, 2), s, dtype=np.float32)
+    assert_np_equal(np.asarray(da), ref)
+
+
+@unittest.skipUnless(_jax_version() >= (0, 4, 31), "Jax version too old for FFI custom_vjp")
+def test_jax_ad_kernel_vmap_simple(test, device):
+    import jax
+    import jax.numpy as jp
+
+    from warp.jax_experimental.ffi import jax_ad_kernel
+
+    @wp.kernel
+    def scale_sum_square_kernel(a: wp.array(dtype=float), b: wp.array(dtype=float), s: float, c: wp.array(dtype=float)):
+        tid = wp.tid()
+        c[tid] = (a[tid] * s + b[tid]) ** 2.0
+
+    jax_func = jax_ad_kernel(scale_sum_square_kernel, num_outputs=1, static_argnames=("s",))
+
+    # per-sample loss; close over static scalar s to avoid vmap over statics
+    def per_sample_loss(a, b):
+        out = jax_func(a, b, 2.0)[0]
+        return jp.sum(out)
+
+    B, N = 4, 12
+    a = jp.arange(B * N, dtype=jp.float32).reshape((B, N))
+    b = jp.ones((B, N), dtype=jp.float32)
+
+    with jax.default_device(wp.device_to_jax(device)):
+        da, db = jax.vmap(jax.grad(per_sample_loss, argnums=(0, 1)), in_axes=(0, 0))(a, b)
+
+    a_np = np.arange(B * N, dtype=np.float32).reshape((B, N))
+    b_np = np.ones((B, N), dtype=np.float32)
+    s = 2.0
+    ref_da = 2.0 * (a_np * s + b_np) * s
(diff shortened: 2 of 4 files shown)


warp/tests/interop/test_jax.py
```diff
@@ -308,6 +308,533 @@ def f_2d():
...
+        out = jax_func(a, b, s)[0]
+        return jp.sum(out)
+
+    n = 16
+    a = jp.arange(n, dtype=jp.float32)
+    b = jp.ones(n, dtype=jp.float32)
+    s = 2.0
+
+    with jax.default_device(wp.device_to_jax(device)):
+        da, db = jax.grad(loss, argnums=(0, 1))(a, b, s)
```
johnpjf, line 1068: This takes a grad of jit, but it's more typical to take the jit of grad. Will this work, can we add a test for it?

johnpjf (changes requested):
Thanks for doing this!
I left one comment, but another overall comment is whether we can get a test of this working with multiple devices? I think the older way to do this would be within a `jax.pmap` but there may be newer ways.

mehdiataei:
> Thanks for doing this! I left one comment, but another overall comment is whether we can get a test of this working with multiple devices? I think the older way to do this would be within a `jax.pmap` but there may be newer ways.

Thank John! I have added an example to the doc with shard_map here for the forward pass, which as far as I know is the most up-to-date way of doing this (maybe @nouiz can confirm): https://nvidia.github.io/warp/modules/interoperability.html#distributed-computation

I can add an example to show backward pass works with shard_map as well (I don't see a reason why not).

warp/tests/interop/test_jax.py
mehdiataei: I'll add a test for it

nouiz:
> > Thanks for doing this! I left one comment, but another overall comment is whether we can get a test of this working with multiple devices? I think the older way to do this would be within a `jax.pmap` but there may be newer ways.
> 
> Thank John! I have added an example to the doc with shard_map here for the forward pass, which as far as I know is the most up-to-date way of doing this (maybe @nouiz can confirm): https://nvidia.github.io/warp/modules/interoperability.html#distributed-computation

shard_map is the right manual way of doing sharding.

johnpjf:
FYI, I'm trying this on a two GPU machine and getting an error:
```

@wp.kernel
def triple_kernel(input: wp.array(dtype=float), output: wp.array(dtype=float)):
  tid = wp.tid()  # pytype: disable=module-attr
  output[tid] = 3.0 * input[tid]

class WarpDeferredTest(googletest.TestCase):

  def test_give_me_a_name(self):

    jax_triple = jax_experimental.jax_ad_kernel(triple_kernel)

    def f(x):
      return jax_triple(x)

    @jax.pmap
    def compute_grads(x):
      def loss_fn(x):
        return jnp.sum(jnp.square(f(x)[0]))

      loss, grads = jax.value_and_grad(
          loss_fn, has_aux=False)(x)
      return grads

    x = jnp.broadcast_to(jnp.arange(4, dtype=jnp.float32)[jnp.newaxis], (jax.local_device_count(), 4))
    print(compute_grads(x))



if __name__ == "__main__":
  googletest.main()

```

This fails with:

```
Module __main__ 89dc8ef load on device 'cuda:0' took 318.63 ms  (compiled)
Warp CUDA error 400: invalid resource handle (in function wp_cuda_launch_kernel, third_party/py/warp/native/warp.cu:4061)
```

Notably, if I hide one of the GPUs (with `export CUDA_VISIBLE_DEVICES=0`) so that I only have 1, the tests passes.

mehdiataei:
Hey @johnpjf 
Yes, if the examples are not executed with mpirun, we get that error. I’m not sure what the root cause is, but it’s probably a stream, event, function, or module created on GPU 0 being accidentally used while GPU 1 is current. Running with mpirun fixes it. I made the following test cases to check whether the grad works with shardmap, and they all pass successfully. However, I’m having trouble merging these tests with the current Warp test cases since they rely on mpirun.

```python
import warp as wp
import jax
import jax.numpy as jnp
import numpy as np
from jax.sharding import PartitionSpec as P
from jax.experimental.multihost_utils import process_allgather as allgather
from jax.experimental.shard_map import shard_map
from warp.jax_experimental.ffi import jax_ad_kernel

jax.distributed.initialize()


# Kernels
@wp.kernel
def multiply_by_two_kernel(a_in: wp.array(dtype=wp.float32), a_out: wp.array(dtype=wp.float32)):
    index = wp.tid()
    a_out[index] = a_in[index] * 2.0


@wp.kernel
def scale_sum_square_kernel(
    a: wp.array(dtype=wp.float32), b: wp.array(dtype=wp.float32), s: float, out: wp.array(dtype=wp.float32)
):
    tid = wp.tid()
    out[tid] = (a[tid] * s + b[tid]) ** 2.0


# multi-output kernel
@wp.kernel
def multi_output_kernel(
    a: wp.array(dtype=wp.float32),
    b: wp.array(dtype=wp.float32),
    s: float,
    c: wp.array(dtype=wp.float32),
    d: wp.array(dtype=wp.float32),
):
    tid = wp.tid()
    c[tid] = a[tid] ** 2.0
    d[tid] = a[tid] * b[tid] * s


# JAX callables with AD
jax_mul = jax_ad_kernel(multiply_by_two_kernel, num_outputs=1)
jax_scale = jax_ad_kernel(scale_sum_square_kernel, num_outputs=1, static_argnames=("s",))
jax_mo = jax_ad_kernel(multi_output_kernel, num_outputs=2, static_argnames=("s",))


def warp_multiply(x):
    (out,) = jax_mul(x)
    return out


def warp_scale_sum_square(a, b, s):
    (out,) = jax_scale(a, b, s)
    return out


def warp_multi_output(a, b, s):
    c, d = jax_mo(a, b, s)
    return c, d


def make_sharded_1d(mesh, n):
    sharding = jax.sharding.NamedSharding(mesh, P("x"))
    a = jnp.arange(n, dtype=jnp.float32)
    shape = (n,)
    arrays = [jax.device_put(a[idx], d) for d, idx in sharding.addressable_devices_indices_map(shape).items()]
    return jax.make_array_from_single_device_arrays(shape, sharding, arrays)


def example_mul2_backward(mesh):
    n = jax.device_count() * 5
    x = make_sharded_1d(mesh, n)

    def loss(x):
        y = shard_map(lambda v: warp_multiply(v), mesh=mesh, in_specs=(P("x"),), out_specs=P("x"), check_rep=False)(x)
        return jnp.sum(y)

    g = jax.grad(loss)(x)
    gf = allgather(g)
    if jax.process_index() == 0:
        ref = np.full(n, 2.0, dtype=np.float32)
        np.testing.assert_allclose(np.asarray(gf), ref, rtol=1e-5, atol=1e-6)
        print("mul2: OK")


def example_scale_sum_square_backward(mesh):
    n = jax.device_count() * 6
    a = make_sharded_1d(mesh, n)
    b = make_sharded_1d(mesh, n)
    s = 1.5

    def loss(a, b):
        y = shard_map(
            lambda aa, bb: warp_scale_sum_square(aa, bb, s),
            mesh=mesh,
            in_specs=(P("x"), P("x")),
            out_specs=P("x"),
            check_rep=False,
        )(a, b)
        return jnp.sum(y)

    da, db = jax.grad(loss, argnums=(0, 1))(a, b)
    daf, dbf = allgather(da), allgather(db)
    if jax.process_index() == 0:
        a_np = np.arange(n, dtype=np.float32)
        b_np = np.arange(n, dtype=np.float32)
        ref_da = 2.0 * (a_np * s + b_np) * s
        ref_db = 2.0 * (a_np * s + b_np)
        np.testing.assert_allclose(np.asarray(daf), ref_da, rtol=1e-5, atol=1e-6)
        np.testing.assert_allclose(np.asarray(dbf), ref_db, rtol=1e-5, atol=1e-6)
        print("scale_sum_square: OK")


def example_mul2_with_psum(mesh):
    n = jax.device_count() * 4
    x = make_sharded_1d(mesh, n)

    def body(v):
        local = warp_multiply(v)
        red = jnp.sum(local)
        total = jax.lax.psum(red, "x")
        return local, total

    def loss(x):
        local, total = shard_map(body, mesh=mesh, in_specs=(P("x"),), out_specs=(P("x"), P()), check_rep=False)(x)
        return jnp.sum(local) + total

    g = jax.grad(loss)(x)
    gf = allgather(g)
    if jax.process_index() == 0:
        ref = np.full(n, 4.0, dtype=np.float32)
        np.testing.assert_allclose(np.asarray(gf), ref, rtol=1e-5, atol=1e-6)
        print("mul2 + psum: OK")


def example_psum_mean(mesh):
    n = jax.device_count() * 3
    x = make_sharded_1d(mesh, n)

    def body(v):
        local = warp_multiply(v)
        red = jnp.sum(local)
        total = jax.lax.psum(red, "x")
        mean = total / n
        return local, mean

    def loss(x):
        local, mean = shard_map(body, mesh=mesh, in_specs=(P("x"),), out_specs=(P("x"), P()), check_rep=False)(x)
        return jnp.sum(local) + mean

    g = jax.grad(loss)(x)
    gf = allgather(g)
    if jax.process_index() == 0:
        # d/dx sum(2x) = 2; d/dx mean(2x) = 2/n per element
        ref = np.full(n, 2.0 + 2.0 / n, dtype=np.float32)
        np.testing.assert_allclose(np.asarray(gf), ref, rtol=1e-5, atol=1e-6)
        print("mul2 + psum mean: OK")


def example_multi_output_with_psum(mesh):
    n = jax.device_count() * 5
    a = make_sharded_1d(mesh, n)
    b = make_sharded_1d(mesh, n)
    s = 2.0

    def body(aa, bb):
        c, d = warp_multi_output(aa, bb, s)
        local = c + d
        total = jax.lax.psum(jnp.sum(local), "x")
        return local, total

    def loss(a, b):
        local, total = shard_map(body, mesh=mesh, in_specs=(P("x"), P("x")), out_specs=(P("x"), P()), check_rep=False)(
            a, b
        )
        return jnp.sum(local) + total

    da, db = jax.grad(loss, argnums=(0, 1))(a, b)
    daf, dbf = allgather(da), allgather(db)
    if jax.process_index() == 0:
        a_np = np.arange(n, dtype=np.float32)
        b_np = np.arange(n, dtype=np.float32)
        base_da = 2.0 * a_np + b_np * s
        base_db = a_np * s
        ref_da = 2.0 * base_da
        ref_db = 2.0 * base_db
        np.testing.assert_allclose(np.asarray(daf), ref_da, rtol=1e-5, atol=1e-6)
        np.testing.assert_allclose(np.asarray(dbf), ref_db, rtol=1e-5, atol=1e-6)
        print("multi_output + psum: OK")


def main():
    devices = jax.devices()
    mesh = jax.sharding.Mesh(np.array(devices), "x")
    example_mul2_backward(mesh)
    example_scale_sum_square_backward(mesh)
    example_mul2_with_psum(mesh)
    example_psum_mean(mesh)
    example_multi_output_with_psum(mesh)


if __name__ == "__main__":
    main()
```

You can run them with 
`mpirun -np 2 python test.py`

johnpjf:
Ok, is the intent to fix this multi-GPU issue before checking in?

mehdiataei:
It is fairly common to run multi-GPU workloads with multiple processes in JAX even on a single machine, so I don’t think this is a blocking issue (and this limitation existed before this PR). I have some ideas about the root cause and a potential fix, so let me take a shot at it first.

johnpjf:
Yes, multi-GPU training within a pmap is essentially a requirement for all of our workloads so it would be great to get a fix for this. Thanks

mehdiataei:
@johnpjf Good news. I was able to resolve the CUDA 400 issue by launching FFI on callframe CUDA stream and preload modules across GPUs to prevent per-device race conditions. With these changes in place, your example now runs correctly without requiring the use of `mpirun`. Now any multi-GPU (and distributed multi-GPU with MPI) should be working similar to JAX.

This fix also made it possible to integrate both the `shardmap` and `pmap` test cases directly into `test_jax.py`. 

Can you give it a try? Thanks.

johnpjf:
Thanks, but I don't think it's working for me yet:

```
class WarpDeferredTest(googletest.TestCase):

  def test_give_me_a_name(self):

    jax_triple = jax_experimental.jax_ad_kernel(triple_kernel)

    def f(x):
      return jax_triple(x)

    @jax.pmap
    def compute_grads(x):
      def loss_fn(x):
        return jnp.sum(jnp.square(f(x)[0]))

      loss, grads = jax.value_and_grad(
          loss_fn, has_aux=False)(x)
      return loss, grads

    x = jnp.array(((1.0, 2.0), (1.0, 2.0)))
    print('input: ', x)

    loss, grads = compute_grads(x)
    print('loss: ', loss)
    print('grads: ', grads)

```
Outputs:
```
loss:  [45.  0.]
grads:  [[18. 36.]
 [ 0.  0.]]
```
So it seems like something went wrong on the second GPU.

mehdiataei:
Hmm I can't reproduce it:
```
input:  [[1. 2.]
 [1. 2.]]
Warp 1.8.1 initialized:
   CUDA Toolkit 12.9, Driver 12.9
   Devices:
     "cpu"      : "x86_64"
     "cuda:0"   : "NVIDIA RTX 6000 Ada Generation" (47 GiB, sm_89, mempool enabled)
     "cuda:1"   : "NVIDIA RTX 6000 Ada Generation" (47 GiB, sm_89, mempool enabled)
   CUDA peer access:
     Supported fully (all-directional)
   Kernel cache:
     /home/mehdi/.cache/warp/1.8.1
Module warp.jax_experimental.ffi d07651c load on device 'cuda:0' took 0.41 ms  (cached)
Module test_john 09f893d load on device 'cuda:0' took 223.74 ms  (compiled)
Module warp.jax_experimental.ffi d07651c load on device 'cuda:1' took 0.35 ms  (cached)
Module test_john 09f893d load on device 'cuda:1' took 0.24 ms  (cached)
loss:  [45. 45.]
grads:  [[18. 36.]
 [18. 36.]]
.
----------------------------------------------------------------------
Ran 1 test in 1.761s

OK
```

My code:

```python
import jax
import jax.numpy as jnp
import warp.jax_experimental as jax_experimental
import unittest
import warp as wp


@wp.kernel
def triple_kernel(input: wp.array(dtype=float), output: wp.array(dtype=float)):
  tid = wp.tid()
  output[tid] = 3.0 * input[tid]


class WarpDeferredTest(unittest.TestCase):

  def test_give_me_a_name(self):

    jax_triple = jax_experimental.jax_ad_kernel(triple_kernel)

    def f(x):
      return jax_triple(x)

    @jax.pmap
    def compute_grads(x):
      def loss_fn(x):
        return jnp.sum(jnp.square(f(x)[0]))

      loss, grads = jax.value_and_grad(
          loss_fn, has_aux=False)(x)
      return loss, grads

    x = jnp.array(((1.0, 2.0), (1.0, 2.0)))
    print('input: ', x)

    loss, grads = compute_grads(x)
    print('loss: ', loss)
    print('grads: ', grads)

```

Can you delete the kernel cache folder and try again?

johnpjf:
Tried that, but didn't fix it. Looks like we have a copy of warp in our
monorepo that's from 8/7, is it possible that your change depends on
something since then. I'll try to get ours updated and try again tomorrow.
Thanks again for working on this!

johnpjf:
Ok, turns out I have a bad GPU, pmap works great on a different dual-GPU machine, thanks!

mehdiataei:
Great! I think for JAX to work you need to have identical GPUs (maybe that's the cause)

nouiz:
> Great! I think for JAX to work you need to have identical GPUs (maybe that's the cause)

Exact. You can't mix different kinds of GPUs. This isn't supported.

johnpjf:
Yeah, it wasn't that, I have identical GPUs and have been happily pmap-ing for years and now it just stopped working and giving all zeros for the other shard, even in straight JAX.

mehdiataei:
Added an inverse kinematics example using JAX and Warp, and added changes that enable direct kernel wrapping, reducing the amount of code required when using the interop.
image

btaba:
I tried this PR with this code https://github.com/google-deepmind/mujoco_warp/issues/475#issuecomment-3259898579
and the module loading works fine. But the graph capture doesn't work with pmap. I get

```
E0905 22:27:46.696872   14773 pjrt_stream_executor_client.cc:3008] Execution of replica 0 failed: UNKNOWN: FFI callback error: RuntimeError: Warp CUDA error 904: capturing stream has unjoined work (in function wp_cuda_graph_pause_capture, /builds/omniverse/warp/warp/native/warp.cu:2948)                                                                                                                 
E0905 22:27:46.697047   14776 pjrt_stream_executor_client.cc:3008] Execution of replica 1 failed: UNKNOWN: FFI callback error: RuntimeError: Warp CUDA error 904: capturing stream has unjoined work (in function wp_cuda_graph_pause_capture, /builds/omniverse/warp/warp/nat
ive/warp.cu:2948)     
```

To repro, I modified this [file](https://github.com/google-deepmind/mujoco/blob/main/mjx/mujoco/mjx/third_party/warp/jax_experimental/ffi.py) directly with the implementation in this PR. Any ideas what could be going wrong?

mehdiataei:
@btaba could you share more details on the [WIP]( https://github.com/google-deepmind/mujoco/compare/main...btaba:mujoco:btaba-patch-1?expand=1) ? I played around with this a bit over the weekend and it looks like the issue comes from how modules are being loaded in the “third_party” FFI (in Mujoco). I made some progress but was planning to dig deeper into it this week. Is this the issue you’re working on fixing?

btaba:
> @btaba could you share more details on the [WIP](https://github.com/google-deepmind/mujoco/compare/main...btaba:mujoco:btaba-patch-1?expand=1) ? I played around with this a bit over the weekend and it looks like the issue comes from how modules are being loaded in the “third_party” FFI (in Mujoco). I made some progress but was planning to dig deeper into it this week. Is this the issue you’re working on fixing?

I'm not actively working on it, just wanted to take a stab to try to get to a semi-working state. If you have some cycles, please let us know what you find! There is currently a race condition, and data on one device is corrupt. Other than that, the code at least runs with pmap (occasionally).  

I'm a bit confused by this PR as `__call__` uses the default device `device = wp.get_device("cuda")`, why is it not running on the device from the XLA context (in my branch, I do a bit of a hack to get the actual device from the XLA buffer)? It might make sense to split this PR into one for adjoint support and one for pmap support.

warp/jax_experimental/ffi.py
```diff
@@ -823,6 +905,292 @@ def ffi_callback(call_frame):
...
+
+        non_static_input_grads = jax_bwd_kernel(*bwd_call_args, output_dims=out_dims_map)
+        return tuple(non_static_input_grads)
+
+    jax_func = jax.custom_vjp(jax_fwd_kernel, nondiff_argnums=tuple(static_args))
+    jax_func.defvjp(fwd_function, bwd_function)
+
+    # If static scalar args are present, wrap with jitted callable that marks them static
+    if static_args:
+        static_names = [parameters[i].name for i in static_args]
```
johnpjf, line 1035: I don't understand what this is and why a jit happens?

warp/jax_experimental/ffi.py
```diff
@@ -823,6 +905,292 @@ def ffi_callback(call_frame):
...
+):
+    """Create a JAX-callable from a Warp kernel with a custom VJP.
+
+    This helper wraps an existing Warp kernel in a JAX-compatible callable that
+    supports jit and grad via a custom VJP implemented with the FFI path.
+
+    Args:
+        kernel: The Warp kernel to wrap.
+        num_outputs: Number of output arrays produced by the kernel.
+        static_argnames: Optional iterable of argument names that should be treated
```
johnpjf: It seems there is some confusion in this function between static-for-jit arguments and arguments that you don't want the gradients for. I can have a jax array as an input that is not static-for-jit but that I don't want gradients for. For example, when ray tracing, I don't want gradients of rays, but I don't want them to be static for jit (i.e. they are regular JAX Arrays).

warp/jax_experimental/ffi.py
```diff
@@ -823,6 +905,292 @@ def ffi_callback(call_frame):
...
+            as static (non-differentiable) in JAX. If None, scalar (non-array) inputs
+            are treated as static by default.
+        vmap_method: How the callback transforms under jax.vmap.
+        launch_dim_arg_index: Index (in the kernel's argument list) of the input
```
johnpjf: (Sorry was in wrong place before)
I don't this will work for everyone, for example if I want to do something for every pixel in an RGB image, then I want the kernel to be over the height,width but not channel dimension. Also, this is different than the existing jax_kernel interface. I wonder if it's better to force providing these explicitly?

mehdiataei: I believe we do currently handle this when channels are represented as dtype dims (e.g., wp.vec3) and appear as the trailing axis in the JAX array. I pushed a test to demonstrate it. 

Caveat: this works when channels are encoded as dtype dims (e.g., vec3) and are trailing in the JAX tensor. If someone insists on modeling channels as an extra array axis in the Warp type (e.g., wp.array3d(dtype=float)), then the kernel would launch over H×W×C. If we need to support that representation while still launching over H×W, we can extend jax_ad_kernel to accept an explicit launch_dims or launch_dims_axes parameter and update both forward/backward wrappers accordingly.

Please let me know what you think.

@nouiz @nvlukasz would be great to chime in .

warp/jax_experimental/ffi.py
mehdiataei: hmm I don't think in JAX you can pass jax-arrays as static argnums....? It was the case before and I believe it should the case now as arrays are not hashable. Is this still true @nouiz?

Maybe I don’t fully understand your request. Could you provide an example of something we currently do not support?

warp/jax_experimental/ffi.py
mehdiataei: For static argnums to be registered we need to call jit
https://docs.jax.dev/en/latest/_autosummary/jax.jit.html
This is only used if the user provide the static arguments (or if there are static inputs such as primitive types)

mehdiataei:
> > @btaba could you share more details on the [WIP](https://github.com/google-deepmind/mujoco/compare/main...btaba:mujoco:btaba-patch-1?expand=1) ? I played around with this a bit over the weekend and it looks like the issue comes from how modules are being loaded in the “third_party” FFI (in Mujoco). I made some progress but was planning to dig deeper into it this week. Is this the issue you’re working on fixing?
> 
> I'm not actively working on it, just wanted to take a stab to try to get to a semi-working state. If you have some cycles, please let us know what you find! There is currently a race condition, and data on one device is corrupt. Other than that, the code at least runs with pmap (occasionally).
> 
> I'm a bit confused by this PR as `__call__` uses the default device `device = wp.get_device("cuda")`, why is it not running on the device from the XLA context (in my branch, I do a bit of a hack to get the actual device from the XLA buffer)? It might make sense to split this PR into one for adjoint support and one for pmap support.

Thanks. I did try to fix the issue in Mujoco. I had a bit of success but it seems like there are multiple versions of warp in that repo that needs to be synced based on the same ffi implementation.

The use of wp.get_device("cuda") in the FFI path is that inside the FFI callback we bind the Warp stream to the CUDA stream that XLA passes us and rely on the current CUDA context that XLA sets for the shard. That guarantees the stream/device match across pmap replicas and avoids subtle mismatches that can occur when re-deriving the device from a JAX device handle or from a buffer pointer. Note that the stream itself comes from XLA and aot we preload everything to avoid build races. 

If you try pmap with this Warp branch, it works quite well on multi-GPU devices. All the changes above were made in response to different issues that came up during the implementation.

warp/jax_experimental/ffi.py
johnpjf: I guess I just don't think it's a good idea to hard code that the launch dims must batch one of the input dims, you can definitely think of examples where that wouldn't be true.

warp/jax_experimental/ffi.py
johnpjf: Actually I think my comment was a little wrong, but I do think this code slightly conflates static args for jitting and for custom gradients. I don't really think this code should be concerned with jit-ing at all, static args for this might not be (and usually won't be!) static args for jit.
For specific changes:
Also, you should call these `nondiff_argnums` to match JAX's terminology. https://docs.jax.dev/en/latest/_autosummary/jax.custom_jvp.html
I also wouldn't try to be clever and use heuristics to figure what should or should not be `nondiff_argnums`, that could cause silent bugs where someone forgets to wrap a float in a `jnp.array`.
I think we should also allow for JAX input array's that shouldn't have gradients computed (i.e. as if we had wrapped them in `jax.stop_gradient`). I don't know what to call these or if Warp really supports *not* computing gradients correctly.

btaba:
> > > @btaba could you share more details on the [WIP](https://github.com/google-deepmind/mujoco/compare/main...btaba:mujoco:btaba-patch-1?expand=1) ? I played around with this a bit over the weekend and it looks like the issue comes from how modules are being loaded in the “third_party” FFI (in Mujoco). I made some progress but was planning to dig deeper into it this week. Is this the issue you’re working on fixing?
> > 
> > 
> > I'm not actively working on it, just wanted to take a stab to try to get to a semi-working state. If you have some cycles, please let us know what you find! There is currently a race condition, and data on one device is corrupt. Other than that, the code at least runs with pmap (occasionally).
> > I'm a bit confused by this PR as `__call__` uses the default device `device = wp.get_device("cuda")`, why is it not running on the device from the XLA context (in my branch, I do a bit of a hack to get the actual device from the XLA buffer)? It might make sense to split this PR into one for adjoint support and one for pmap support.
> 
> Thanks. I did try to fix the issue in Mujoco. I had a bit of success but it seems like there are multiple versions of warp in that repo that needs to be synced based on the same ffi implementation.
> 
> The use of wp.get_device("cuda") in the FFI path is that inside the FFI callback we bind the Warp stream to the CUDA stream that XLA passes us and rely on the current CUDA context that XLA sets for the shard. That guarantees the stream/device match across pmap replicas and avoids subtle mismatches that can occur when re-deriving the device from a JAX device handle or from a buffer pointer. Note that the stream itself comes from XLA and aot we preload everything to avoid build races.
> 
> If you try pmap with this Warp branch, it works quite well on multi-GPU devices. All the changes above were made in response to different issues that came up during the implementation.

Thanks @mehdiataei. We vendor the warp ffi code into MuJoCo/MJX for version related reasons. All that should be needed to test, is to dump your `ffi.py` into `mjx/mujoco/mjx/third_party/warp/jax_experimental/ffi.py`. I just tried it right now again and I get this:

```
RuntimeError: Error launching kernel '_subtree_com_init', trying to launch on device='cuda:0', but input array for argument 'body_mass' is on device=cuda:1.
```

What am I missing?

mehdiataei:
> What am I missing?

I digged into how Mujoco uses Warp. It seems like it uses jax_callable function instead of jax_(ad)_kernel. I went ahead and added adjoint support for jax_callable as well (as the first step to fix all the issue). BUT it seems like there is still issues with grabbing  the exact device of the CUDA stream that XLA gives us when using pmap. If you run the pmap tests, sometime they pass sometime they don't!! Unfortunately the XLA Python FFI doesn’t expose the device ordinal; we need a tiny native helper in Warp to query it from the stream/context (I did add this wrapper in C++ but still haven't been able to fix this issue). Do you know of any way to get teh device ordinal?

Another option is to use the Warp kernels as kernels (using jax_ad_kernel) which mitigates this issue all together, it shouldn't require much change on the mujoco side (and it works perfectly well under pmap).

btw, when testing, pls set `wp.config.enable_backward = True` right after importing mjx, because the constraint.py turns it off. Otherwise you see another error if you use jax_ad_kernel with mjx.

mehdiataei:
Another point to use jax_ad_kernel: as you can see I have added automatic diff to jax_callable (as well as allowing passing custom functions for bwd), but it requires global lock to guard Tape usage in auto-grad, which may not be ideal. jax_ad_kernel doesn't have this issue.

btaba:
> > What am I missing?
> 
> I digged into how Mujoco uses Warp. It seems like it uses jax_callable function instead of jax_(ad)_kernel. I went ahead and added adjoint support for jax_callable as well (as the first step to fix all the issue). BUT it seems like there is still issues with grabbing the exact device of the CUDA stream that XLA gives us when using pmap. If you run the pmap tests, sometime they pass sometime they don't!! Unfortunately the XLA Python FFI doesn’t expose the device ordinal; we need a tiny native helper in Warp to query it from the stream/context (I did add this wrapper in C++ but still haven't been able to fix this issue). Do you know of any way to get teh device ordinal?
> 
> Another option is to use the Warp kernels as kernels (using jax_ad_kernel) which mitigates this issue all together, it shouldn't require much change on the mujoco side (and it works perfectly well under pmap).
> 
> btw, when testing, pls set `wp.config.enable_backward = True` right after importing mjx, because the constraint.py turns it off. Otherwise you see another error if you use jax_ad_kernel with mjx.

Thanks for taking a look! This is why I had the hack around device pointers, to get `device_ordinal` directly from the buffer. I'm not sure of a better way. Perhaps someone on the NVIDIA side can help here, in which case, our lives would be made much easier. 

What in jax_ad_kernel makes it work with pmap, but the jax_callable does not? Is there a way to tease apart the pmap support from the grad support? That would be really amazing to get pmap support into the main branch soon-ish. 

@nvlukasz for visibility

johnpjf:
More generally, can jax_ad_kernel just replace jax_kernel.

mehdiataei:
> More generally, can jax_ad_kernel just replace jax_kernel.

Yes. One is enough. I haven't removed it yet but will do after @nvlukasz review and confirm.

mehdiataei:
> > > What am I missing?
> > 
> > 
> > I digged into how Mujoco uses Warp. It seems like it uses jax_callable function instead of jax_(ad)_kernel. I went ahead and added adjoint support for jax_callable as well (as the first step to fix all the issue). BUT it seems like there is still issues with grabbing the exact device of the CUDA stream that XLA gives us when using pmap. If you run the pmap tests, sometime they pass sometime they don't!! Unfortunately the XLA Python FFI doesn’t expose the device ordinal; we need a tiny native helper in Warp to query it from the stream/context (I did add this wrapper in C++ but still haven't been able to fix this issue). Do you know of any way to get teh device ordinal?
> > Another option is to use the Warp kernels as kernels (using jax_ad_kernel) which mitigates this issue all together, it shouldn't require much change on the mujoco side (and it works perfectly well under pmap).
> > btw, when testing, pls set `wp.config.enable_backward = True` right after importing mjx, because the constraint.py turns it off. Otherwise you see another error if you use jax_ad_kernel with mjx.
> 
> Thanks for taking a look! This is why I had the hack around device pointers, to get `device_ordinal` directly from the buffer. I'm not sure of a better way. Perhaps someone on the NVIDIA side can help here, in which case, our lives would be made much easier.
> 
> What in jax_ad_kernel makes it work with pmap, but the jax_callable does not? Is there a way to tease apart the pmap support from the grad support? That would be really amazing to get pmap support into the main branch soon-ish.
> 
> @nvlukasz for visibility

Thanks. The jax_ad_kernel simply launches the adjoint kernels created by Warp, whearas in jax_callable we need to use Tape to record the order of computations (because jax_callable is a generic wrapper for arbitrary annotated Python functions that may contain multiple Warp ops and not a single kernel with a compiled adjoint). 

Why jax_ad_kernel works with pmap is because I think it uses pure kernel launches. I'm doing some debugging here now.

Removing the adjoint part really won't make any difference for the current issue we have for the pmap support. We'll have the exact same issue for forward as well. If we fix it the adjoint computation will just work without any issues. 

The solution is probably on the C++/CUDA world. I am looking into how to get the device ordinal. There is hope.

nvlukasz:
> > More generally, can jax_ad_kernel just replace jax_kernel.
> 
> Yes. One is enough. I haven't removed it yet but will do after @nvlukasz review and confirm.

Agreed, we should only keep one if possible.

nvlukasz:
> The solution is probably on the C++/CUDA world. I am looking into how to get the device ordinal. There is hope.

So we can't get the CUDA device from JAX, via `get_jax_device()` or similar?

I suppose it's not possible to wrap code in Python contexts like `jax.default_device` or `wp.ScopedDevice`?

I don't see a way to get the device from XLA, but we can get the stream. Then we can use `cuStreamGetCtx()` under the hood to figure out the device. IIRC, one caveat with that function is that it's not allowed during graph capture, so I've been avoiding it in general. We'd need to be careful there.

We can also query the current CUDA context in Warp, we have some functions for that already. But if JAX does not set it, then this approach won't work (same with using the special device `"cuda"`).

Another approach is to get the device from the JAX arrays or buffers, and assume that the launch is to take place on the same device. There is a way to query the CUDA context associated with a pointer (though I think it has the same caveats with graph capture as `cuStreamGetCtx`).

If none of these work, I'm slightly at a loss. Maybe it's a question for JAX or XLA folks?

nvlukasz:
> What in jax_ad_kernel makes it work with pmap, but the jax_callable does not? Is there a way to tease apart the pmap support from the grad support? That would be really amazing to get pmap support into the main branch soon-ish.

I agree that it would be good to split up the PR into separate ones for pmap and AD support. @mehdiataei how painful would that be for you?

mehdiataei:
Ok… finally I was able to fix the device ordinal match by adding new C++ APIs to Warp (it wasn't an easy fix, it took a bit of trial and error). `jax_callable` now fully supports both `pmap` for forward and adjoint, and I tested them extensively (many test cases are added). I did the tests in a long loop as well to avoid false positives due to randomness.

I am now working on fixing this on the Mujoco side, but the graph capture in Mujoco is a bit convoluted and relies heavily on various capture statements, which I think are causing issues.

P.S. Although I don’t think removing adjoint will help fixing the current issue, disabling it is straightforward: just avoid passing the backward function and set auto_grad to false for jax_callable (which is the only thing used on mujoco side). If reviewing is simpler with it separated, we can do that, but I don’t think it will help with the problem.

mehdiataei:
Good news! I think I fixed mojuco issue as well (Seems like the backward is not enabled for some of the functions which caused some issues). Summary:

Fixed two issues that broke pmap+vmap with MJX Warp FFI:
  1) Removed Warp CUDA graph capture inside the FFI callback when running on the XLA stream.
  2) Temporarily disabled Warp backward codegen during the FFI callback.

this helps because the original FFI wrapper tried to capture a CUDA graph on the XLA stream and also called Warp APIs that can insert conditional nodes (`wp.capture_if`). CUDA’s capture is thread-local and stream-bound, so mixing XLA’s external stream capture with Warp’s internal capture paths caused “Graph capture is not active on this stream” and related insert/pause errors under parallel execution. Separately, Warp was JIT-building adjoint kernels (enable_backward=True) for collision/narrowphase code during the callback. Under pmap/vmap and without a stable capture context, that triggered NVRTC compile issues and module load mismatches

So by wrapping the call with a small context that sets `wp.config.enable_backward = False` for the duration of the callback, guarded by a lock, I prevent on-the-fly generation of backward kernels while still allowing forward kernels to compile and run. Non-XLA stream path retains Warp’s optional graph capture.

Notes
  - All adjoint computations are unaffected and all tests pass becuase already built kernels (including their adjoints) are unaffected.

  ```
  ----------------------------------------------------------------------
Ran 64 tests in 120.365s

OK
```
  - If/when gradients are needed through these kernels, we can revisit enabling backward selectively or prebuilding the required adjoint kernels outside the callback so they don’t compile during pmap/vmap execution. If a kernel is first-time compiled while inside the callback, it will be compiled without an adjoint. If you later want gradients through that specific kernel via Warp’s Tape, you’ll need to prebuild it with backward enabled (or clear cache and rebuild) outside the callback.

To test yourself copy the content of `ffi.py` to the mujoco and test with the following:

**IMPORTANT: make sure to rebuild Warp from source since changes were made in the C++ code.**

@btaba example:

```python
import mujoco
from mujoco import mjx
import jax


xml = """
<mujoco>
  <worldbody>
    <light name="top" pos="0 0 1"/>
    <body name="floor" pos="0 0 0">
      <geom name="floor" type="plane" size="1 1 0.1" rgba=".8 .8 .8 1"/>
    </body>
    <body name="sphere" pos="0 0 1">
      <joint type="free"/>
      <geom name="sphere" type="sphere" size="0.1" rgba="0 0 1 1"/>
    </body>
  </worldbody>
</mujoco>
"""

m = mujoco.MjModel.from_xml_string(xml)
mx = mjx.put_model(m, impl='warp')

def make_data(rng):
  dx = mjx.make_data(m, impl='warp', nconmax=100, njmax=10)
  dx = dx.replace(qpos=jax.random.uniform(rng, mx.nq))
  return dx

device_count = jax.local_device_count()
assert device_count > 1
batch_size = 16

per_device_keys = jax.random.split(jax.random.PRNGKey(0), device_count)
dx_batch = jax.pmap(lambda k: jax.vmap(make_data)(jax.random.split(k, batch_size)))(per_device_keys)
assert dx_batch.qpos.shape == (device_count, batch_size, m.nq)

out = jax.pmap(jax.vmap(mjx.step, in_axes=(None, 0)), in_axes=(None, 0))(mx, dx_batch)
print(out.time)
assert (out.time == 0.002).all()
```

output:
```
Warp 1.8.1 initialized:
   CUDA Toolkit 12.9, Driver 12.9
   Devices:
     "cpu"      : "x86_64"
     "cuda:0"   : "NVIDIA RTX 6000 Ada Generation" (47 GiB, sm_89, mempool enabled)
     "cuda:1"   : "NVIDIA RTX 6000 Ada Generation" (47 GiB, sm_89, mempool enabled)
   CUDA peer access:
     Supported fully (all-directional)
   Kernel cache:
     /home/mehdi/.cache/warp/1.8.1
Module mujoco.mjx.third_party.mujoco_warp._src.smooth 0c1050e load on device 'cuda:0' took 8.68 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver dc593c3 load on device 'cuda:0' took 0.30 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver._nxn_broadphase 2441629 load on device 'cuda:0' took 0.27 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.collision_primitive 00c6f22 load on device 'cuda:0' took 0.31 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.constraint 7a9fd39 load on device 'cuda:0' took 1.39 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.forward 76a01ca load on device 'cuda:0' took 0.61 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.passive 4223120 load on device 'cuda:0' took 0.54 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.support aa34105 load on device 'cuda:0' took 0.34 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.smooth._tile_cholesky_factorize_solve a1700c8 load on device 'cuda:0' took 26.94 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver b37b8b0 load on device 'cuda:0' took 1.16 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.support.mul_m_dense 397bd0e load on device 'cuda:0' took 21.33 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_constraint_gauss_cost cd535c8 load on device 'cuda:0' took 0.26 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_gradient_cholesky 4039252 load on device 'cuda:0' took 28.84 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver.linesearch_jv_fused 37529ee load on device 'cuda:0' took 0.31 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.forward._tile_euler_dense 756e4cd load on device 'cuda:0' took 29.54 ms  (cached)
Module mujoco.mjx.warp.ffi c043961 load on device 'cuda:0' took 0.28 ms  (cached)
Module mujoco.mjx.warp.ffi c043961 load on device 'cuda:1' took 0.28 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.smooth 0c1050e load on device 'cuda:1' took 13.61 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver dc593c3 load on device 'cuda:1' took 0.65 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver._nxn_broadphase 2441629 load on device 'cuda:1' took 0.61 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.collision_primitive 00c6f22 load on device 'cuda:1' took 0.31 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.constraint 7a9fd39 load on device 'cuda:1' took 1.56 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.forward 76a01ca load on device 'cuda:1' took 0.72 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.passive 4223120 load on device 'cuda:1' took 0.53 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.support aa34105 load on device 'cuda:1' took 0.33 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.smooth._tile_cholesky_factorize_solve a1700c8 load on device 'cuda:1' took 29.98 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver b37b8b0 load on device 'cuda:1' took 1.15 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.support.mul_m_dense 397bd0e load on device 'cuda:1' took 22.49 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_constraint_gauss_cost cd535c8 load on device 'cuda:1' took 0.24 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_gradient_cholesky 4039252 load on device 'cuda:1' took 29.70 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.solver.linesearch_jv_fused 37529ee load on device 'cuda:1' took 0.33 ms  (cached)
Module mujoco.mjx.third_party.mujoco_warp._src.forward._tile_euler_dense 756e4cd load on device 'cuda:1' took 31.55 ms  (cached)
[[0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002
  0.002 0.002 0.002 0.002]
 [0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002
  0.002 0.002 0.002 0.002]]
 ```

Enjoy!!

**Comment by chaserileyroberts** [comment]
> I don't see a way to get the device from XLA.

There is a way actually. We should be able to use `XLA_FFI_DeviceOrdinal_Get` from here: https://github.com/openxla/xla/blob/97b0e76e6c706fe0a7b22e02a9d8e6949b417ee9/xla/ffi/ffi_api.cc#L632

We would need to add it the XLA_FFI_Api list [here](https://github.com/NVIDIA/warp/blob/46dc8b6a17ae9ec381e8a133420e1bba8904ec00/warp/jax_experimental/xla_ffi.py#L406) and plumb it through. That might be safer/easier than trying to get it from the stream or array pointers.

**Comment by nvlukasz** [comment]
> > I don't see a way to get the device from XLA.
> 
> There is a way actually. We should be able to use `XLA_FFI_DeviceOrdinal_Get` from here: https://github.com/openxla/xla/blob/97b0e76e6c706fe0a7b22e02a9d8e6949b417ee9/xla/ffi/ffi_api.cc#L632
> 
> We would need to add it the XLA_FFI_Api list [here](https://github.com/NVIDIA/warp/blob/46dc8b6a17ae9ec381e8a133420e1bba8904ec00/warp/jax_experimental/xla_ffi.py#L406) and plumb it through. That might be safer/easier than trying to get it from the stream or array pointers.

Ah, that sounds like just what the doctor ordered. Thanks for pointing it out, @chaserileyroberts!

@mehdiataei FYI

**Comment by mehdiataei** [comment]
Thanks @chaserileyroberts . We can definitely switch to that when the API is exposed. Until then the pointer monkey patch should be workable.

**Comment by chaserileyroberts** [comment]
@mehdiataei it's a relatively small change so I'll have a PR up for the integration in a bit.

**Comment by btaba** [comment]
> Good news! I think I fixed mojuco issue as well (Seems like the backward is not enabled for some of the functions which caused some issues). Summary:
> 
> Fixed two issues that broke pmap+vmap with MJX Warp FFI:
> 
> 1. Removed Warp CUDA graph capture inside the FFI callback when running on the XLA stream.
> 2. Temporarily disabled Warp backward codegen during the FFI callback.
> 
> this helps because the original FFI wrapper tried to capture a CUDA graph on the XLA stream and also called Warp APIs that can insert conditional nodes (`wp.capture_if`). CUDA’s capture is thread-local and stream-bound, so mixing XLA’s external stream capture with Warp’s internal capture paths caused “Graph capture is not active on this stream” and related insert/pause errors under parallel execution. Separately, Warp was JIT-building adjoint kernels (enable_backward=True) for collision/narrowphase code during the callback. Under pmap/vmap and without a stable capture context, that triggered NVRTC compile issues and module load mismatches
> 
> So by wrapping the call with a small context that sets `wp.config.enable_backward = False` for the duration of the callback, guarded by a lock, I prevent on-the-fly generation of backward kernels while still allowing forward kernels to compile and run. Non-XLA stream path retains Warp’s optional graph capture.
> 
> Notes
> 
> * All adjoint computations are unaffected and all tests pass becuase already built kernels (including their adjoints) are unaffected.
> 
> ```
> ----------------------------------------------------------------------
> Ran 64 tests in 120.365s
> 
> OK
> ```
> 
> * If/when gradients are needed through these kernels, we can revisit enabling backward selectively or prebuilding the required adjoint kernels outside the callback so they don’t compile during pmap/vmap execution. If a kernel is first-time compiled while inside the callback, it will be compiled without an adjoint. If you later want gradients through that specific kernel via Warp’s Tape, you’ll need to prebuild it with backward enabled (or clear cache and rebuild) outside the callback.
> 
> To test yourself copy the content of `ffi.py` to the mujoco and test with the following:
> 
> **IMPORTANT: make sure to rebuild Warp from source since changes were made in the C++ code.**
> 
> @btaba example:
> 
> ```python
> import mujoco
> from mujoco import mjx
> import jax
> 
> 
> xml = """
> <mujoco>
>   <worldbody>
>     <light name="top" pos="0 0 1"/>
>     <body name="floor" pos="0 0 0">
>       <geom name="floor" type="plane" size="1 1 0.1" rgba=".8 .8 .8 1"/>
>     </body>
>     <body name="sphere" pos="0 0 1">
>       <joint type="free"/>
>       <geom name="sphere" type="sphere" size="0.1" rgba="0 0 1 1"/>
>     </body>
>   </worldbody>
> </mujoco>
> """
> 
> m = mujoco.MjModel.from_xml_string(xml)
> mx = mjx.put_model(m, impl='warp')
> 
> def make_data(rng):
>   dx = mjx.make_data(m, impl='warp', nconmax=100, njmax=10)
>   dx = dx.replace(qpos=jax.random.uniform(rng, mx.nq))
>   return dx
> 
> device_count = jax.local_device_count()
> assert device_count > 1
> batch_size = 16
> 
> per_device_keys = jax.random.split(jax.random.PRNGKey(0), device_count)
> dx_batch = jax.pmap(lambda k: jax.vmap(make_data)(jax.random.split(k, batch_size)))(per_device_keys)
> assert dx_batch.qpos.shape == (device_count, batch_size, m.nq)
> 
> out = jax.pmap(jax.vmap(mjx.step, in_axes=(None, 0)), in_axes=(None, 0))(mx, dx_batch)
> print(out.time)
> assert (out.time == 0.002).all()
> ```
> 
> output:
> 
> ```
> Warp 1.8.1 initialized:
>    CUDA Toolkit 12.9, Driver 12.9
>    Devices:
>      "cpu"      : "x86_64"
>      "cuda:0"   : "NVIDIA RTX 6000 Ada Generation" (47 GiB, sm_89, mempool enabled)
>      "cuda:1"   : "NVIDIA RTX 6000 Ada Generation" (47 GiB, sm_89, mempool enabled)
>    CUDA peer access:
>      Supported fully (all-directional)
>    Kernel cache:
>      /home/mehdi/.cache/warp/1.8.1
> Module mujoco.mjx.third_party.mujoco_warp._src.smooth 0c1050e load on device 'cuda:0' took 8.68 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver dc593c3 load on device 'cuda:0' took 0.30 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver._nxn_broadphase 2441629 load on device 'cuda:0' took 0.27 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.collision_primitive 00c6f22 load on device 'cuda:0' took 0.31 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.constraint 7a9fd39 load on device 'cuda:0' took 1.39 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.forward 76a01ca load on device 'cuda:0' took 0.61 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.passive 4223120 load on device 'cuda:0' took 0.54 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.support aa34105 load on device 'cuda:0' took 0.34 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.smooth._tile_cholesky_factorize_solve a1700c8 load on device 'cuda:0' took 26.94 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver b37b8b0 load on device 'cuda:0' took 1.16 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.support.mul_m_dense 397bd0e load on device 'cuda:0' took 21.33 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_constraint_gauss_cost cd535c8 load on device 'cuda:0' took 0.26 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_gradient_cholesky 4039252 load on device 'cuda:0' took 28.84 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver.linesearch_jv_fused 37529ee load on device 'cuda:0' took 0.31 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.forward._tile_euler_dense 756e4cd load on device 'cuda:0' took 29.54 ms  (cached)
> Module mujoco.mjx.warp.ffi c043961 load on device 'cuda:0' took 0.28 ms  (cached)
> Module mujoco.mjx.warp.ffi c043961 load on device 'cuda:1' took 0.28 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.smooth 0c1050e load on device 'cuda:1' took 13.61 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver dc593c3 load on device 'cuda:1' took 0.65 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.collision_driver._nxn_broadphase 2441629 load on device 'cuda:1' took 0.61 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.collision_primitive 00c6f22 load on device 'cuda:1' took 0.31 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.constraint 7a9fd39 load on device 'cuda:1' took 1.56 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.forward 76a01ca load on device 'cuda:1' took 0.72 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.passive 4223120 load on device 'cuda:1' took 0.53 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.support aa34105 load on device 'cuda:1' took 0.33 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.smooth._tile_cholesky_factorize_solve a1700c8 load on device 'cuda:1' took 29.98 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver b37b8b0 load on device 'cuda:1' took 1.15 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.support.mul_m_dense 397bd0e load on device 'cuda:1' took 22.49 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_constraint_gauss_cost cd535c8 load on device 'cuda:1' took 0.24 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver.update_gradient_cholesky 4039252 load on device 'cuda:1' took 29.70 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.solver.linesearch_jv_fused 37529ee load on device 'cuda:1' took 0.33 ms  (cached)
> Module mujoco.mjx.third_party.mujoco_warp._src.forward._tile_euler_dense 756e4cd load on device 'cuda:1' took 31.55 ms  (cached)
> [[0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002
>   0.002 0.002 0.002 0.002]
>  [0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002 0.002
>   0.002 0.002 0.002 0.002]]
> ```
> 
> Enjoy!!

Hi @mehdiataei I just tried this locally. I pulled your branch, built warp from source, copied over `ffi.py` to the third_party mujoco copy, and ran the example. I get a bunch of "operation not permitted...wp_free_device_async" due to a `wp_memcpy_d2d`. 

Here's the last error `XlaRuntimeError: INTERNAL: CUDA error: Failed to end stream capture: CUDA_ERROR_STREAM_CAPTURE_INVALIDATED: operation failed due to a previous error during capture: while running replica 1 and partition 0 of a replicated computation (other replicas may have failed as well)`

Are there any gotchas to keep in mind here?

**Comment by mehdiataei** [comment]
Hey @btaba let me do a clean setup from scratch and I'll get back to you.

**Comment by btaba** [comment]
@mehdiataei I'm also a bit confused, the graph_mode is no longer respected in `ffi_callback` https://github.com/NVIDIA/warp/blob/6673762c5f6883c50a260b4c8d72968054d14ac4/warp/jax_experimental/ffi.py#L729-L738

and the JAX graph mode is removed?

**Comment by mehdiataei** [comment]
@btaba 
```
uv venv
source .venv/bin/activate
git clone https://github.com/google-deepmind/mujoco.git
cd mujoco/mjx
uv pip install -e .
uv pip install jax["cuda12"]
uv pip install -e <ADDRESS_TO_WARP>/warp/ # Note that we need to build warp first build_lib.py
# Copy the content of ffi.py into ffi.py file in mujoco's third_party/warp
cd ../../
python <example>.py
```
the above worked for me can you try?

btaba:
Thanks @mehdiataei , I just tried with the latest branch but still get zeros on the second device. I'll have to try on another machine; do you have any ideas otherwise?

nvlukasz:
> Thanks @mehdiataei , I just tried with the latest branch but still get zeros on the second device. I'll have to try on another machine; do you have any ideas otherwise?

@btaba @mehdiataei, I have a pmap solution based on [GH-976](https://github.com/NVIDIA/warp/pull/976) from @chaserileyroberts. Thanks Chase! It's currently under review by the Warp team, but I expect it to be merged shortly.

I had to add some additional thread safety, but the Mujoco example works fine now. That should take care of pmap, so once that's merged, @mehdiataei you can modify this PR to only add the autodiff functionality.

Thank you all for your contributions and patience!

mehdiataei:
> Thanks @mehdiataei , I just tried with the latest branch but still get zeros on the second device. I'll have to try on another machine; do you have any ideas otherwise?

hmm no idea. I did follow thoese steps again and it works fine with me. Is it working correctly with pure jax and pmap?

@nvlukasz That's great! Yes. I can merge those changes and add the adjoint stuff.

nvlukasz:
> @nvlukasz That's great! Yes. I can merge those changes and add the adjoint stuff.

Alright, the new pmap support is now merged (https://github.com/NVIDIA/warp/commit/599ee98c9c1f4be1f37a18279335f3ca954b1d48)

warp/examples/optim/example_inverse_kinematics_jax.py
shi-eric, line 1: This example cannot be added since `warp.sim` is now removed from the `main` branch. Maybe it can be contributed to `newton` instead.

mehdiataei: Yeah just removed it. I'm making some final changes and they will just squash the commits.

shi-eric:
FYI @mehdiataei, I'm planning to make the release candidate for Warp v1.10 sometime next Friday, October 17th, so the changes in this pull request need to be merged into the `main` branch by then to be included in the release. Please work with @nvlukasz to resolve any outstanding issues.

nvlukasz:
This is now merged in https://github.com/NVIDIA/warp/commit/b66f253c39762c48b5eb4a8a1abf077677121da2. Thank you @mehdiataei for this great contribution!

PR #971, opened 2025-09-18
Add type imports and overloads for function decorators

liblaf:
## Description

This PR improves the type annotations for Warp's function decorators to provide better static type checking support and IDE experience for users.

### Limitations

There are still many functions that are not annotated.

### DIFF
--- a/warp/__init__.py
+++ b/warp/__init__.py
@@ -316,5 +316,6 @@ from warp.builtins import static as static
 from warp.math import *
 
 from . import config as config
+from . import types as types
 
 __version__ = config.version
--- a/warp/context.py
+++ b/warp/context.py
@@ -48,6 +48,7 @@ from typing import (
     get_args,
     get_origin,
 )
+from typing import overload as _overload
 
 import numpy as np
 
@@ -849,8 +850,15 @@ class Kernel:
 # ----------------------
 
 
-# decorator to register function, @func
-def func(f: Callable | None = None, *, name: str | None = None):
+@_overload
+def func(f: Callable, *, name: str | None = None) -> Function: ...
+@_overload
+def func(*, name: str | None = None) -> Callable[[Callable], Function]: ...
+def func(f: Callable | None = None, *, name: str | None = None) -> Any:
+    """
+    Decorator to register function, @func
+    """
+
     def wrapper(f, *args, **kwargs):
         if name is None:
             key = warp.codegen.make_full_qualified_name(f)
@@ -1066,12 +1074,17 @@ def func_replay(forward_fn):
     return wrapper
 
 
+@_overload
 def kernel(
-    f: Callable | None = None,
-    *,
-    enable_backward: bool | None = None,
-    module: Module | Literal["unique"] | None = None,
-):
+    f: Callable, *, enable_backward: bool | None = None, module: Module | Literal["unique"] | None = None
+) -> Kernel: ...
+@_overload
+def kernel(
+    *, enable_backward: bool | None = None, module: Module | Literal["unique"] | None = None
+) -> Callable[[Callable], Kernel]: ...
+def kernel(
+    f: Callable | None = None, *, enable_backward: bool | None = None, module: Module | Literal["unique"] | None = None
+) -> Any:
     """
     Decorator to register a Warp kernel from a Python function.
     The function must be defined with type annotations for all arguments.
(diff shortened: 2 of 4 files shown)


warp/__init__.py
```diff
@@ -316,5 +316,6 @@
 from warp.math import *
 
 from . import config as config
+from . import types as types
```
christophercrouzet: We consider anything in `__init__.py` to be part of the public API, and the `types` module is not something that we would like to expose as a whole.

warp/context.py
```diff
@@ -849,8 +849,15 @@ def __call__(self, *args, **kwargs):
...
-def func(f: Callable | None = None, *, name: str | None = None):
+@_overload
+def func(f: Callable, *, name: str | None = None) -> Function: ...
+@_overload
+def func(*, name: str | None = None) -> Callable[[Callable], Function]: ...
+def func(f: Callable | None = None, *, name: str | None = None) -> Any:
+    """
+    Decorator to register function, @func
+    """
+
```
christophercrouzet: I'm afraid this is fairly confusing/overwhelming to have such stubs inline with the source code. If we want these, we would need to find a way to add them to `__init__.pyi` using the `context.export_stubs()` function.

liblaf: If I understand correctly, inline type annotations are the preferred way to provide type information according to Python [typing documentation](https://typing.python.org/en/latest/guides/libraries.html#how-to-provide-type-annotations):

> [PEP 561](https://peps.python.org/pep-0561/) documents several ways type information can be provided for a library:
>
> - inline type annotations (preferred)
> - type stub files included in the package
> - a separate companion type stub package
> - type stubs in the typeshed repository
>
> Inline type annotations simply refers to the use of annotations within your `.py` files. In contrast, with type stub files, type information lives in separate `.pyi` files; ...

On the other hand, `__init__.pyi` imports these functions directly from `warp.context`:

https://github.com/NVIDIA/warp/blob/46dc8b6a17ae9ec381e8a133420e1bba8904ec00/warp/__init__.pyi#L175-L182

and `context.py` does not have a corresponding `context.pyi` stub file, which means that the stubs exposed by `__init__.pyi` comes directly from the inline type annotations in source code.

If we don't want inline type annotations, I think we could add a `context.pyi`. That way we don't need to introduce additional complex logic in `export_stubs()`. But then we would need to annotate all public functions in `context.pyi` for static type checkers to work.

warp/__init__.py
liblaf: When I was reading documentation, I noticed that `wp.types` was used [here](https://nvidia.github.io/warp/modules/runtime.html#matrices), [here](https://nvidia.github.io/warp/codegen.html#customize-dimensions) and [here](https://github.com/NVIDIA/warp/blob/46dc8b6a17ae9ec381e8a133420e1bba8904ec00/warp/examples/fem/utils.py#L365).

And comment here:

https://github.com/NVIDIA/warp/blob/41239153947a3710ad234f8544bd7e4390ca816a/warp/__init__.py#L146-L148

is encouraging `wp.types.vector`, `wp.types.matrix`.

I assumed that `wp.types` was part of the public API. If `wp.types` is not, I'll remove it from `__init__.pyi`.

christophercrouzet: Thanks @liblaf, these calls to `wp.types.vector()` and `wp.types.matrix()` have since then been made available with `wp.vec()` and `wp.mat()` respectively, so the documentation needs to be updated. I've just opened GH-986 to do just that.

The call to `wp.types.type_size_in_bytes()` is fine since it's done from a Warp module, and Warp modules have access to all private API. That being said, if it would make sense to have some of these functions available from the public API, we're open to do so by exposing these individually, as opposed to promoting the whole `types` module.

warp/context.py
christophercrouzet: If there's any way we could generate these stubs automatically from the definition of these functions, then I think we'd rather take this path since it would be less error-prone and would require less maintenance in the long term. But we haven't looked into what is feasible/expected here, so I can't really comment on this.

We can always do this work later and focus this MR on the changes you've made to `def struct` and `def overload`?

christophercrouzet: Hi @liblaf, what do you think of my suggestion or repurposing this MR to only keep the type annotations to `def struct` and `def overload`, and to remove the other changes for now?

If so, we could include this MR in our next release if it's ready by next Friday.

warp/__init__.py
liblaf: Thanks for your information. I'll remove it.

warp/context.py
liblaf: Sorry for the late reply. I agree. I'll remove the other changes.

PR #976, opened 2025-09-24
Add device ordinal XLA plumbing

chaserileyroberts:
## Description

This PR fixes failures related to using Warp within a `jax.pmap` or a `jax.shard_map`. The main issue was that JAX would initiate the Warp callback from multiple threads targeting different devices, but the device Warp would target was always the default one. To fix this, we hookup `XLA_FFI_DeviceOrdinal_Get` to `xla_ffi.py`, and plumb that value to the warp device context so that JAX and Warp would each target the same device on each callback. 

Since the device Warp uses is dependent on a global context, JAX calling the FFI callable from multiple threads would smash this value and cause errors. To fix this, we include a threading lock before entering the context. We may consider making the device context value a thread local instead of a program global one instead.  

- [ x ] All commits are [signed-off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) to indicate that your contribution adheres to the [Developer Certificate of Origin](https://developercertificate.org/) requirements
- [ x ] Necessary tests have been added
- [ x ] Documentation is up-to-date
- [ x ] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `__init__.pyi`, `functions.rst`)
- [ x ] Code passes formatting and linting checks with `pre-commit run -a`

### DIFF
--- a/warp/jax_experimental/ffi.py
+++ b/warp/jax_experimental/ffi.py
@@ -23,10 +23,11 @@ import jax
 
 import warp as wp
 from warp.codegen import get_full_arg_spec, make_full_qualified_name
-from warp.jax import get_jax_device
 from warp.types import array_t, launch_bounds_t, strides_from_shape, type_to_warp
 
 from .xla_ffi import *
+
+WARP_DEVICE_CONTEXT_LOCK = threading.Lock()
 
 
 def check_jax_version():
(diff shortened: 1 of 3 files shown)


nvlukasz (commented):
Thank you @chaserileyroberts, this is great. I will get it merged shortly.

PR #990, opened 2025-09-28
Remove call to deprecated overload of createTargetMachine

boomanaiden154:
The overload using the string version of the triple is deprecated and will likely be removed soon. Update the call site to use the version with llvm::Triple for versions of LLVM newer than v22.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved build incompatibilities with LLVM 22+, ensuring projects compile successfully on newer toolchains and maintain correct data layout across supported versions.
  * Improves cross-version stability without altering public behavior.

* **Chores**
  * Updated compiler integration to maintain forward compatibility with recent LLVM releases, reducing the need for manual patches on modern environments.

### DIFF
--- a/warp/native/clang/clang.cpp
+++ b/warp/native/clang/clang.cpp
@@ -271,7 +271,11 @@ WP_API int wp_compile_cpp(const char* cpp_src, const char *input_file, const cha
         target_options.AllowFPOpFusion = llvm::FPOpFusion::Strict;
     llvm::Reloc::Model relocation_model = llvm::Reloc::PIC_;  // Position Independent Code
     llvm::CodeModel::Model code_model = llvm::CodeModel::Large;  // Don't make assumptions about displacement sizes
+    #if LLVM_VERSION_MAJOR >= 22
+    llvm::TargetMachine* target_machine = target->createTargetMachine(llvm::Triple(target_triple), CPU, features, target_options, relocation_model, code_model);
+    #else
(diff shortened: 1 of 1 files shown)


c0d1f1ed:
Thanks for the patch! Out of curiosity, what makes you want to build Warp with a newer version of LLVM?

I had a look at it myself recently, in the hopes that it would address a compilation bug we've been running into. I found that there are more issues than just the `Triple` argument for `createTargetMachine()` though. I'll create an MR soon.

boomanaiden154:
> Thanks for the patch! Out of curiosity, what makes you want to build Warp with a newer version of LLVM?

Internally we build everything against a version of LLVM pretty close to HEAD (usually only a couple of days old). I want to remove the old functions upstream and warp is one of the projects that we build against our internal LLVM.

> I had a look at it myself recently, in the hopes that it would address a compilation bug we've been running into. I found that there are more issues than just the Triple argument for createTargetMachine() though. I'll create an MR soon.

Yeah, we probably have patches for everything else needed to build against current tip of tree that are internal. Usually there's not a lot of interest in trying to upstream those patches though. I only was looking at this one since I was looking at doing the cleanup to remove it in upstream.

boomanaiden154:
This specific patch also isn't necessary to build against LLVM at HEAD, but it will be shortly in the future.

shi-eric:
Thanks for the heads up @boomanaiden154! This is now merged.

PR #1023, opened 2025-10-11
Add alpha and beta scalings to tile_matmul

RSchwan:
## Description
This PR adds alpha and beta scalings to `tile_matmul` as in the [cublasdx API](https://docs.nvidia.com/cuda/cublasdx/using_cublasdx.html). The main motivation is to generalize the `tile_matmul` API with the aim to reduce the number of operations and shared memory of otherwise necessary operations. For example, in the tile_block_cholesky example
```diff
@@ -78,8 +78,7 @@ def blocked_cholesky_kernel(
                 for j in range(0, k, block_size):
                     L_block = wp.tile_load(L, shape=(block_size, block_size), offset=(k, j))
                     L_block_T = wp.tile_transpose(L_block)
-                    L_L_T_block = wp.tile_matmul(L_block, L_block_T)
-                    A_kk_tile -= L_L_T_block
+                    wp.tile_matmul(L_block, L_block_T, A_kk_tile, alpha=-1.0)
```
this reduces shared memory usage (`L_L_T_block` doesn't exist) and fuses the subtraction.

## Summary by CodeRabbit

- **New Features**
  - tile_matmul adds alpha and beta scalars: out = alpha*(A @ B) + beta*out (defaults 1.0); supports in-place scaled matmul.

- **Documentation**
  - API docs and overloads updated to describe alpha/beta, defaults, semantics, and note adjoint support is not yet available.

- **Examples**
  - Cholesky example simplified to use tile_matmul(alpha=-1.0) for in-place subtraction.

- **Tests**
  - Tests updated for alpha/beta forward behavior and adjusted gradient expectations.

### DIFF
--- a/warp/tests/tile/test_tile_mathdx.py
+++ b/warp/tests/tile/test_tile_mathdx.py
@@ -40,8 +40,8 @@ def tile_math_matmul_kernel(
     i, j = wp.tid()
     a = wp.tile_load(ga, shape=(TILE_M, TILE_K), offset=(i * TILE_M, j * TILE_K))
     b = wp.tile_load(gb, shape=(TILE_K, TILE_N), offset=(i * TILE_K, j * TILE_N))
-    c = wp.tile_zeros(shape=(TILE_M, TILE_N), dtype=wp.float64)
-    wp.tile_matmul(a, b, c)
+    c = wp.tile_load(gc, shape=(TILE_M, TILE_N), offset=(i * TILE_M, j * TILE_N))
+    wp.tile_matmul(a, b, c, alpha=0.5, beta=-1.3)
     wp.tile_store(gc, c, offset=(i * TILE_M, j * TILE_N))
 
 
@@ -50,7 +50,7 @@ def test_tile_math_matmul(test, device):
 
     A = rng.random((TILE_M, TILE_K), dtype=np.float64).astype(np.float16)
     B = rng.random((TILE_K, TILE_N), dtype=np.float32)
-    C = np.zeros((TILE_M, TILE_N), dtype=np.float64)
+    C = rng.random((TILE_M, TILE_N), dtype=np.float64)
 
     A_wp = wp.array(A, requires_grad=True, device=device)
     B_wp = wp.array(B, requires_grad=True, device=device)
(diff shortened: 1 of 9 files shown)


warp/tests/tile/test_tile_mathdx.py
```diff
@@ -40,8 +40,8 @@ def tile_math_matmul_kernel(
     i, j = wp.tid()
     a = wp.tile_load(ga, shape=(TILE_M, TILE_K), offset=(i * TILE_M, j * TILE_K))
     b = wp.tile_load(gb, shape=(TILE_K, TILE_N), offset=(i * TILE_K, j * TILE_N))
-    c = wp.tile_zeros(shape=(TILE_M, TILE_N), dtype=wp.float64)
-    wp.tile_matmul(a, b, c)
+    c = wp.tile_load(gc, shape=(TILE_M, TILE_N), offset=(i * TILE_M, j * TILE_N))
+    wp.tile_matmul(a, b, c, alpha=0.5, beta=-1.3)
     wp.tile_store(gc, c, offset=(i * TILE_M, j * TILE_N))
```
daedalus5, line 45: There's a subtle complication here that was always in fact an issue. This change has brought it to the forefront however.

It's always safe in an autodiff system to have unique outputs and inputs. In this case however we are constrained. We cannot write:
`D = alpha * A * B + beta * C`

We use cublasdx under the hood, and this framework chooses to save memory by overwriting `C` instead:
`C = alpha * A * B + beta * C`

This forces some awkwardness when it comes to autodiff. First and foremost is violates the principle: "Don't write to an array/tile after reading from it". In a more complicated graph, eg:
```
D = f(C)
matmul(A, B, C)
Loss = g(D, C)
```
The the adjoint of C would be computed incorrectly if f() is nonlinear, since the wrong C would be used in the backward pass.

Fortunately matmul is linear in C, so that isn't an issue in that particular case, and I think you've done the right thing to compute
`adj_C *= beta * adj_C` for a replace assignment.

To do things properly from an autodiff perspective, I think we'd have to create an additional tile D, copy C's data to it, then use `fun_forward` with A, B, and D (leaving C untouched). So this function would return a new tile D. Unfortunately this comes with a performance cost.

This brings us to this test. Previously this issue was swept under the rug, because the C tile wasn't part of the adjoint chain. Here, it get's computed, and the test correctly expects:
`assert_np_equal(C_wp.grad.numpy(), adj_C - 1.3 * adj_C, tol=1e-2)`

Which is non-intuitive but technically correct. I think it'd be preferable however to allocate an additional array d and store the c tile to that, so instead we'd check:
`assert_np_equal(C_wp.grad.numpy(), -1.3 * adj_C, tol=1e-2)`

Less memory efficient but it is a more intuitive result.

Otherwise this PR looks good to me. We'll plan a follow-up analysis regarding the above. At the very least this warrants extra documentation.

RSchwan: Yes, this is a general issue with inplace operations and autodiff. In my use case, I'm not using the autodiff though and use warp to write high performing kernels. In my case, the memory reduction of this PR and https://github.com/NVIDIA/warp/pull/1025, results in quite significant speed-ups, since it massively improves the occupancy of my kernel.

Maybe one option could be to just disallow differentiability for the in-place operations (`is_differentiable=False`) or make sure it's clearly documented so that people understand the _technically correct_ outcome. If autodiff is required, and the user doesn't want to deal with this technicality, he can always use `c_out = wp.tile_matmul(a, b, alpha=0.5) - 1.3 * c`. I believe warp should at least offer the option to get the most performant kernel with the least memory used.

daedalus5: Agreed, this is a good change to have. This looks good as is -- we'll follow up with additional documentation. Thanks as always for your contribution. 

Would you mind rebasing and resolving any merge conflicts? We'll proceed with the merge then.

RSchwan:
Rebased and squashed the commits

PR #1075, opened 2025-11-09
Fix Explicit release of temporaries due to ref-cycle in fem.borrow_temporary 

Cucchi01:
**Context / Problem**

`fem.borrow_temporary()` gives back `warp.cache.Temporary` objects whose `array` attribute self-references the temporary. If we simply drop local references, the buffers keep a ref-cycle and only return to the pool when Python’s GC runs, contributing to situation like this issue in the related Newton repository newton-physics/newton#1042 

**Possible alternatives in my opinion**

1. **Wrap `temporary.array` in a `weakref.proxy`.**  
   - Pro: prevents the ref-cycle entirely; buffers go home as soon as user pointers die.  
   - Con: not backward compatible. Any code that stores `temp.array` or passes it to kernels now receives a proxy, so `wp.launch` can’t infer types (`weakref.ProxyType` shows up as the dtype).

2. **Patch `wp.array` itself (inside Warp) so `array` becomes a property and the temporary installs weakref-backed `release/detach`.**  
   - Pro: fixes the GC issue globally.  
   - Con: still changes the semantics of every `wp.array`: every array-like would be affected even if they never touch `fem.borrow_temporary`.

3. **Explicit release/detach in the whole codebase.**  
   - Keep the current `Temporary` API but audit the Warp library to release or detach buffers.  
   - Benefits: preserves user-facing semantics, keeps changes localized to Warp internals, and makes ownership obvious.  
   - Drawback: requires touching every call site that borrows or stores a temporary.

**Notes**

- Whenever we assign `self.attr = fem.borrow_temporary(...)`, in my changes the **old buffer is released first**.
Even in situation in which the temporary could be returned externally, e.g.:
```
    def fill_arg(self, args: Arg, device):
        args.cell_particle_offsets = self._cell_particle_offsets.to(device)
        args.cell_particle_indices = self._cell_particle_indices.to(device)
        args.particle_fraction = self._particle_fraction.to(device)
        args.particle_coords = self.particle_coords.to(device)
```

 Please correct me if this is not the expected behavior. 

**Still pending**

- Examples and any remaining `fem.integrate()` call sites outside the core solver still need to be changed. I'm going to do a follow-up commit to cover those.  
- Once that pass is done I’ll update this PR; for now the staged changes reflect the core library fixes that motivated the draft.

- [X] All commits are [signed-off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) to indicate that your contribution adheres to the [Developer Certificate of Origin](https://developercertificate.org/) requirements
- [ ] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `__init__.pyi`, `functions.rst`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management throughout the finite element module by releasing temporary resources more promptly after use, reducing memory retention and preventing potential leaks during object lifecycle transitions.

* **Performance**
  * Enhanced resource efficiency by implementing systematic cleanup of intermediate computational data structures during rebuilds and operations, resulting in more deterministic memory deallocation.

### DIFF
--- a/warp/_src/fem/adaptivity.py
+++ b/warp/_src/fem/adaptivity.py
@@ -65,6 +65,7 @@ def adaptive_nanogrid_from_hierarchy(
             device=device,
             inputs=[l, voxel_offsets[l], grid_voxels, merged_ijks],
         )
+        grid_voxels.release()
 
     # Allocate merged grid
     grid_info = grids[0].get_grid_info()
@@ -74,6 +75,7 @@ def adaptive_nanogrid_from_hierarchy(
         translation=grid_info.translation,
         device=device,
     )
+    merged_ijks.release()
 
     # Get unique voxel and corresponding level
     cell_count = cell_grid.get_voxel_count()
@@ -89,6 +91,7 @@ def adaptive_nanogrid_from_hierarchy(
         dim=cell_count,
         inputs=[level_count, cell_grid_ids, cell_ijk, cell_level],
     )
+    cell_ijk.release()
 
     cell_grid, cell_level = enforce_nanogrid_grading(
         cell_grid, cell_level, level_count=level_count, grading=grading, temporary_store=temporary_store
@@ -170,10 +173,17 @@ def adaptive_nanogrid_from_field(
                 fine_level,
             ],
         )
+        cell_refinement.release()
+
+        prev_cell_ijk = cell_ijk
+        prev_cell_level = cell_level
 
         # Fine is the new coarse
         cell_ijk = fine_ijk
         cell_level = fine_level
+        prev_cell_ijk.release()
+        prev_cell_level.release()
+    fine_count.release()
 
     wp.launch(_adjust_refined_ijk, dim=fine_shape, device=device, inputs=[cell_ijk, cell_level])
 
@@ -195,6 +205,8 @@ def adaptive_nanogrid_from_field(
         device=device,
         inputs=[fine_grid.id, cell_ijk, cell_level, fine_level],
     )
+    cell_ijk.release()
+    cell_level.release()
 
     fine_grid, fine_level = enforce_nanogrid_grading(
         fine_grid, fine_level, level_count=level_count, grading=grading, temporary_store=temporary_store
@@ -262,6 +274,8 @@ def enforce_nanogrid_grading(
         # Add new coordinates
         fine_shape = int(fine_count.numpy()[0])
         if fine_shape == cell_count:
+            cell_ijk.release()
+            refinement.release()
             break
 
         fine_ijk = cache.borrow_temporary(temporary_store, shape=fine_shape, dtype=wp.vec3i, device=device)
@@ -280,6 +294,8 @@ def enforce_nanogrid_grading(
                 fine_level,
             ],
         )
+        cell_ijk.release()
+        refinement.release()
 
         # Rebuild grid and levels
         cell_grid = wp.Volume.allocate_by_voxels(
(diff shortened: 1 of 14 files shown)


gdaviet:
Thanks @Cucchi01  for this thorough PR! I 'll have a closer look a the changes. 

In the meantime: here's a branch https://github.com/gdaviet/warp/tree/fix/temporary_self_reference  that marks the `.array` access as deprecated and uses weakref for release/detach  (I have a corresponding internal PR). This is similar to your 2) suggestion, since the attribute will be removed soon I think that it is the best workaround in the meantime. These changes reduce the memory increase by quite a bit in your mpm example, though it is still slightly increasing between GC runs.

Edit: I have switched to a less intrusive construct, using an array view. Downside is that we lose the deprecation warning, but at least it does not affect other array instances

gdaviet:
>  Whenever we assign self.attr = fem.borrow_temporary(...), in my changes the old buffer is released first.
>  Even in situation in which the temporary could be returned externally, e.g.

Yes, in this case releasing the old temporary first should be fine (there is not reason for the struct argument lifetime to exceed that of the class attribute), though  in some situations I would still prefer to have the garbage collector be in charge of the release (this was the main reason for the refactor of the Temporary object, as the old way of having separate temporary and arrays was creating too much logic related external accesses). 
Hopefully with the removal of the self-reference in the above branch we can get away with not doing explicit releases everywhere.

Cucchi01:
Perfect, then I am going to use your private branch for the moment @gdaviet. Thank you!

gdaviet:
When investigating this I actually found a couple more ref cycles. The fixes are now merged back to the main branch, and the memory usage is now much more stable!

```step    100 cpu=   606.39 MiB gpu=  1896.00 MiB
step    200 cpu=   606.53 MiB gpu=  1896.00 MiB
step    300 cpu=   606.67 MiB gpu=  1896.00 MiB
step    400 cpu=   606.81 MiB gpu=  1928.00 MiB
step    500 cpu=   606.81 MiB gpu=  1928.00 MiB
[...]
step   6200 cpu=   606.81 MiB gpu=  1896.00 MiB
step   6300 cpu=   606.81 MiB gpu=  1896.00 MiB
step   6400 cpu=   606.81 MiB gpu=  1896.00 MiB
step   6500 cpu=   606.81 MiB gpu=  1928.00 MiB
```

Cucchi01:
Great! Thank you!
Then let me know if I can simply close this PR and if I should keep open the one in Newton ([LINK](https://github.com/newton-physics/newton/pull/1057)) @gdaviet

gdaviet:
> Great! Thank you! Then let me know if I can simply close this PR and if I should keep open the one in Newton ([LINK](https://github.com/newton-physics/newton/pull/1057)) @gdaviet

Let's keep those open for now if that is ok with you. Explicit release is still a plus, I'll pull in your commits so you can get properly credited and adjust where relevant

gdaviet:
It's been long overdue but the early release part of the PR has finally been merged as a69d1023434411c89454e50ba2337b0cbeb508e6 (Unfortunately @Cucchi01 you handle has been lost somewhere in the commit authorship; likely a mistake on my side, attribution was correct in the internal MR though...)
The deprecated `.array` attribute has also been removed from the Temporary objects, ensuring there is no more self-ref issues (though the last implementation seemed to avoid the issue as well)

Cucchi01:
No problem at all @gdaviet 
 I hope it helped :)

PR #1155, opened 2026-01-07
Support @property getters for warp.struct

### LINKED ISSUE #1152: [REQ] Add support for python `@property` to `warp.struct`
### Description

Similar to standard Python classes, it would be useful to support the `@property` decorator for `warp.struct` definitions.

Currently, if a user wants to derive a value from struct members, they must define a separate global `warp.func` and pass the struct as an argument.

Supporting read-only getters via `@property` will make the transformation code closer to the struct definition. At the code generation step, these could be lowered to `warp.Functions` that accept the struct instance as `self`.

### Context

`@property` is an idiomatic Python feature that is useful for defining class member transformations close to the class definition.

Having this feature would be useful for:

- Keeping derived logic bundled with the data structure rather than scattering helper functions in the global namespace.
- Enabling dot-notation (inst.prop) inside kernels, which is often cleaner than function wrapping (get_prop(inst)).

--

I am happy to provide a PR in this direction to discuss implementation details.

Ghelfi:
## Description

Add support for `@property` in `warp.struct`.
Allows user to define property getters in the warp.struct. The properties are converted into warp functions at the codegen step.

Closes #1152 

- [x] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `__init__.pyi`, `docs/api_reference/`, `docs/language_reference/`)

## Summary by CodeRabbit

* **New Features**
  * Read-only properties on structs can be defined and accessed as first-class members; struct metadata now exposes property functions.

* **Bug Fixes / Behavior**
  * Struct attribute reads dispatch to property implementations instead of direct field access.
  * Setters/deleters for struct properties are disallowed and will raise an error.

* **Tests**
  * Added tests validating property reads on structs and rejecting property setters.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/warp/_src/codegen.py
+++ b/warp/_src/codegen.py
@@ -460,6 +460,7 @@ class Struct:
         self.cls = cls
         self.module = module
         self.vars: dict[str, Var] = {}
+        self.properties: dict[str, warp._src.context.Function] = {}
 
         if isinstance(self.cls, Sequence):
             raise RuntimeError("Warp structs must be defined as base classes")
@@ -482,6 +483,12 @@ class Struct:
                     warp.init()
                 fields.append((label, var.type._type_))
 
+        # Collect properties, but postpone Function creation until after native_name is set
+        property_members = []
+        for name, item in inspect.getmembers(self.cls):
+            if isinstance(item, property):
+                property_members.append((name, item))
+
         class StructType(ctypes.Structure):
             # if struct is empty, add a dummy field to avoid launch errors on CPU device ("ffi_prep_cif failed")
             _fields_ = fields or [("_dummy_", ctypes.c_byte)]
@@ -502,11 +509,50 @@ class Struct:
             if isinstance(type_hint, Struct):
                 ch.update(type_hint.hash)
 
-        self.hash = ch.digest()
+        # Hash property names (to ensure layout/identity stability if names change)
+        for name, _ in property_members:
+            ch.update(bytes(name, "utf-8"))
 
+        self.hash = ch.digest()
         # generate unique identifier for structs in native code
         hash_suffix = f"{self.hash.hex()[:8]}"
         self.native_name = f"{self.key}_{hash_suffix}"
+
+        # Extract properties and create Functions
+        # self.native_name is now defined, so Function() can resolve 'self' type code.
+        for name, item in property_members:
+            # We currently support only getters
+            if item.fset is not None:
+                raise TypeError("Struct properties with setters are not supported")
+            if item.fdel is not None:
+                raise TypeError("Struct properties with deleters are not supported")
+            getter = item.fget
+            # We need to add 'self' as the first argument, with the type of the struct itself.
+            # This allows overload resolution to match the struct instance to the 'self' argument.
+            if not hasattr(getter, "__annotations__"):
+                getter.__annotations__ = {}
+
+            # Find the name of the first argument (conventionally 'self')
+            argspec = get_full_arg_spec(getter)
+            if len(argspec.args) > 0:
+                self_arg = argspec.args[0]
+                getter.__annotations__[self_arg] = self
+
+            # Create the Warp Function.
+            # We pass 'func=getter' so that input_types and return_types are inferred.
+            # We set 'namespace=""' and a unique 'native_func' to generate a free function
+            # in C++ that takes the struct as the first argument (e.g., StructName_propName(struct_inst)).
+            p_func = warp._src.context.Function(
+                func=getter,
+                key=f"{self.key}.{name}",
+                namespace="",
+                module=module,
+            )
+
+            # Ensure the C++ function name is unique and predictable
+            p_func.native_func = f"{self.native_name}_{name}"
+
+            self.properties[name] = p_func
 
         # create default constructor (zero-initialize)
         self.default_constructor = warp._src.context.Function(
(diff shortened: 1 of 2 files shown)


warp/_src/codegen.py
```diff
@@ -501,13 +508,52 @@ class StructType(ctypes.Structure):
...
         # generate unique identifier for structs in native code
         hash_suffix = f"{self.hash.hex()[:8]}"
         self.native_name = f"{self.key}_{hash_suffix}"
 
+        # Extract properties and create Functions
+        # self.native_name is now defined, so Function() can resolve 'self' type code.
+        for name, item in property_members:
+            # We currently support only getters
+            if item.fset is not None:
+                raise TypeError("Struct properties with setters are not supported")
```
Ghelfi, line 528: For now, only getters are supported. Setters should be doable too.

warp/tests/test_struct.py
```diff
@@ -895,6 +895,33 @@ def test_nested_vec_assignment(self):
...
+    def neg_value(self) -> float:
+        return -self.value
+
+
+@wp.kernel
+def kernel_struct_property(s: StructWithProperty, out: wp.array(dtype=float)):
+    out[0] = s.neg_value
+
+
+def test_struct_property(test, device):
```
antoinebrl, line 913: Should there be another test to assert we raise when property setter is used?

antoinebrl (commented):
Would it make sense to augment the documentation to illustrate this new feature?

warp/tests/test_struct.py
Ghelfi: Added!

warp/_src/codegen.py
```diff
@@ -502,12 +509,51 @@ class StructType(ctypes.Structure):
...
+            # We need to add 'self' as the first argument, with the type of the struct itself.
+            # This allows overload resolution to match the struct instance to the 'self' argument.
+            if not hasattr(getter, "__annotations__"):
+                getter.__annotations__ = {}
+
+            # Find the name of the first argument (conventionally 'self')
+            argspec = get_full_arg_spec(getter)
+            if len(argspec.args) > 0:
+                self_arg = argspec.args[0]
+                getter.__annotations__[self_arg] = self
```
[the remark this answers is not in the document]
Ghelfi: Applied

Ghelfi:
@shi-eric @nvlukasz any chance I can have a review on this?

warp/_src/codegen.py
```diff
@@ -502,12 +511,53 @@ class StructType(ctypes.Structure):
             if isinstance(type_hint, Struct):
                 ch.update(type_hint.hash)
 
-        self.hash = ch.digest()
+        # Hash property names (to ensure layout/identity stability if names change)
+        for name, _ in property_members:
+            ch.update(bytes(name, "utf-8"))
```
nvlukasz, line 516: This only considers the name, but it should also include the function source. Otherwise modifying the getter won't trigger recompilation.

warp/_src/codegen.py
```diff
@@ -502,12 +511,53 @@ class StructType(ctypes.Structure):
...
+            self_arg = argspec.args[0]
+            getter.__annotations__[self_arg] = self
+
+            # Create the Warp Function.
+            # We pass 'func=getter' so that input_types and return_types are inferred.
+            # We set 'namespace=""' and a unique 'native_func' to generate a free function
+            # in C++ that takes the struct as the first argument (e.g., StructName_propName(struct_inst)).
+            p_func = warp._src.context.Function(
+                func=getter,
+                key=f"{self.key}.{name}",
```
nvlukasz, line 551: The key shouldn't include the `.` character (or other non-identifier characters). Can use underscore instead.

warp/_src/codegen.py
```diff
@@ -502,12 +511,53 @@ class StructType(ctypes.Structure):
...
+                namespace="",
+                module=module,
+            )
+
+            # Ensure the C++ function name is unique and predictable
+            p_func.native_func = f"{self.native_name}_{name}"
```
nvlukasz, line 557: This should be passed to the `Function` constructor as `native_func=...`.

warp/_src/codegen.py
nvlukasz: Without setters, this feature feels incomplete.

warp/_src/codegen.py
```diff
@@ -502,12 +511,53 @@ class StructType(ctypes.Structure):
...
+            getter = item.fget
+            # We need to add 'self' as the first argument, with the type of the struct itself.
+            # This allows overload resolution to match the struct instance to the 'self' argument.
+            if not hasattr(getter, "__annotations__"):
+                getter.__annotations__ = {}
+
+            # Find the name of the first argument (conventionally 'self')
+            argspec = get_full_arg_spec(getter)
+
+            if len(argspec.args) == 0:
```
nvlukasz, line 540: Does it make sense to allow more than one argument for the getter? How should that work? Would those need to be annotated by the user? I think it would currently fail during codegen without additional logic.

Might be better to reject any function that doesn't have exactly one argument here.

nvlukasz (commented):
Thanks @Ghelfi, it looks like a potentially useful feature, but there are some downsides.

I agree that `foo.bar` is simpler and easier to read/write than `get_bar(foo)`, but it's just syntactic sugar. Without setters, adding properties to structs feels incomplete. One concern is that we're adding complexity to the hashing and codegen. There are already some subtle problems with the implementation, like hashing not taking into account the function source.

Warp structs were originally intended to be simple POD (plain-old-data) types, without any higher-level OO semantics. I'm concerned that by adding properties we are venturing into OO territory that can end up being burdensome. If we add properties, why not methods? After all, properties are just methods in disguise. And then inheritance etc... Slippery slope :)

Of course, we can draw the line at just properties, perhaps even just getters. I'm not entirely against the idea, but it's something that needs a bit more thought and more extensive testing. For example, we should make sure that everything works fine with auto-differentiation. Should we allow custom gradient and replay function on properties? Things can get gnarly there.

Curious to hear @mmacklin's perspective.

PR #1169, opened 2026-01-12
Expose textures

nvtw:
## Description

Exposes textures to warp. This is a mirror MR, the main one is on the internal gitlab.

- [ ] All commits are [signed-off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) to indicate that your contribution adheres to the [Developer Certificate of Origin](https://developercertificate.org/) requirements
- [ ] Auto-generated files modified by compiling Warp and building the documentation have been updated (e.g. `__init__.pyi`, `docs/api_reference/`, `docs/language_reference/`)

## Summary by CodeRabbit

* **New Features**
  * Added Texture, Texture2D, Texture3D types with CPU/GPU support, and texture sampling (texture_sample) for 2D/3D (vector and scalar coords).
  * Added TextureFilterMode and TextureAddressMode enums and device/host texture creation + kernel interop.

* **Documentation**
  * Added API reference, language reference, and user guide sections for textures and sampling.

* **Tests**
  * Added extensive CPU/CUDA tests covering channels, dtypes, filtering, address modes, and edge cases.

* **Chores**
  * Enabled native texture subsystem in the build.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/warp/_src/codegen.py
+++ b/warp/_src/codegen.py
@@ -376,10 +376,22 @@ class StructInstance:
         return tuple(npvalue)
 
 
+def _is_texture_type(var_type: type) -> bool:
+    """Check if var_type is a Texture subclass (Texture2D, Texture3D, etc.)."""
+    from warp._src.texture import Texture  # noqa: PLC0415
+
+    try:
+        return issubclass(var_type, Texture)
+    except TypeError:
+        return False
+
+
 def _make_struct_field_constructor(field: str, var_type: type):
     if isinstance(var_type, Struct):
         return lambda ctype: var_type.instance_type(ctype=getattr(ctype, field))
     elif isinstance(var_type, warp._src.types.array):
+        return lambda ctype: None
+    elif _is_texture_type(var_type):
         return lambda ctype: None
     elif issubclass(var_type, ctypes.Array):
         # for vector/matrices, the Python attribute aliases the ctype one
@@ -442,10 +454,23 @@ def _make_struct_field_setter(cls, field: str, var_type: type):
 
         cls.__setattr__(inst, field, value)
 
+    def set_texture_value(inst, value):
+        # Texture2D, Texture3D, etc.
+        if value is None:
+            # create texture with null/default handle
+            setattr(inst._ctype, field, var_type._wp_ctype_())
+        else:
+            # texture instance
+            setattr(inst._ctype, field, value.__ctype__())
+
+        cls.__setattr__(inst, field, value)
+
     if isinstance(var_type, array):
         return set_array_value
     elif isinstance(var_type, Struct):
         return set_struct_value
+    elif _is_texture_type(var_type):
+        return set_texture_value
     elif issubclass(var_type, ctypes.Array):
         return set_vector_value
     else:
@@ -476,6 +501,8 @@ class Struct:
                 fields.append((label, var.type.ctype))
             elif issubclass(var.type, ctypes.Array):
                 fields.append((label, var.type))
+            elif _is_texture_type(var.type):
+                fields.append((label, var.type._wp_ctype_))
             else:
                 # HACK: fp16 requires conversion functions from warp.so
                 if var.type is warp.float16:
--- a/warp/tests/cuda/test_texture.py
+++ b/warp/tests/cuda/test_texture.py
@@ -1750,6 +1750,191 @@ def test_texture3d_backward_compat_address_mode(test, device):
 
 
 # ============================================================================
+# Texture as Struct Member Tests
+# ============================================================================
+
+
+@wp.struct
+class TextureStruct2D:
+    """Struct containing a 2D texture member."""
+
+    tex: wp.Texture2D
+    scale: float
+
+
+@wp.struct
+class TextureStruct3D:
+    """Struct containing a 3D texture member."""
+
+    tex: wp.Texture3D
+    offset: float
+
+
+@wp.struct
+class TextureStructBoth:
+    """Struct containing both 2D and 3D texture members."""
+
+    tex2d: wp.Texture2D
+    tex3d: wp.Texture3D
+    multiplier: float
+
+
+@wp.kernel
+def sample_texture2d_from_struct(
+    s: TextureStruct2D,
+    uv: wp.vec2f,
+    output: wp.array(dtype=float),
+):
+    """Sample a 2D texture from a struct member."""
+    tid = wp.tid()
+    value = wp.texture_sample(s.tex, uv, dtype=float)
+    output[tid] = value * s.scale
+
+
+@wp.kernel
+def sample_texture3d_from_struct(
+    s: TextureStruct3D,
+    uvw: wp.vec3f,
+    output: wp.array(dtype=float),
+):
+    """Sample a 3D texture from a struct member."""
+    tid = wp.tid()
+    value = wp.texture_sample(s.tex, uvw, dtype=float)
+    output[tid] = value + s.offset
+
+
+@wp.kernel
+def sample_both_textures_from_struct(
+    s: TextureStructBoth,
+    uv: wp.vec2f,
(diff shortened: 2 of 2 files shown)


StafaH:
Hi @nvtw, Sorry for the late update. Had a chance to try this out today, but was having difficulty integrating it into the existing renderer. The problem stems from how the textures are created, since we might have N number of textures. Usually we would want to create an array of handles which we store, and then pass them into the render kernel. The current implementation is not ergonomic for this, since it assumes you create a Texture2D (or 3D) object to store these individually, since creating a wp.array of the texture type is not allowed.

What's the best way to proceed for multiple textures?

nvtw:
> Hi @nvtw, Sorry for the late update. Had a chance to try this out today, but was having difficulty integrating it into the existing renderer. The problem stems from how the textures are created, since we might have N number of textures. Usually we would want to create an array of handles which we store, and then pass them into the render kernel. The current implementation is not ergonomic for this, since it assumes you create a Texture2D (or 3D) object to store these individually, since creating a wp.array of the texture type is not allowed.
> 
> What's the best way to proceed for multiple textures?

Thanks for having a look. Is having 2 arrays also an option, one for 2D textures and one for 3D textures? I can try to allow an array of a generic texture type but then we lose type safety on the sampling methods. You could call the 3d texture interpolation on a 2d texture and it would not be compile time checked. At least I have not figured out yet a way to make it compile type checked.

I can experiment a bit more, but please let me know if two separate warp arrays, one for 2d, one for 3d textures would be an option. I mean you need different arguments anyway when sampling them, once 2 floats, once 3.

nvtw:
There was indeed an issue with having Texture2D and Texture3D instances inside a warp array. It should be fixed now. But you still need separate warp arrays for 2d and 3d textures.

StafaH:
Still hitting some errors. I think I implemented it correct (from what I can tell from the docs/docstrings), but I'm hitting cuda errors with illegal memory access.

You can test this downstream locally using my fork of the renderer: https://github.com/StafaH/mujoco_warp/tree/render_cleanup_textures

Just clone the repo and then run a rendering test with:
`python contrib/render.py mujoco_warp/test_data/primitives.xml`

You can see from the most recent commit in that branch how I implement cuda textures (for me only 2D is needed so I don't need to worry about mixing 2d/3d). I also tried using the exact method in the docstring (randn of shape (x, y, 4)), and it still hits an illegal memory access error. Not sure the root cause but either I implemented it incorrectly or the texture object got garbage collected even though we stuck it into an array and the sampling becomes invalid.

nvtw:
> Still hitting some errors. I think I implemented it correct (from what I can tell from the docs/docstrings), but I'm hitting cuda errors with illegal memory access.
> 
> You can test this downstream locally using my fork of the renderer: https://github.com/StafaH/mujoco_warp/tree/render_cleanup_textures
> 
> Just clone the repo and then run a rendering test with: `python contrib/render.py mujoco_warp/test_data/primitives.xml`
> 
> You can see from the most recent commit in that branch how I implement cuda textures (for me only 2D is needed so I don't need to worry about mixing 2d/3d). I also tried using the exact method in the docstring (randn of shape (x, y, 4)), and it still hits an illegal memory access error. Not sure the root cause but either I implemented it incorrectly or the texture object got garbage collected even though we stuck it into an array and the sampling becomes invalid.

Thanks for testing it @StafaH , this is very valuable. Can you try again after you applied the patch below? The main issue seems to be that you don't keep the texture object references alive on the host. That will trigger the python garbage collector to clean them up and therefore you get an invalid memory access.

```
diff --git a/mujoco_warp/_src/render_context.py b/mujoco_warp/_src/render_context.py
index a577227..a063fff 100644
--- a/mujoco_warp/_src/render_context.py
+++ b/mujoco_warp/_src/render_context.py
@@ -200,10 +200,12 @@ class RenderContext:
       self.flex_radius = mjm.flex_radius
       self.flex_render_smooth = flex_render_smooth
     
-    textures = []
+    # IMPORTANT: Keep references to Texture2D objects alive, otherwise they get
+    # garbage collected and the CUDA textures are destroyed (use-after-free bug).
+    self._texture_objects = []
     for i in range(mjm.ntex):
-      textures.append(_create_warp_texture(mjm, i))
-    self.textures = wp.array(textures, dtype=wp.Texture2D)
+      self._texture_objects.append(_create_warp_texture(mjm, i))
+    self.textures = wp.array(self._texture_objects, dtype=wp.Texture2D)
 
     tex_data_packed, tex_adr_packed = _create_packed_texture_data(mjm)
 
@@ -371,7 +373,7 @@ def build_primary_rays(
   y = top + (bottom - top) * v
   ray_out[offset + tid] = wp.normalize(wp.vec3(x, y, -znear))
 
-def _create_warp_texture(mjm: mujoco.MjModel, tex_id: int) -> wp.array:
+def _create_warp_texture(mjm: mujoco.MjModel, tex_id: int) -> wp.Texture2D:
   """Create a Warp texture from a MuJoCo model texture data."""
   print(f"Creating texture {tex_id}")
   print("Texture width: ", mjm.tex_width[tex_id])
@@ -382,7 +384,7 @@ def _create_warp_texture(mjm: mujoco.MjModel, tex_id: int) -> wp.array:
   tex_height = mjm.tex_height[tex_id]
   nchannel = mjm.tex_nchannel[tex_id]
 
-  tex_data = wp.zeros((tex_width, tex_height, 4), dtype=wp.float32)
+  tex_data = wp.zeros((tex_height, tex_width, 4), dtype=wp.float32)
   
   @wp.kernel
   def convert_texture_to_vec4(
@@ -404,10 +406,10 @@ def _create_warp_texture(mjm: mujoco.MjModel, tex_id: int) -> wp.array:
     g = tex_data_in[src_idx + 1] if nc > 1 else wp.uint8(0)
     b = tex_data_in[src_idx + 2] if nc > 2 else wp.uint8(0)
     a = wp.uint8(255)
-    tex_data_out[ix, iy, 0] = float(r) / 255.0
-    tex_data_out[ix, iy, 1] = float(g) / 255.0
-    tex_data_out[ix, iy, 2] = float(b) / 255.0
-    tex_data_out[ix, iy, 3] = float(a) / 255.0
+    tex_data_out[iy, ix, 0] = float(r) / 255.0
+    tex_data_out[iy, ix, 1] = float(g) / 255.0
+    tex_data_out[iy, ix, 2] = float(b) / 255.0
+    tex_data_out[iy, ix, 3] = float(a) / 255.0
   
   wp.launch(
     convert_texture_to_vec4,
@@ -761,3 +763,4 @@ def _make_flex_mesh(
     flex_faceadr,
     nface,
   )
+

```

nvlukasz:
> The main issue seems to be that you don't keep the texture object references alive on the host. That will trigger the python garbage collector to clean them up and therefore you get an invalid memory access.

Right, this is similar to how the `wp.array` class works - once all Python references go away, it triggers deallocation.

Keeping the Python `Texture` objects alive somehow is the way to go. I wonder if we can do this automatically with arrays of textures, e.g.:
```python
texture_list = [wp.Texture(...), ...]
texture_arr = wp.array(data=texture_list, dtype=wp.Texture)
```

Could we store a reference to `texture_list` in the `texture_arr`? If `dtype` is Texture, retain the input list in the constructor (e.g., `self._dataref = data`).

StafaH:
Thanks @nvtw, yes if I keep a second list with the original python references then it does work. It's not ideal but we already do the same for all the BVH objects that we make.

Can confirm there are no performance regressions in downstream rendering (no speedups either but that's to be expected). I've finalized the changes in the renderer, same branch as above. Once this feature is in nightly I'll update the mjwarp renderer.

Here are some nice rendering (first is plane texture, second is a mesh texture):

image
image

shi-eric:
Changes were merged in 9574be87091d65fd7f33aba394370c3331090f4b

Issue #1235, opened 2026-02-17
[BUG] vswhere.exe discovery fails for non-default Visual Studio installations


shi-eric:
`find_host_compiler()` in [`warp/_src/build_dll.py` (lines 127–130)](https://github.com/NVIDIA/warp/blob/main/warp/_src/build_dll.py#L127-L130) locates `vswhere.exe` at a single hardcoded path:

```python
vswhere_path = r"%ProgramFiles(x86)%/Microsoft Visual Studio/Installer/vswhere.exe"
vswhere_path = os.path.expandvars(vswhere_path)
if not os.path.exists(vswhere_path):
    return ""  # Signal to caller that VS not found
```

This works for the vast majority of users since `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\` is the default install location for `vswhere.exe`. However, in rare cases where Visual Studio is installed in a non-default directory — or where `vswhere.exe` was installed standalone via a package manager (Chocolatey, Scoop, winget) — the lookup silently fails and produces:

```
Warp build error: Could not find MSVC compiler
```

even though MSVC is installed and functional.

**Workaround**: Use `--msvc-path` and `--sdk-path` flags to explicitly specify toolchain locations.

## Proposed Fix

**1. Add a `_find_vswhere()` helper** to `warp/_src/build_dll.py` with multi-strategy discovery:

```python
def _find_vswhere() -> str:
    """Locate vswhere.exe using multiple discovery strategies.

    Search order:
    1. ``PATH`` (covers Chocolatey, Scoop, winget, or manual installs)
    2. Default VS Installer location under ``%ProgramFiles(x86)%``
    3. Fallback under ``%ProgramFiles%`` (32-bit Windows or older setups)

    Returns:
        Absolute path to vswhere.exe, or empty string if not found.
    """
    path_result = shutil.which("vswhere")
    if path_result:
        if verbose_cmd:
            print(f"Found vswhere.exe in PATH: {path_result}")
        return path_result

    candidates = [
        os.path.expandvars(r"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe"),
        os.path.expandvars(r"%ProgramFiles%\Microsoft Visual Studio\Installer\vswhere.exe"),
    ]
    for candidate in candidates:
        if os.path.isfile(candidate):
            if verbose_cmd:
                print(f"Found vswhere.exe at: {candidate}")
            return candidate

    if verbose_cmd:
[...]

PR #1248, opened 2026-02-22
 Fix vswhere.exe discovery for non-default Visual Studio installations

Adityakk9031:
# Description

This PR fixes an issue where `vswhere.exe` discovery would silently fail for non-default Visual Studio installations or when `vswhere` was installed via a standalone package manager (e.g., Chocolatey, Scoop, winget).

**Changes:**
- Added a `_find_vswhere()` helper in `warp/_src/build_dll.py` with multi-strategy path discovery. It now searches `PATH` before falling back to the standard VS Installer paths under `%ProgramFiles(x86)%` and `%ProgramFiles%`.
- Replaced the hardcoded path lookup in `find_host_compiler()` with a call to the new helper.
- Improved the error message in `build_lib.py` when MSVC cannot be found. The message now mentions actionable workarounds, such as using the `--msvc-path` and `--sdk-path` flags or ensuring `vswhere.exe` is in the `PATH`.
- Added an entry to `CHANGELOG.md` under the `Fixed` section.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable Visual Studio/MSVC discovery for non-default and package-manager installations.

* **Improvements**
  * Clearer, multi-line diagnostics when MSVC is not found, with actionable troubleshooting and custom toolchain options.
  * More robust handling and decoding of discovery outputs with additional verbose diagnostics for failure scenarios.

* **Documentation**
  * Added changelog entry noting the discovery and messaging updates.

### DIFF
--- a/warp/_src/build_dll.py
+++ b/warp/_src/build_dll.py
@@ -56,7 +56,7 @@ def run_cmd(cmd):
         output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
         # Print output even on success to show warnings
         if output:
-            decoded_output = output.decode()
+            decoded_output = output.decode("utf-8", errors="replace")
             if decoded_output.strip():  # Only print if not just whitespace
                 # In parallel builds, associate output with its command for clarity
                 # Use single print to avoid interleaving with other processes
@@ -64,7 +64,9 @@ def run_cmd(cmd):
         return output
     except subprocess.CalledProcessError as e:
         # Single print to avoid interleaving in parallel builds
-        print(f"Command failed with exit code {e.returncode}: {cmd}\nCommand output was:\n{e.output.decode()}")
+        print(
+            f"Command failed with exit code {e.returncode}: {cmd}\nCommand output was:\n{e.output.decode('utf-8', errors='replace')}"
+        )
         raise e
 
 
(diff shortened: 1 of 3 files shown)


Adityakk9031:
@shi-eric have a look

Adityakk9031:
@christophercrouzet please review it

c0d1f1ed:
This looks good to me. I'll see if we can get it into the 12.1 release.

c0d1f1ed:
@Adityakk9031 Thanks for the MR. I don't have any strictly technical concerns with it, but it adds 67 lines of code so I just wanted to verify that it's worth the complexity. Could you detail the exact issue you encountered and why all of the changes in this MR are necessary to address it?

Adityakk9031:
@c0d1f1ed The exact issue details are in #1235 . I just implemented the proposed fix described there as-is. extra lines are the error handling

Adityakk9031:
@c0d1f1ed sir it need any modification. if any please let me know

warp/_src/build_dll.py
```diff
@@ -96,6 +98,38 @@ def set_msvc_env(msvc_path, sdk_path):
...
+    Returns:
+        Absolute path to vswhere.exe, or empty string if not found.
+    """
+    path_result = shutil.which("vswhere.exe")
+    if path_result:
+        if verbose_cmd:
+            print(f"Found vswhere.exe in PATH: {path_result}")
+        return path_result
+
+    candidates = [
```
c0d1f1ed, line 118: Under what conditions do we actually end up here having to try other candidate paths? As noted on line 105, accessing vswhere from PATH should already cover any known installation method.

warp/_src/build_dll.py
```diff
@@ -124,26 +158,52 @@ def find_host_compiler() -> str:
...
+            if verbose_cmd:
+                print("Warning: vswhere.exe found no Visual Studio installation")
+            return ""
         vsvars_path = os.path.join(vs_path, "VC\\Auxiliary\\Build\\vcvars64.bat")
 
         if not os.path.exists(vsvars_path):
             return ""  # Signal to caller that VS environment script not found
 
-        output = run_cmd(f'"{vsvars_path}" && set').decode()
+        try:
```
c0d1f1ed, line 182: I don't think this is necessary. Trying to call the batch script will already produce an error if it's not found or it failed.

warp/_src/build_dll.py
```diff
@@ -124,26 +158,52 @@ def find_host_compiler() -> str:
...
+            return ""  # Signal to caller that VS environment script failed
 
         for line in output.splitlines():
             pair = line.split("=", 1)
             if len(pair) >= 2:
                 os.environ[pair[0]] = pair[1]
 
         cl_path = shutil.which("cl.exe")
-        cl_version = os.environ["VCToolsVersion"].split(".")
+        if not cl_path:
```
c0d1f1ed, line 193: This also seems unnecessary. If it's not found in the environment something went seriously wrong and printing nicely formatted error messages isn't any more useful than just the Python exception that gets raised.

PR #1280, opened 2026-03-10
Fix AttributeError in set_module_options when called via runpy

### LINKED ISSUE #1274: `set_module_options` crashes with `AttributeError` when called from `runpy.run_module()`
## Description

`wp.set_module_options()`, `wp.get_module_options()`, and `wp.load_module()` crash with an unhelpful `AttributeError` when called from code executed via `runpy.run_module()` (e.g. `python -m package.module`).

The root cause is that all three functions use `inspect.getmodule(inspect.stack()[1][0])` to auto-detect the calling module, but `inspect.getmodule()` returns `None` when the module is not yet registered in `sys.modules` — which is the case during `runpy.run_module()` execution.

## Steps to Reproduce

```python
# my_pkg/my_module.py
import warp as wp
wp.set_module_options({"enable_backward": False})
```

```bash
python -c "import runpy; runpy.run_module('my_pkg.my_module', run_name='__main__')"
```

Or equivalently, from a framework that uses `runpy.run_module()` internally (e.g. Newton's example browser):

```bash
python -m newton.examples robot_anymal_c_walk
```

## Error

```
File ".../warp/_src/context.py", line 8148, in set_module_options
    get_module(m.__name__).options.update(options)
               ^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute '__name__'. Did you mean: '__ne__'?
```

## Expected Behavior

`set_module_options()` should correctly identify the calling module even when executed via `runpy.run_module()`, by falling back to `__spec__` (which `runpy` sets on the namespace) or other frame metadata.

## Environment

- Warp 1.12.0
- Python 3.10+
- Any OS

cluster2600:
## Summary

Fixes #1274 — `set_module_options()`, `get_module_options()`, and `load_module()` crash with
`AttributeError: 'NoneType' object has no attribute '__name__'` when the caller is executed via
`runpy.run_module()` (i.e. `python -m package.module`).

The root cause is that `inspect.getmodule()` returns `None` when the calling module has not yet
been registered in `sys.modules`, which is teh case during `runpy.run_module()` execution.

## Changes

- Introduced `_resolve_module_from_frame(frame)` which tries `inspect.getmodule()` first, then
  falls back to `__spec__.name` and `__name__` lookups against `sys.modules`.
- Introduced `_caller_module_name(frame)` which returns a module name string even when no module
  object is found — the `__spec__` or `__name__` values are recognised as sufficient for the Warp
  module registry, which only needs a name to create or retrieve a `Module` entry.
- Updated `set_module_options()`, `get_module_options()`, and `load_module()` to use these helpers
  instead of calling `inspect.getmodule()` directly.
- Added unit tests covering the fallback behaviour and a full `runpy.run_module()` integration test.

## Test plan

- [x] Existing `test_options.py` tests continue to pass (normal import path unchanged)
- [x] New `test_resolve_module_from_frame_returns_module` — verifies helper works for standard frames
- [x] New `test_caller_module_name_returns_name` — verifies name resolution for standard frames
- [x] New `test_caller_module_name_fallback_to_spec` — verifies `__spec__` fallback when `getmodule` returns `None`
- [x] New `test_caller_module_name_fallback_to_dunder_name` — verifies `__name__` fallback
- [x] New `test_set_module_options_via_runpy` — end-to-end test using `runpy.run_module()` with a temporary package

## Summary by CodeRabbit

* **Refactor**
  * Improved module-frame resolution and error messaging with additional fallback logic to better handle non-standard Python execution contexts.

* **Tests**
  * Added tests covering module resolution and name-derivation behavior, including fallback scenarios and invocation via runtime module execution, to ensure reliability across environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
   ([GH-1254](https://github.com/NVIDIA/warp/issues/1254)).
 - Preserve precision of compile-time constants in 64-bit scalar type constructors (`wp.float64()`, `wp.int64()`,
   `wp.uint64()`) and perform constant folding of negation ([GH-485](https://github.com/NVIDIA/warp/issues/485)).
+- Fix `set_module_options()`, `get_module_options()`, and `load_module()` crashing with `AttributeError` when
+  called from code executed via `runpy.run_module()` (e.g. `python -m package.module`)
+  ([GH-1274](https://github.com/NVIDIA/warp/issues/1274)).
 
 ### Documentation
 - Fix internal module path `warp._src.lang` leaking into published documentation page titles, URLs, and search engine
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -7799,6 +7799,74 @@ def force_load(
         runtime.core.wp_cuda_context_set_current(saved_context)
 
 
+def _get_caller_module_name(stack_level: int = 1) -> str:
+    """Return the fully qualified module name of the caller.
+
+    Uses a multi-step fallback chain so that callers running under
+    ``runpy.run_module()`` (e.g. ``python -m pkg.example``) are handled
+    correctly even when the module is not yet registered in
+    ``sys.modules``.
+
+    Args:
+        stack_level: How many frames to walk up from **this** function.
+            Callers that are themselves one level above the user code
+            should pass ``2`` (one for this helper, one for the wrapper).
+
+    Returns:
+        The fully qualified name of the caller's module.
+
+    Raises:
+        RuntimeError: If the calling module cannot be determined.
+    """
+    frame = sys._getframe(stack_level)
+    try:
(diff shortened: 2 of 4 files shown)


cluster2600:
Closing this in favour of the upstream fix merged in ershi/robust-caller-module-detection (e6391e96).

Eric's implementation is a superset of what this PR introduced — it uses `sys._getframe()` instead of `inspect.stack()` for better performance, prioritises `frame.f_globals["__name__"]` for consistency with `@wp.kernel`'s use of `f.__module__`, and includes proper frame cleanup via `try/finally`. The fallback chain is also more thorogh, covering symlink resolution with `os.path.realpath()`.

The original issue (#1274) is referenced in the merge commit and should be resolved by the upstream changes. Thanks for the quick turnaround on the fix, @ershi — it's a cleaner approach than what I had here.

PR #1283, opened 2026-03-11
Fix UB in float-to-unsigned scalar type conversions

shivansh023023:
## Description
This PR fixes a bug causing undefined behavior when casting negative float values to unsigned integer scalar types (like `wp.uint8`) inside Warp kernels.
When `wp.uint8(float_val)` is used, the codegen emitted a direct C-style functional cast `wp::uint8(float_val)`. Since [uint8](cci:2://file:///home/shivanshsingh/Desktop/warp/warp/warp/_src/types.py:311:0-315:27) maps directly to `uint8_t` (unsigned char), casting a float with a negative truncated value to this unsigned type triggers **undefined behavior** per the C++ standard (§7.3.11).
* **Fix:** For `float* -> uint*` pairs during codegen, the emitted C++ code now routes via custom safe-cast functional overloads in [builtin.h](cci:7://file:///home/shivanshsingh/Desktop/warp/warp/warp/native/builtin.h:0:0-0:0) (e.g., [float32_to_uint8](cci:1://file:///home/shivanshsingh/Desktop/warp/warp/warp/native/builtin.h:109:0-109:111)). These explicitly perform a two-step cast: `float -> int64 -> unsigned`, leveraging C++'s well-defined modular arithmetic rule for signed-to-unsigned conversion.
* **Test Enabled:** Re-enabled the [test_type_closure_scalar](cci:1://file:///home/shivanshsingh/Desktop/warp/warp/warp/tests/test_codegen_instancing.py:1093:0-1098:73) case for [uint8](cci:2://file:///home/shivanshsingh/Desktop/warp/warp/warp/_src/types.py:311:0-315:27) in `test_codegen_instancing.py:1099`.
Closes FIXME in `warp/tests/test_codegen_instancing.py:1099`.
## Checklist
## Test plan
- Ensured that `python3 -m pytest warp/tests/test_codegen_instancing.py::TestCodeGenInstancing::test_type_closure_scalar_cpu -v` passes.
- Confirmed that explicit wrap-around bounds are handled correctly with arbitrary bounds:
  - `wp.uint8(-0.1)` yields `0`
  - `wp.uint8(255.1)` yields `255`
  - `wp.uint8(128.0)` yields `128`
  - `wp.uint8(-100.0)` yields `156` (expected modular wrap-around)
## Bug fix
```python
import warp as wp
wp.init()
@wp.kernel
def cast_negative_float():
    x = wp.float32(-1.5)
    
    # Without this PR, mapping a negative float directly to 
    # unsigned integer triggers undefined behavior in C++
    y = wp.uint8(x)
    wp.printf("y = %d\\n", int(y))
wp.launch(cast_negative_float, dim=1)

## Summary by CodeRabbit

* **New Features**
  * Added safe, well-defined float→unsigned integer conversion routines (including float16 paths) with corresponding no-op adjoint support.

* **Tests**
  * Expanded scalar conversion tests: re-enabled and broadened uint8 cases and added uint64/float64 boundary coverage.

* **Notes**
  * Public APIs and documented behavior remain unchanged.

### DIFF
--- a/repro_uint8_bug.py
+++ b/repro_uint8_bug.py
@@ -1,0 +1,40 @@
+"""Minimal reproduction of the uint8 type closure conversion bug."""
+
+import warp as wp
+
+wp.init()
+
+
+def create_type_closure_scalar(scalar_type):
+    @wp.kernel
+    def k(input: float, expected: float):
+        x = scalar_type(input)
+        wp.expect_eq(float(x), expected)
+
+    return k
+
+
+# These work fine (int, float closures)
+type_closure_kernel_int = create_type_closure_scalar(int)
+type_closure_kernel_float = create_type_closure_scalar(float)
+
+# This is the broken one
+type_closure_kernel_uint8 = create_type_closure_scalar(wp.uint8)
+
+print("Testing int closure...")
+wp.launch(type_closure_kernel_int, dim=1, inputs=[-1.5, -1.0], device="cpu")
+wp.synchronize()
+print("  PASSED")
+
+print("Testing float closure...")
+wp.launch(type_closure_kernel_float, dim=1, inputs=[-1.5, -1.5], device="cpu")
+wp.synchronize()
+print("  PASSED")
+
+print("Testing uint8 closure...")
+try:
+    wp.launch(type_closure_kernel_uint8, dim=1, inputs=[-1.5, 255.0], device="cpu")
+    wp.synchronize()
+    print("  PASSED")
+except Exception as e:
+    print(f"  FAILED with exception: {type(e).__name__}: {e}")
--- a/warp/_src/builtins.py
+++ b/warp/_src/builtins.py
@@ -1007,8 +1007,17 @@ add_builtin(
 
 # scalar type constructors between all storage / compute types
 scalar_types_all = [*scalar_types, bool, int, float]
+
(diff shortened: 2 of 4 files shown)


shi-eric:
Thanks for looking into this. I agree there is a real issue here: float-to-unsigned casts currently appear to behave differently between CPU and CUDA, so this is worth addressing.

That said, this PR changes the semantics/codegen path for float-to-unsigned casts, and I think that needs discussion before we pick a solution. Please open an issue first describing the CPU/CUDA discrepancy, the affected unsigned types, the current behavior, and the proposed semantics. In particular, we need to decide whether Warp should preserve wraparound behavior after truncation, saturate, reject out-of-range values, or follow some other rule.

There are also some contribution hygiene issues here. `repro_uint8_bug.py` should not be added at the repository root; a repro should either live in the issue description or be converted into an appropriate unit test once the desired behavior is agreed on.

@c0d1f1ed could you weigh in on what semantics we want for `float -> uint` casts in Warp? The PR focuses on `uint8`, apparently because of the disabled `type_closure_kernel_uint8` test, but the proposed change affects all unsigned integer types.

I’m going to close this PR for now so we can discuss the desired behavior in an issue first. If we agree on the semantics, we can take a focused follow-up PR with the repo-root repro removed and tests covering the selected behavior across CPU and CUDA.

PR #1299, opened 2026-03-17
Add support for assigning values to boolean vectors in Python

thomasbbrunner:
## Description

Currently, attempting to write to a field of a boolean vector within a kernel will lead to the error:

```
warp._src.codegen.WarpCodegenError: Error while parsing function "assign_bool_vector"
```

This PR adds support for assigning values to boolean vectors.

## Checklist

## Test plan

I've added a test to the repo. The test is passing with the changes.

## Bug fix

To reproduce this in the current main:

```python
import warp as wp

@wp.kernel
def assign_int_vector():
    v = wp.vector(length=2, dtype=wp.int32)
    v[0] = 1

@wp.kernel
def assign_bool_vector():
    v = wp.vector(length=2, dtype=wp.bool)
    v[0] = True

wp.launch(assign_int_vector, dim=1)
wp.synchronize()

# Raises:
# warp._src.codegen.WarpCodegenError: Error while parsing function "assign_bool_vector"
wp.launch(assign_bool_vector, dim=1)
wp.synchronize()
```

The compilation of the kernel with the int vector (`assign_int_vector`) works, but the one with the boolean vector (`assign_bool_vector`) fails.

## Summary by CodeRabbit

* **New Features**
  * Support for boolean vector element assignment and mutation, including negative indexing.

* **Tests**
  * New test verifying boolean vector assignment and negative-index behavior across device launches.

* **Chores**
  * Changelog updated to note the fix for boolean vector element assignment.

### DIFF
--- a/warp/tests/test_bool.py
+++ b/warp/tests/test_bool.py
@@ -188,6 +188,36 @@ def test_bool_mat_typing(test, device):
     wp.launch(test_bool_mat_anonymous_typing, (1,), inputs=[], device=device)
 
 
+@wp.func
+def bool_vec_assign():
+    v = vec3bool_type(True, False, True)
+
+    v[0] = False
+    v[2] = False
+
+    wp.expect_eq(v[0], False)
+    wp.expect_eq(v[1], False)
+    wp.expect_eq(v[2], False)
+
+    v[-1] = True
+    v[-3] = True
+
+    wp.expect_eq(v[-1], True)
+    wp.expect_eq(v[-2], False)
+    wp.expect_eq(v[-3], True)
+
+
+@wp.kernel
+def run_bool_vec_assign():
+    bool_vec_assign()
+
+
+def test_bool_vec_assign(test, device):
+    wp.launch(run_bool_vec_assign, 1, device=device)
+    wp.synchronize_device(device)
+    bool_vec_assign()
+
+
 devices = get_test_devices()
 
 
@@ -201,6 +231,7 @@ add_function_test(TestBool, "test_bool_constant_vec", test_bool_constant_vec, de
 add_function_test(TestBool, "test_bool_constant_mat", test_bool_constant_mat, devices=devices)
 add_function_test(TestBool, "test_bool_vec_typing", test_bool_vec_typing, devices=devices)
 add_function_test(TestBool, "test_bool_mat_typing", test_bool_mat_typing, devices=devices)
+add_function_test(TestBool, "test_bool_vec_assign", test_bool_vec_assign, devices=devices)
 
 
 if __name__ == "__main__":
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@
 - Fix struct field assignment unwrapping Warp scalar types (e.g., ``wp.float32``, ``wp.int32``) to their
   underlying Python types, causing subsequent reads to return plain ``float`` or ``int`` instead of the
   original Warp type ([GH-1288](https://github.com/NVIDIA/warp/issues/1288)).
+- Fix element assignment for boolean vectors failing with a missing function overload
+  ([GH-1302](https://github.com/NVIDIA/warp/issues/1302)).
 
 ### Documentation
 
--- a/warp/_src/builtins.py
+++ b/warp/_src/builtins.py
@@ -9828,6 +9828,17 @@ add_builtin(
     group="Utility",
 )
 
+# Bool vector assign_inplace (bool is not part of Scalar)
+add_builtin(
+    "assign_inplace",
+    input_types={"a": vector(length=Any, dtype=bool), "i": Any, "value": Any},
+    value_type=None,
+    dispatch_func=vector_assign_dispatch_func,
+    hidden=True,
+    export=False,
+    group="Utility",
+)
+
 # implements quaternion[index] = value
 add_builtin(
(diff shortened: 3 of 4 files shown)


warp/tests/test_bool.py
```diff
@@ -200,6 +200,34 @@ def test_bool_mat_typing(test, device):
     wp.launch(test_bool_mat_anonymous_typing, (1,), inputs=[], device=device)
 
 
+def test_bool_vec_assign(test, device):
```
thomasbbrunner: Unsure if this is the best place for this test. An alternative would be the `warp/tests/test_vec.py` file.

thomasbbrunner:
Also, should I add a changelog entry regarding this?

thomasbbrunner:
@shi-eric it would be great if we could add this to the next patch release.

This is my first contribution, so please let me know if there's anything I'm missing here.

warp/tests/test_bool.py
```diff
@@ -200,6 +200,34 @@ def test_bool_mat_typing(test, device):
...
+        v[-1] = True
+        v[-3] = True
+
+        wp.expect_eq(v[-1], True)
+        wp.expect_eq(v[-2], False)
+        wp.expect_eq(v[-3], True)
+
+    @wp.kernel(module="unique")
+    def kernel():
+        fn()
```
shi-eric: Can you move these definitions to the module level and pick descriptive names?

This is a micro-optimization as defining kernels inside functions can result in some extra compilation overhead due to runtime hash changes.

warp/tests/test_bool.py
shi-eric: Current location is fine, there's no clear winner to me.

shi-eric:
Thanks for the contribution! You've done a pretty good job of following project conventions and I appreciated the pull request description. It should be no problem to have this included in v1.12.1.

One thing missing: the `assign_copy` builtin isn't exercised by any test. It's only hit when `wp.config.enable_vector_component_overwrites = True`, so without a test for that path, a regression there would go unnoticed.

Here's an example of what that could look like, added to `test_bool.py` using the save/try/finally pattern:

```python
def test_bool_vec_assign_copy(test, device):
    old_setting = wp.config.enable_vector_component_overwrites

    try:
        wp.config.enable_vector_component_overwrites = True

        @wp.kernel(module="unique")
        def bool_vec_assign_copy_kernel():
            a = vec3bool_type(True, True, True)
            b = a
            b[1] = False

            # a should be unchanged (copy semantics)
            wp.expect_eq(a[0], True)
            wp.expect_eq(a[1], True)
            wp.expect_eq(a[2], True)

            # b should reflect the mutation
            wp.expect_eq(b[0], True)
            wp.expect_eq(b[1], False)
            wp.expect_eq(b[2], True)

        wp.launch(bool_vec_assign_copy_kernel, 1, device=device)
        wp.synchronize_device(device)
    finally:
        wp.config.enable_vector_component_overwrites = old_setting
```

I verified that this test fails with a clear codegen error (`Couldn't find function overload for 'assign_copy'`) when the `assign_copy` builtin registration is removed, and passes with it present.

Could you also file a GitHub issue for the original bug (assigning to boolean vector components not working) and associate this PR to it? Then add a changelog entry under the `Unreleased` section of `CHANGELOG.md` referencing that issue.

shi-eric (changes requested):
Just some minor things. As you revise your pull request it would help if you could rebase with main and keep changes limited to one commit for this fix, thanks!

warp/tests/test_bool.py
```diff
@@ -200,6 +200,34 @@ def test_bool_mat_typing(test, device):
...
+
+    wp.launch(kernel, 1, device=device)
+    wp.synchronize_device(device)
+    fn()
```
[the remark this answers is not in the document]
shi-eric: @greptileai The direct `fn()` call is still valuable — it exercises the Python-side `__setitem__` on `vec3bool_type`, which is a separate code path from kernel compilation. Testing both paths is intentional and follows conventions elsewhere in the test suite. Dismissing this one.

warp/tests/test_bool.py
thomasbbrunner: Done!

thomasbbrunner:
Thanks for the review!

> Could you also file a GitHub issue for the original bug

Done (https://github.com/NVIDIA/warp/issues/1302)

> Then add a changelog entry

I've added an entry there.

> One thing missing: the assign_copy builtin isn't exercised by any test.

Good point. I thought this was needed for the assignment, but it's not. Should I remove this from this PR, as it is not relevant for fixing the core bug? Still, I've added a test for this in `test_vec_assign_copy.py`.

I've also rebased onto the latest main and kept the changes to a single commit.

shi-eric:
> Thanks for the review!
> 
> > Could you also file a GitHub issue for the original bug
> 
> Done (#1302)
> 
> > Then add a changelog entry
> 
> I've added an entry there.
> 
> > One thing missing: the assign_copy builtin isn't exercised by any test.
> 
> Good point. I thought this was needed for the assignment, but it's not. Should I remove this from this PR, as it is not relevant for fixing the core bug? Still, I've added a test for this in `test_vec_assign_copy.py`.
> 
> I've also rebased onto the latest main and kept the changes to a single commit.

I think it's great to fix both issues at once since they're related. Changes look good, I'll test and merge this today.

PR #1305, opened 2026-03-18
Add adjoints for `texture_sample` for all address modes

### LINKED ISSUE #1301: [REQ] Add adjoints for `texture_sample`
I am trying to differentiable a dense voxelgrid w.r.t. the sampling position. In the example below, the gradient from warp is zero.

```python
import numpy as np
import warp as wp

wp.init()


@wp.kernel
def sample(tex: wp.Texture1D, pos: wp.array(dtype=float), out: wp.array(dtype=float)):
    tid = wp.tid()
    out[tid] = wp.texture_sample(tex, pos[tid], dtype=float)


data = np.arange(16, dtype=np.float32)
tex = wp.Texture1D(data, normalized_coords=False)
pos = wp.array([4.0], dtype=float, requires_grad=True)
out = wp.zeros(1, dtype=float, requires_grad=True)

tape = wp.Tape()
with tape:
    wp.launch(sample, dim=1, inputs=[tex, pos], outputs=[out])
out.grad = wp.ones(1, dtype=float)
tape.backward()

print("out:", out.numpy())            # [3.5]
print("pos.grad:", pos.grad.numpy())  # [0.] expected [1.]
```

The pytorch version returns the correct gradient (1).

```python
import torch
import torch.nn.functional as F

data = torch.arange(16, dtype=torch.float32).reshape(1, 1, 1, 1, 16)
pos = torch.tensor([4.0], requires_grad=True)

pos_norm = 2.0 * pos / 16.0 - 1.0
grid = torch.stack([pos_norm, torch.zeros_like(pos_norm), torch.zeros_like(pos_norm)], dim=-1)
grid = grid.reshape(1, 1, 1, 1, 3)  # [N, D, H, W, 3]
out = F.grid_sample(data, grid, mode="bilinear", align_corners=False)
out.backward()

print("out:", out.item())             # 3.5
print("pos.grad:", pos.grad.item())   # 1.0
```

Is texture sampling not currently differentiable or did I implement the gradient incorrectly? Also using the latest nightly.

eigenvivek:
## Description

Implements adjoints for `wp.texture_sample()` so gradients propagate to the sampling coordinates and the `lod` argument, on both the CPU and CUDA backends, under **all four address modes** (WRAP, CLAMP, MIRROR, BORDER). Closes #1301.

**Driving workload:** [nanodrr](https://github.com/eigenvivek/nanodrr), a differentiable X-ray renderer for 2D/3D medical-image registration, ray-marches a `wp.Texture3D` (held in a `wp.struct`) and backpropagates to camera poses via `wp.Tape` inside a `torch.library.custom_op`. On upstream Warp its pose gradients are silently zero; I have confirmed this branch addresses that need end-to-end — pose gradients track finite differences and a pose-registration loop recovers a ground-truth camera offset (details in the validation summary below).

This supersedes the earlier BORDER-only implementation in this PR and addresses all outstanding review feedback:

- **All address modes are supported** (previously P1: silent wrong gradients for WRAP/MIRROR/CLAMP, which is the default). Instead of porting PyTorch's per-mode multiplier functions (`clip_coordinates_set_grad` / `reflect_coordinates_set_grad`), the adjoint re-samples the texture through the same forward sampler at the quarter points of the current interpolation cell. Every address-mode coordinate mapping is piecewise affine with folds only at cell boundaries, so the scaled difference of two in-cell samples is the exact derivative of the actual forward function — each mode's boundary behavior (CLAMP's zero slope, MIRROR's sign flip, WRAP's periodicity, BORDER's zero padding) is inherited from the sampler with no mode-specific code.
- **No kernel-side `printf` validation** (review: register waste) — there is no unsupported configuration left to warn about, so the guard is gone entirely, as is the Python-side warning.
- **Boundary bands use zeros-padding gradients** for BORDER ("Option A" from the review thread), matching PyTorch `grid_sample(padding_mode="zeros")`; verified by hand-computed cases in the tests.
- **Rebased onto the current texture API**: the adjoints handle the mipmap/LOD signatures, including per-level coordinate gradients for fractional LODs and the LOD gradient itself.
- Test kernels moved to module scope; `normalized_coords=True` (the default) is covered; the unrelated PUBLICATIONS.md changes are dropped; the changelog entry is a Towncrier fragment (`changelog/1301.added.md`).

Quarter points are used instead of texel centers because they sit a safe distance from floor/fold boundaries and the fractions 0.25/0.75 are exact in the CUDA hardware's 9-bit fixed-point interpolator, so the gradient of the ideal lerp is recovered exactly.

Gradients are **not** propagated to the texture data itself (documented); that needs a scatter-add companion buffer and is left as a separate feature. `CLOSEST` filtering yields zero coordinate gradients (matches `grid_sample(mode="nearest")`).

## Changes

- `warp/native/texture.h`: real `adj_texture_sample` implementations for all five overloads via shared per-axis quarter-point helpers; descriptor structs (`texture{1,2,3}d_t`) gain `filter_mode`, `mip_filter_mode`, `num_mip_levels`, `use_normalized_coords`.
- `warp/_src/texture.py`: ctypes mirrors of the new descriptor fields, populated in `__ctype__` (covers kernel params, texture arrays, and `wp.struct` members).
- `warp/_src/builtins.py`: `texture_sample` overloads flagged `is_differentiable=True`; docstrings document the gradient semantics.
- `warp/tests/cuda/test_texture.py`: adjoint tests (all address modes incl. boundary bands, normalized/unnormalized, multi-channel, 2D/3D scalar and vector overloads, mip integer/fractional LOD for both mip filters, LOD gradients, texture-in-struct).
- `docs/user_guide/runtime.rst`, changelog fragment, regenerated stubs.

## Checklist

## Validation summary

- The new `test_texture_sample_grad_*` tests check the adjoint against central finite differences of Warp's own forward sampler with stencils confined to one interpolation cell (where the forward is exactly linear), across every address mode, both coordinate conventions, 1D/2D/3D, all overloads, multi-channel dtypes, and mipmapped LODs — on CPU and CUDA. `test_texture_sample_grad_1d_analytic` additionally pins hand-computed boundary gradients (cross-checked against PyTorch `grid_sample` semantics: `padding_mode="zeros"` for BORDER, `"border"` for CLAMP).
- On `main` without this change, the issue's repro prints `pos.grad = [0.]`; on this branch it prints `[1.]` on both devices, and the new tests fail on `main` by construction (zero gradients).
- Full `warp/tests/cuda/test_texture.py` passes (151 tests), plus the `TestTape`/`TestGrad`/`TestStruct` suites.
- Docs build clean with `--warnings-as-errors`; doctests pass.
- End-to-end: validated with [nanodrr](https://github.com/eigenvivek/nanodrr) (differentiable X-ray renderer that ray-marches a `wp.Texture3D` held in a `wp.struct` inside a `torch.library.custom_op`): camera-pose gradients through the renderer track finite differences, and a pose-registration loop recovers a ground-truth camera offset (loss reduced ~700x).

## New feature / enhancement

```python
import numpy as np
import warp as wp

wp.init()


@wp.kernel
def sample(tex: wp.Texture1D, pos: wp.array(dtype=float), out: wp.array(dtype=float)):
    tid = wp.tid()
    out[tid] = wp.texture_sample(tex, pos[tid], dtype=float)


data = np.arange(16, dtype=np.float32)
tex = wp.Texture1D(data, normalized_coords=False)
pos = wp.array([4.0], dtype=float, requires_grad=True)
out = wp.zeros(1, dtype=float, requires_grad=True)

tape = wp.Tape()
with tape:
    wp.launch(sample, dim=1, inputs=[tex, pos], outputs=[out])
out.grad = wp.ones(1, dtype=float)
tape.backward()

print("out:", out.numpy())            # [3.5]
print("pos.grad:", pos.grad.numpy())  # [1.]
```

## Summary by CodeRabbit

* **New Features**
  * Added automatic gradients for texture sampling coordinates and mipmap LOD values across 1D, 2D, and 3D textures.
  * Supports normalized and texel coordinates, address modes, linear filtering, multichannel textures, and mipmap blending.
  * Closest filtering produces zero coordinate gradients; texture data does not receive gradients.

* **Documentation**
  * Updated texture sampling documentation and changelog with gradient behavior and limitations.

* **Tests**
  * Added comprehensive gradient validation for texture sampling scenarios.

### DIFF
--- a/docs/user_guide/runtime.rst
+++ b/docs/user_guide/runtime.rst
@@ -2759,6 +2759,12 @@ Supported data types include ``wp.uint8``, ``wp.uint16``, ``wp.uint32``, ``wp.in
 normalized to the [0, 1] range when sampled; signed integer textures are normalized to [-1, 1];
 float types are returned as-is.
 
+Texture sampling is differentiable with respect to the sampling coordinates and the ``lod``
+argument under every address mode, so sampling positions can be optimized with
+:class:`wp.Tape <warp.Tape>`. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
+value is piecewise constant in the coordinates, so their gradients are zero. Gradients are not
+propagated to the texture data itself.
+
 .. seealso:: `Reference <../language_reference/builtins.html#textures>`__ for the texture sampling functions available in kernels.
 
 
--- a/docs/user_guide/runtime.rst
+++ b/docs/user_guide/runtime.rst
@@ -2759,11 +2759,12 @@ Supported data types include ``wp.uint8``, ``wp.uint16``, ``wp.uint32``, ``wp.in
 normalized to the [0, 1] range when sampled; signed integer textures are normalized to [-1, 1];
 float types are returned as-is.
 
-Texture sampling is differentiable with respect to the sampling coordinates and the ``lod``
-argument under every address mode, so sampling positions can be optimized with
-:class:`wp.Tape <warp.Tape>`. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
-value is piecewise constant in the coordinates, so their gradients are zero. Gradients are not
-propagated to the texture data itself.
+Texture sampling is differentiable with respect to sampling coordinates wherever the sample is
+locally linear, and with respect to ``lod`` for linear mip-level blending, so sampling positions
+can be optimized with :class:`wp.Tape <warp.Tape>`. Gradients at interpolation/address-mode
+boundaries and integer LOD kinks follow Warp's branch convention. With closest spatial or mip
+filtering the corresponding coordinate or ``lod`` gradients are zero. Texture data gradients are
+not propagated.
 
 .. seealso:: `Reference <../language_reference/builtins.html#textures>`__ for the texture sampling functions available in kernels.
 
--- a/warp/_src/builtins.py
+++ b/warp/_src/builtins.py
@@ -9828,6 +9828,17 @@ add_builtin(
     group="Utility",
 )
 
+# Bool vector assign_inplace (bool is not part of Scalar)
+add_builtin(
+    "assign_inplace",
+    input_types={"a": vector(length=Any, dtype=bool), "i": Any, "value": Any},
+    value_type=None,
+    dispatch_func=vector_assign_dispatch_func,
+    hidden=True,
+    export=False,
+    group="Utility",
+)
+
 # implements quaternion[index] = value
 add_builtin(
     "assign_inplace",
@@ -9859,6 +9870,17 @@ def vector_assign_copy_value_func(arg_types: Mapping[str, type], arg_values: Map
 add_builtin(
     "assign_copy",
     input_types={"a": vector(length=Any, dtype=Scalar), "i": Any, "value": Any},
+    value_func=vector_assign_copy_value_func,
+    dispatch_func=vector_assign_dispatch_func,
+    hidden=True,
+    export=False,
+    group="Utility",
+)
+
+# Bool vector assign_copy (bool is not part of Scalar)
+add_builtin(
+    "assign_copy",
+    input_types={"a": vector(length=Any, dtype=bool), "i": Any, "value": Any},
     value_func=vector_assign_copy_value_func,
     dispatch_func=vector_assign_dispatch_func,
     hidden=True,
--- a/warp/_src/builtins.py
+++ b/warp/_src/builtins.py
@@ -7809,7 +7809,7 @@ add_builtin(
         The sampled value of the specified ``dtype``.
 
     Filtering mode is :attr:`warp.TextureFilterMode.CLOSEST` or :attr:`warp.TextureFilterMode.LINEAR`.""",
-    is_differentiable=False,
+    is_differentiable=True,
 )
 
 
@@ -7856,7 +7856,7 @@ add_builtin(
         The sampled value of the specified ``dtype``.
 
     Filtering mode is :attr:`warp.TextureFilterMode.CLOSEST` or :attr:`warp.TextureFilterMode.LINEAR`.""",
-    is_differentiable=False,
+    is_differentiable=True,
 )
 
 # texture_sample for 2D textures with separate u, v coordinates
@@ -7886,7 +7886,7 @@ add_builtin(
         The sampled value of the specified ``dtype``.
 
     Filtering mode is :attr:`warp.TextureFilterMode.CLOSEST` or :attr:`warp.TextureFilterMode.LINEAR`.""",
-    is_differentiable=False,
+    is_differentiable=True,
 )
 
 
@@ -7933,7 +7933,7 @@ add_builtin(
         The sampled value of the specified ``dtype``.
 
     Filtering mode is :attr:`warp.TextureFilterMode.CLOSEST` or :attr:`warp.TextureFilterMode.LINEAR`.""",
-    is_differentiable=False,
+    is_differentiable=True,
 )
 
 # texture_sample for 3D textures with separate u, v, w coordinates
@@ -7965,7 +7965,7 @@ add_builtin(
         The sampled value of the specified ``dtype``.
 
     Filtering mode is :attr:`warp.TextureFilterMode.CLOSEST` or :attr:`warp.TextureFilterMode.LINEAR`.""",
-    is_differentiable=False,
+    is_differentiable=True,
 )
 
 
@@ -9828,17 +9828,6 @@ add_builtin(
     group="Utility",
 )
 
-# Bool vector assign_inplace (bool is not part of Scalar)
-add_builtin(
-    "assign_inplace",
-    input_types={"a": vector(length=Any, dtype=bool), "i": Any, "value": Any},
-    value_type=None,
-    dispatch_func=vector_assign_dispatch_func,
-    hidden=True,
-    export=False,
-    group="Utility",
-)
-
 # implements quaternion[index] = value
 add_builtin(
     "assign_inplace",
@@ -9870,17 +9859,6 @@ def vector_assign_copy_value_func(arg_types: Mapping[str, type], arg_values: Map
 add_builtin(
     "assign_copy",
     input_types={"a": vector(length=Any, dtype=Scalar), "i": Any, "value": Any},
-    value_func=vector_assign_copy_value_func,
-    dispatch_func=vector_assign_dispatch_func,
-    hidden=True,
-    export=False,
-    group="Utility",
-)
-
-# Bool vector assign_copy (bool is not part of Scalar)
-add_builtin(
-    "assign_copy",
-    input_types={"a": vector(length=Any, dtype=bool), "i": Any, "value": Any},
     value_func=vector_assign_copy_value_func,
     dispatch_func=vector_assign_dispatch_func,
     hidden=True,
--- a/warp/_src/builtins.py
+++ b/warp/_src/builtins.py
@@ -10997,6 +10997,11 @@ add_builtin(
     :class:`warp.Texture`. On CUDA, ``WRAP`` and ``MIRROR`` are treated as ``CLAMP`` when
     ``normalized_coords=False`` (the CPU sampler honors them).
 
+    The backward pass propagates gradients to the sampling coordinates and ``lod`` under
+    every address mode. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
+    value is piecewise constant in the coordinates, so their gradients are zero. Gradients
+    are not propagated to the texture data itself.
+
     Example:
 
         .. testcode::
@@ -11024,7 +11029,7 @@ add_builtin(
         .. testoutput::
 
             5.0 1.0 1.0""",
-    is_differentiable=False,
+    is_differentiable=True,
 )
 
 
@@ -11093,8 +11098,13 @@ add_builtin(
     The filtering mode (:class:`warp.TextureFilterMode`) and the addressing of out-of-range
     coordinates (:class:`warp.TextureAddressMode`) are those set when the texture was created; see
     :class:`warp.Texture`. On CUDA, ``WRAP`` and ``MIRROR`` are treated as ``CLAMP`` when
-    ``normalized_coords=False`` (the CPU sampler honors them).""",
-    is_differentiable=False,
+    ``normalized_coords=False`` (the CPU sampler honors them).
+
+    The backward pass propagates gradients to the sampling coordinates and ``lod`` under
+    every address mode. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
+    value is piecewise constant in the coordinates, so their gradients are zero. Gradients
+    are not propagated to the texture data itself.""",
+    is_differentiable=True,
 )
 
 # texture_sample for 2D textures with separate u, v coordinates
@@ -11147,8 +11157,13 @@ add_builtin(
     The filtering mode (:class:`warp.TextureFilterMode`) and the addressing of out-of-range
     coordinates (:class:`warp.TextureAddressMode`) are those set when the texture was created; see
     :class:`warp.Texture`. On CUDA, ``WRAP`` and ``MIRROR`` are treated as ``CLAMP`` when
-    ``normalized_coords=False`` (the CPU sampler honors them).""",
-    is_differentiable=False,
+    ``normalized_coords=False`` (the CPU sampler honors them).
+
+    The backward pass propagates gradients to the sampling coordinates and ``lod`` under
+    every address mode. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
+    value is piecewise constant in the coordinates, so their gradients are zero. Gradients
+    are not propagated to the texture data itself.""",
+    is_differentiable=True,
 )
 
 
@@ -11218,8 +11233,13 @@ add_builtin(
     The filtering mode (:class:`warp.TextureFilterMode`) and the addressing of out-of-range
     coordinates (:class:`warp.TextureAddressMode`) are those set when the texture was created; see
     :class:`warp.Texture`. On CUDA, ``WRAP`` and ``MIRROR`` are treated as ``CLAMP`` when
-    ``normalized_coords=False`` (the CPU sampler honors them).""",
-    is_differentiable=False,
+    ``normalized_coords=False`` (the CPU sampler honors them).
+
+    The backward pass propagates gradients to the sampling coordinates and ``lod`` under
+    every address mode. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
+    value is piecewise constant in the coordinates, so their gradients are zero. Gradients
+    are not propagated to the texture data itself.""",
+    is_differentiable=True,
 )
 
 # texture_sample for 3D textures with separate u, v, w coordinates
@@ -11277,8 +11297,13 @@ add_builtin(
     The filtering mode (:class:`warp.TextureFilterMode`) and the addressing of out-of-range
     coordinates (:class:`warp.TextureAddressMode`) are those set when the texture was created; see
     :class:`warp.Texture`. On CUDA, ``WRAP`` and ``MIRROR`` are treated as ``CLAMP`` when
-    ``normalized_coords=False`` (the CPU sampler honors them).""",
-    is_differentiable=False,
+    ``normalized_coords=False`` (the CPU sampler honors them).
+
+    The backward pass propagates gradients to the sampling coordinates and ``lod`` under
+    every address mode. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled
+    value is piecewise constant in the coordinates, so their gradients are zero. Gradients
+    are not propagated to the texture data itself.""",
+    is_differentiable=True,
 )
 
 
--- a/warp/_src/builtins.py
+++ b/warp/_src/builtins.py
@@ -10988,6 +10988,9 @@ add_builtin(
             values blend between neighbouring mip levels when ``mip_filter_mode`` is
             :attr:`warp.TextureFilterMode.LINEAR`; the coordinate is evaluated independently at
             each level used in the blend. Ignored for textures created with a single mip level.
+            The ``lod`` adjoint is nonzero only for mipmapped textures with
+            ``mip_filter_mode=TextureFilterMode.LINEAR`` and an unclamped, nonnegative LOD;
+            integer LODs are mip-blend kinks where Warp uses the forward branch convention.
 
     Returns:
         The sampled value of the specified ``dtype``.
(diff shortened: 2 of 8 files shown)


warp/native/texture.h
```diff
@@ -785,34 +890,213 @@ template <typename T> CUDA_CALLABLE T texture_sample(const texture3d_t& tex, flo
...
+
+    if (x0 >= 0 && x1 < tex.width) {
+        float u0 = tex.use_normalized_coords ? ((float)x0 + 0.5f) / (float)tex.width : (float)x0 + 0.5f;
+        float u1 = tex.use_normalized_coords ? ((float)x1 + 0.5f) / (float)tex.width : (float)x1 + 0.5f;
+        float gtx = 0.0f;
+        for (int c = 0; c < tex.num_channels; c++)
+            gtx += (texture_sample_helper<T>::fetch_1d(tex.tex, u1, c)
+                    - texture_sample_helper<T>::fetch_1d(tex.tex, u0, c))
+                * ((const float*)&adj_ret)[c];
+        adj_u += gtx_mult * gtx;
```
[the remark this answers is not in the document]
eigenvivek: This was done to perfectly match pytorch's grid sample. Happy to implement either.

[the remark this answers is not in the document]
eigenvivek: A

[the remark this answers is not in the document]
eigenvivek: Sure, implement these changes.

warp/native/texture.h
```diff
@@ -785,34 +890,213 @@ template <typename T> CUDA_CALLABLE T texture_sample(const texture3d_t& tex, flo
...
 template <typename T>
 CUDA_CALLABLE void adj_texture_sample(
     const texture2d_t& tex, float u, float v, texture2d_t& adj_tex, float& adj_u, float& adj_v, const T& adj_ret
 )
 {
-    // Texture sampling is not differentiable in this implementation
+    if (tex.filter_mode == WP_TEXTURE_FILTER_CLOSEST)
+        return;
+
+    if (tex.address_mode_u != WP_TEXTURE_ADDRESS_BORDER || tex.address_mode_v != WP_TEXTURE_ADDRESS_BORDER) {
```
shi-eric: Generally, we don't do any error checking in kernels since the print messages waste registers. If it really can't be checked at the Python scope instead, we can issue printf's when the kernel is run in a debug mode.

warp/tests/cuda/test_texture.py
```diff
@@ -2600,6 +2600,462 @@ def test_texture3d_array(test, device):
...
+        normalized_coords=False,
+        filter_mode=wp.TextureFilterMode.LINEAR,
+        address_mode=wp.TextureAddressMode.BORDER,
+        device=device,
+    )
+
+    @wp.kernel
+    def sample_2d_vec2(tex: wp.Texture2D, pos: wp.array(dtype=wp.vec2f), out: wp.array(dtype=wp.vec2f)):
+        tid = wp.tid()
+        out[tid] = wp.texture_sample(tex, pos[tid], dtype=wp.vec2f)
```
shi-eric: An minor optimization is to move the `@wp.kernel` definitions out of the test function scope and into the module scope to prevent the module's hash from changing multiple times during test execution and resulting in multiple unnecessary recompilations.

shi-eric:
Please make sure there is a `CHANGELOG.md` entry that reflects the change

shi-eric:
Closing this PR due to inactivity. There have been no code updates since March
19, the branch now conflicts with `main`, and the texture API has changed since
this implementation. The outstanding review feedback also remains unresolved.

The feature request remains open in #1301. If you would like to continue this
work, update the branch against the current `main`, address the outstanding
feedback, and comment here so we can reopen the PR.

eigenvivek:
@shi-eric I've updated the branch as suggested — could you reopen this PR?

The branch is merged up to the current `main` and the implementation is new, superseding the BORDER-only approach. How each outstanding review point was addressed:

- **Silent wrong gradients for WRAP/MIRROR/CLAMP (P1)**: all four address modes are now supported with correct gradients on both backends. The adjoint re-samples through the same forward sampler at the quarter points of the current interpolation cell; since every address-mode mapping is piecewise affine with folds only at cell boundaries, the scaled in-cell sample difference is the exact derivative of the actual forward function under any mode — no mode-specific code, and no unsupported configuration left.
- **Kernel-side `printf` / Python warning**: removed entirely — nothing to warn about anymore.
- **Boundary bands (Option A)**: BORDER now produces PyTorch-style zeros-padding gradients in the half-texel bands, pinned by hand-computed cases in `test_texture_sample_grad_1d_analytic`.
- **Texture API drift**: the adjoints implement the current mipmap/LOD signatures, including per-level coordinate gradients for fractional LODs and `adj_lod`.
- **Test hygiene**: kernels at module scope, `normalized_coords=True` covered, gradients verified against single-cell finite differences of Warp's own forward across modes/dims/dtypes/LODs on CPU and CUDA.
- **Unrelated PUBLICATIONS.md changes**: dropped.
- **Changelog**: added as a Towncrier fragment (`changelog/1301.added.md`) per the current convention.

The PR description has been updated with the full details and validation summary (including an end-to-end differentiable X-ray registration check). Once reopened, I'll squash the branch into a single clean commit per the contribution guidelines.

shi-eric:
@eigenvivek welcome back. Thanks for adopting the branch to our new processes.

@daedalus5 can you help me review this?

shi-eric:
@eigenvivek Also if you have a workload that is driving this feature and you have confirmed that the changes here have been confirmed to address your needs, it would be helpful to note in the PR description to give some more context.

eigenvivek:
Thanks @shi-eric! I've added the driving workload to the top of the PR description: [nanodrr](https://github.com/eigenvivek/nanodrr), a differentiable X-ray renderer for 2D/3D medical-image registration that backpropagates to camera poses through a ray-marched `wp.Texture3D`. I've confirmed end-to-end that this branch addresses that need (pose gradients track finite differences, and a registration loop recovers a ground-truth camera offset).

Also pushed 8fab181 addressing CodeRabbit's follow-up review: the 2D/3D gradient tests now cover WRAP and MIRROR too (using normalized coordinates, since CUDA treats those modes as CLAMP otherwise), the mip finite-difference tolerance is a named constant, and the integer-LOD skip is documented. One subtlety worth noting: the new 2D/3D sample points are quarter-aligned so every interpolation weight — including off-axis ones — is exact in the CUDA sampler's fixed-point format; a misaligned off-axis weight couples through the cell's mixed derivative and the half-texel stencil amplifies it into percent-level disagreement between the adjoint and the finite-difference reference (both of which are otherwise exact).

warp/_src/builtins.py
```diff
@@ -10997,6 +10997,11 @@ def texture_sample_1d_dispatch_func(input_types: Mapping[str, type], return_type
     :class:`warp.Texture`. On CUDA, ``WRAP`` and ``MIRROR`` are treated as ``CLAMP`` when
     ``normalized_coords=False`` (the CPU sampler honors them).
 
+    The backward pass propagates gradients to the sampling coordinates and ``lod`` under
```
daedalus5: Please tighten this backward-pass contract before regenerating `warp/__init__.pyi`. As written, this sounds unconditional, but the implementation is branch-defined at interpolation/address-mode boundaries, and `lod` gradients are only meaningful for linear mip-level blending. Suggested wording:

```rst
The backward pass propagates gradients to the sampling coordinates wherever the sample is locally linear. At interpolation-cell and address-mode boundaries the mathematical derivative is not unique, so Warp follows the branch selected by the forward sampler. With :attr:`warp.TextureFilterMode.CLOSEST` filtering the sampled value is piecewise constant in the coordinates, so coordinate gradients are zero. Gradients are not propagated to the texture data itself.
```

Please also add to the `lod` argument docs:

```rst
The ``lod`` adjoint is nonzero only for mipmapped textures with ``mip_filter_mode=TextureFilterMode.LINEAR`` and an unclamped, nonnegative LOD; integer LODs are mip-blend kinks where Warp uses the forward branch convention.
```

Apply the same contract wording to the other `texture_sample` overloads and regenerate the stubs.

docs/user_guide/runtime.rst
```diff
@@ -2759,6 +2759,12 @@ Supported data types include ``wp.uint8``, ``wp.uint16``, ``wp.uint32``, ``wp.in
 normalized to the [0, 1] range when sampled; signed integer textures are normalized to [-1, 1];
 float types are returned as-is.
 
+Texture sampling is differentiable with respect to the sampling coordinates and the ``lod``
```
daedalus5: Please make the user-guide contract match the implementation's edge cases. This paragraph currently says gradients propagate to coordinates and `lod` under every address mode, but exact interpolation/address-mode boundaries are nondifferentiable, and `mip_filter_mode=CLOSEST`, omitted/negative `lod`, clamped LODs, or single-mip textures do not have a useful LOD gradient. Suggested replacement:

```rst
Texture sampling is differentiable with respect to sampling coordinates wherever the sample is locally linear, and with respect to ``lod`` for linear mip-level blending. Gradients at interpolation/address-mode boundaries and integer LOD kinks follow Warp's branch convention. With closest spatial or mip filtering the corresponding coordinate or ``lod`` gradients are zero. Texture data gradients are not propagated.
```

daedalus5 (changes requested):
The implementation looks solid and the texture-gradient tests pass, but please tighten the public documentation around the adjoint contract before merge. The current wording makes the gradients sound unconditional even at interpolation/address-mode boundaries and for every mip-filter configuration.

eigenvivek:
Thanks @daedalus5 — agreed on both points; the suggested wording is more faithful to what the implementation actually does. Adopted it nearly verbatim in e0492ff:

- The backward-pass contract paragraph now states gradients hold where the sample is locally linear, with the forward sampler's branch convention at interpolation-cell/address-mode boundaries, in all five `texture_sample` overload docstrings.
- The `lod` argument docs now state the adjoint is nonzero only for mipmapped textures with linear mip blending and an unclamped, nonnegative LOD, with integer LODs following the branch convention.
- The user-guide paragraph in `runtime.rst` uses your suggested replacement, and the stubs (`warp/__init__.pyi`) are regenerated.

warp/_src/builtins.py
daedalus5, line 11098: Needs lod caveat for 2d and 3d cases as well, then re-build docs.

daedalus5 (commented):
Looking good, just the one change requested. Please re-base as well. Afterwards we will test then get this integrated.

warp/native/texture.h
```diff
@@ -891,31 +960,230 @@ template <typename T> CUDA_CALLABLE T texture_sample(const texture3d_t& tex, flo
...
 {
-    // MISSINGADJOINT: differentiable for linear interpolation;
-    // route adj_ret to neighboring texels and mip levels (weighted by interpolation factors)
-    // and to adj_u/adj_lod (via the interpolation derivatives along U and LOD).
+    const int level_w = texture_mip_dim(tex.width, level);
+    const float su = texture_coord_scale(tex.use_normalized_coords, level_w, tex.width);
+
+    float ua, ub;
+    texture_grad_cell_points(u, su, ua, ub);
+    const T va = texture_sample_helper<T>::sample_1d(tex, ua, sample_lod);
```
shi-eric, line 1035: Consider returning early when `fl == 0.0f`, before sampling `level1`:

```cpp
float v0 = cpu_sample_3d_channel_at_level(tex, level0, u, v, w, channel);
if (level1 == level0 || fl == 0.0f)
    return v0;
```

Applied to the 1D, 2D, and 3D helpers, this reduced the 3D mip backward benchmark from 28.178 ms to 17.559 ms, a 10.619 ms or 37.69% improvement.

warp/native/texture.h
```diff
@@ -157,18 +157,40 @@ struct texture1d_t {
     uint64 tex;  // CUtexObject handle (GPU) or Texture* (CPU)
     int32 width;
     int32 num_channels;
+    // Sampling configuration mirrored from the owning texture so that
+    // adj_texture_sample can differentiate without host-side lookups.
+    int32 filter_mode;
```
shi-eric, line 162: Consider packing this metadata into a `uint32` flag word and mirroring the layout in the Python `ctypes` descriptors.

This reduced the 1D descriptor from 32 to 16 bytes and the 2D/3D descriptors from 40 to 24 bytes. CUDA texture-array sampling improved from 0.408 ms to 0.316 ms, or 22.52%.

shi-eric: To clarify, `num_channels` is pre-existing. The 16/24/24-byte result comes from using the unused bits in that existing 32-bit word for the new sampling metadata, rather than adding a separate flag word. A separate word would produce 24/24/32-byte descriptors because of alignment.

docs/user_guide/runtime.rst
```diff
@@ -2759,6 +2759,13 @@ Supported data types include ``wp.uint8``, ``wp.uint16``, ``wp.uint32``, ``wp.in
 normalized to the [0, 1] range when sampled; signed integer textures are normalized to [-1, 1];
 float types are returned as-is.
 
+Texture sampling is differentiable with respect to sampling coordinates wherever the sample is
+locally linear, and with respect to ``lod`` for linear mip-level blending, so sampling positions
+can be optimized with :class:`wp.Tape <warp.Tape>`. Gradients at interpolation/address-mode
+boundaries and integer LOD kinks follow Warp's branch convention. With closest spatial or mip
```
shi-eric, line 2765: Instead of saying "Warp's branch convention" , how about something like:

> At boundaries where the derivative is not unique, Warp differentiates the branch selected by the forward sample. At an unclamped integer LOD, linear mip filtering uses the right-hand derivative. The LOD gradient is zero in the clamped ranges.

warp/_src/builtins.py
```diff
@@ -10988,6 +10988,9 @@ def texture_sample_1d_dispatch_func(input_types: Mapping[str, type], return_type
             values blend between neighbouring mip levels when ``mip_filter_mode`` is
             :attr:`warp.TextureFilterMode.LINEAR`; the coordinate is evaluated independently at
             each level used in the blend. Ignored for textures created with a single mip level.
+            The ``lod`` adjoint is nonzero only for mipmapped textures with
+            ``mip_filter_mode=TextureFilterMode.LINEAR`` and an unclamped, nonnegative LOD;
+            integer LODs are mip-blend kinks where Warp uses the forward branch convention.
```
shi-eric, line 10993: I would suggest less jargon, e.g.

> At an unclamped integer LOD, the mip-blend derivative may differ on either side. Warp uses the derivative toward increasing LOD values. The LOD gradient is zero when the LOD is clamped.

warp/tests/cuda/test_texture.py
```diff
@@ -2825,6 +2825,463 @@ def test_texture_mipmap_invalid_levels(test, device):
         wp.Texture2D(data=data, num_mip_levels=-1, device=device)
 
 
+# ============================================================================
+# Adjoint (Gradient) Tests
+# ============================================================================
+#
+# texture_sample gradients are checked against central finite differences of
```
shi-eric, line 2832: The algorithm is already explained in detail in `texture.h`. Consider treating that as the canonical explanation and shortening this block to the details specific to these tests, so the two descriptions do not drift. For example:

```python
# Compare texture_sample adjoints with central finite differences. Sample points
# keep the stencil inside one interpolation cell and exercise every address mode
# on CPU and CUDA.

# CUDA quantizes interpolation fractions to 9 bits. Quarter-aligned points keep
# the interpolation weights exact and reduce finite-difference noise.
_GRAD_FD_TOL = 2e-2

# Mipmapped sampling adds another interpolation stage.
_GRAD_MIP_FD_TOL = 3e-2
```

warp/native/texture.h
```diff
@@ -891,31 +960,230 @@ template <typename T> CUDA_CALLABLE T texture_sample(const texture3d_t& tex, flo
     return texture_sample_helper<T>::sample_3d(tex, u, v, w, lod);
 }
 
-// Adjoint stubs for texture sampling
+// ============================================================================
+// Texture Sampling Adjoints
+// ============================================================================
+//
+// Gradients with respect to the sampling coordinates (and LOD) are computed by
```
shi-eric, line 967: This comment combines two different calculations. The quarter-point samples are used for coordinate gradients. The LOD gradient comes from the difference between adjacent mip levels:

```cpp
sample(level1) - sample(level0)
```

Please describe these separately.

warp/_src/builtins.py
eigenvivek: Done in 4eead43: all five `texture_sample` overloads (1D, both 2D, both 3D) now carry the `lod` adjoint caveat, and I regenerated `__init__.pyi` and rebuilt the docs with no warnings.

warp/native/texture.h
eigenvivek: Done in 4eead43: the 1D, 2D, and 3D CPU mip-blend helpers now return `v0` when `level1 == level0 || fl == 0.0f`, so the adjoint's integer-LOD samples no longer read a second level.

warp/native/texture.h
eigenvivek: Done in 4eead43. The channel count, `dtype` (added on `main` by GH-1731), both filter modes, the normalized-coordinates flag, and the mip-level count now share one 32-bit `info` word (layout: `WP_TEXTURE_INFO_*` in `texture.h`, mirrored by `_pack_texture_info()` in `texture.py`). `ctypes.sizeof` gives 16/24/24 bytes for 1D/2D/3D. The Python side builds the word once when the texture is created, not on every launch.

docs/user_guide/runtime.rst
eigenvivek: Adopted your wording in 4eead43.

warp/_src/builtins.py
eigenvivek: Adopted in 4eead43, in all five overloads.

warp/tests/cuda/test_texture.py
eigenvivek: Done in 4eead43: the test block is now a short summary plus the two tolerance notes, and it points to the adjoint notes in `texture.h` for the full explanation.

warp/native/texture.h
eigenvivek: Done in 4eead43: the adjoint notes now describe the coordinate gradient (quarter-point samples) and the LOD gradient (`sample(level1) - sample(level0)` at the original coordinates) in separate paragraphs, and the `texture_grad_lod_levels` comment notes that the derivative toward increasing LOD values is used at integer LODs.

eigenvivek:
@daedalus5 @shi-eric Rebased onto current `main` and addressed the review feedback in 4eead43 (squashed into one commit, since the earlier commits all conflicted with GH-1731's descriptor changes). Replies are on each thread.

One more change from the rebase: GH-1811 added a workaround for divergent texture handles on CUDA 12 sm_101/sm_120. The adjoint was re-sampling through `texture_sample_helper` directly, so it skipped that workaround. It now goes through `texture_sample()`, so the backward pass gets the same protection as the forward pass. I couldn't test that path locally (sm_89 only).

Verified on CPU and cuda:0 (RTX 6000 Ada, sm_89, CUDA 13.0): all `TestTexture` tests (including the 32-bit integer rejection tests from `main`), plus `TestTextureCompiler`, `TestStruct`, `TestCodeGen`, and `TestFactoryStyleArrayAnnotations`. The docs build has no warnings. I did not re-run your benchmarks.

PR #1328, opened 2026-03-30
Add support for indexed arrays in structs and related kernels

### LINKED ISSUE #1327: Support `indexedarray` fields in `struct`
`wp.indexedarray` cannot currently be used as a field in `wp.struct` (assignment/device transfer/NumPy structured values) while it can be passed as an argument to a `wp.kernel`. There is no apparent reason not to support this possibility as all the blocks are already in place.

Minimal example for currently erroneous behavior:

```python
import warp as wp

@wp.struct
class Foo:
    arr: wp.indexedarray(dtype=wp.float32)
```

Result:

```
Traceback (most recent call last):
  File "examples/tmp.py", line 4, in <module>
    @wp.struct
     ^^^^^^^^^
  File ".pixi/envs/default/lib/python3.12/site-packages/warp/_src/context.py", line 1451, in struct
    return wrapper(c)
           ^^^^^^^^^^
  File ".pixi/envs/default/lib/python3.12/site-packages/warp/_src/context.py", line 1443, in wrapper
    s = warp._src.codegen.Struct(key=warp._src.codegen.make_full_qualified_name(c), cls=c, module=m)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".pixi/envs/default/lib/python3.12/site-packages/warp/_src/codegen.py", line 522, in __init__
    elif issubclass(var.type, ctypes.Array):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: issubclass() arg 1 must be a class
```

Expected:
Allow `wp.indexedarray` fields in structs with the same behavior as `wp.array`:
- assign `indexedarray` (or `None`) to struct fields
- `StructInstance.to(device)` moves indexedarray members
- works in kernels, nested structs, and struct arrays
- `Struct.numpy_dtype()` / `StructInstance.numpy_value()` work

FabienPean-Virtonomy:
## Description

This PR adds support for `indexedarray` within `struct`.

Closes #1327

## Checklist

## Test plan

Tests were added in `warp/tests/test_indexedarray.py`

## New feature / enhancement

This now works

```python
import warp as wp

@wp.struct
class Foo:
    arr: wp.indexedarray(dtype=wp.float32)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Proper support for indexedarray fields inside structs: assignment, device transfers (including cross-device/CUDA behavior), and NumPy structured-value conversion; works for nested structs and arrays-of-structs.

* **Tests**
  * Added tests exercising indexedarrays in structs, nested structs, struct arrays, NumPy conversion, and device-transfer scenarios.

* **Documentation**
  * Changelog updated to document indexedarray support in struct operations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@
   based on occupancy ([GH-1270](https://github.com/NVIDIA/warp/issues/1270)).
 - Add `module_options` dict parameter to `@wp.kernel` for inline module-level compilation options
   on `"unique"` modules ([GH-1250](https://github.com/NVIDIA/warp/issues/1250)).
+- Add support for `wp.indexedarray` fields in `@wp.struct` (assignment, device transfer, and NumPy structured values)
+  ([GH-1327](https://github.com/NVIDIA/warp/issues/1327)).
 
 ### Removed
 
--- a/warp/_src/codegen.py
+++ b/warp/_src/codegen.py
@@ -237,6 +237,11 @@ class StructInstance:
             if matches_array_class(var.type, array):
                 # array_t
                 setattr(dst, name, value.to(device))
+            elif matches_array_class(var.type, indexedarray):
+                # indexedarray_t
+                # `.to` returns an array if on different device, force to identity indexedarray
+                cloned = value.to(device)
+                setattr(dst, name, cloned if isinstance(cloned, indexedarray) else indexedarray(cloned))
             elif isinstance(var.type, Struct):
                 # nested struct
                 new_struct = var.type()
@@ -263,6 +268,9 @@ class StructInstance:
 
             if matches_array_class(var.type, array):
                 # array_t
+                npvalue.append(value.numpy_value())
+            elif matches_array_class(var.type, indexedarray):
+                # indexedarray_t
                 npvalue.append(value.numpy_value())
             elif isinstance(var.type, Struct):
                 # nested struct
@@ -299,6 +307,8 @@ def _make_struct_field_constructor(field: str, var_type: type):
         return lambda ctype: var_type.instance_type(ctype=getattr(ctype, field))
     elif matches_array_class(var_type, warp._src.types.array):
         return lambda ctype: None
+    elif matches_array_class(var_type, warp._src.types.indexedarray):
+        return lambda ctype: None
     elif _is_texture_type(var_type):
         return lambda ctype: None
     elif issubclass(var_type, ctypes.Array):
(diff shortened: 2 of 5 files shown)


warp/_src/codegen.py
```diff
@@ -330,6 +330,9 @@ def to(self, device):
             if matches_array_class(var.type, array):
                 # array_t
                 setattr(dst, name, value.to(device))
+            elif matches_array_class(var.type, indexedarray):
+                # indexedarray_t
+                setattr(dst, name, value.to(device))
```
nvlukasz: Note that `indexedarray.to()` will create a regular `array` if it needs to be copied to a different device.

We _could_ consider doing a full clone of data + indices to create an `indexedarray`, but I think that's beyond the scope of this PR. So in the meantime, we could raise an exception if the device does not match.

nvlukasz (commented):
Thanks, this looks pretty good. Just a few minor comments worth addressing.

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@
 - Fix `@wp.func` with tile parameters failing to compile when called with shared-memory tiles. Tile parameters in
   `@wp.func` are now passed by reference for both register and shared storage, matching Python's semantics for mutable
   objects. Previously, register tiles were passed by value ([GH-1313](https://github.com/NVIDIA/warp/issues/1313)).
+- Fix `@wp.struct` packing to support `wp.indexedarray` fields (assignment, device transfer, and NumPy structured values).
```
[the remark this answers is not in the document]
shi-eric: CodeRabbit is correct here. The changelog entry should link to the issue you filed.

warp/_src/codegen.py
FabienPean-Virtonomy: Ah indeed, and the test only test same device copy. I find the shrinking to an array a good approach in general, I would prefer to leave it up to the user to consider where they store the backing array if they care about reuse or optimization.

An easy way around would be to use the passthrough property of `indexedarray_t` when `indices=None` as it would behave like an `array_t` if I read right.
So it means modifying either at codegen point above, or adding method `to` on python side `indexedarray`, or modifying noncontiguous base `to` method. Would any of the three be ok with you, are there issues I did not foresee?

CHANGELOG.md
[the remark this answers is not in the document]
FabienPean-Virtonomy: Thanks it was lost in translation when organizing the PR. Moved it to `added` section as I do not think it is a bug per say. I have difficulty to judge the validity of the other AI comments though.

warp/_src/codegen.py
FabienPean-Virtonomy: The different possibilities:

Option 1: at codegen, coerce to indexedarray only in struct packing

```python
setattr(dst, name, value.to(device) if value.device is device else wp.indexedarray(value.to(device))
```

---

Option 2: in indexedarray, always return an indexedarray and discards return array path, which means a pointer check for index access compared to array path

```python
    def to(self, device):
        device = warp.get_device(device)
        if self.device == device:
            return self

        dense = warp.clone(self, device=device)
        return indexedarray(data=dense, indices=None)
```

---

Option 3: in noncontiguousbase, not applicable due to other child class `indexedfabricarray`

```python
# types.py L4437
return type(self)(warp.clone(self, device=device))
```

shi-eric:
@FabienPean-Virtonomy: @nvlukasz is on travel this week so to set your expectations, I wouldn't expect a re-review until next week. Thanks for your patience.

FabienPean-Virtonomy:
Thanks for the heads-up ! The PR is stabilized, and will be awaiting for further feedback.

I went on enabling `indexedarray` to be transferred to different device via struct, by wrapping the resulting array into a default passthrough `indexedarray` (Option 1 in comment above https://github.com/NVIDIA/warp/pull/1328#discussion_r3016857633)

About the remaining comment from coderabbits:

> Guard nullable indexedarray fields in StructInstance.to()

It was not done for array just above so I don't think it is applicable.

> Clear indexedarray grad keepalive when assigning None

Not familiar enough with the AD system about the need for this section or requirement to clear the attributes (as it seems not done for array above):
 https://github.com/NVIDIA/warp/pull/1328/changes#diff-4509963cccfb2a792b7af2b720eeec5e8a3a87991a12e605522b73fbd752c2f1R352-R359

warp/_src/codegen.py
nvlukasz: Option 1 is a good one, thanks!

nvlukasz:
Thanks @FabienPean-Virtonomy, this should be merged shortly once our internal CI pipelines pass.

shi-eric:
Thanks @FabienPean-Virtonomy this was merged in 537ae37fda6ff5ffbbc573ac8a5c1fd88583f347

Issue #1270, opened 2026-03-05
[REQ] Enhance ease of use for grid-stride loop


lmtss:
### Description
* Provide an interface similar to `cudaOccupancyMaxActiveBlocksPerMultiprocessor` to facilitate max_blocks estimation
* Add an option to disable the grid-stride loop

### Context
I noticed that the grid-stride loop is used by default in cuda_kernel_template_forward. I have a basic understanding of its benefits, but I've encountered some issues:
* The grid-stride loop is enabled by default, but in cases where I'm certain it's unnecessary, disabling it could provide a slight performance improvement (possibly due to better register usage).
* It's not trivial to estimate an appropriate max_blocks value — I typically need to use tools like Nsight Compute to determine how many blocks can run on an SM before arriving at the final max_blocks setting.
* The grid-stride loop uses size_t (64-bit integer), which may introduce unnecessary 64-bit arithmetic in some cases.

shi-eric:
Thanks for the ideas, I'll do some exploring and see what we can do for you.

adenzler-nvidia:
Hi @lmtss, thanks for the detailed issue. We've been investigating the performance impact of both `wp.tid()` and the grid-stride loop. These are two separate concerns, so I'll address them individually.

## 1. Grid-stride loop overhead

We prototyped a `grid_stride=False` kernel option that replaces the for-loop with a simple if-guard:

```python
@wp.kernel(grid_stride=False)
def my_kernel(out: wp.array(dtype=float)):
    i = wp.tid()
    out[i] = float(i)
```

On trivial microbenchmarks (CUDA graph replay, Blackwell), the if-guard is consistently ~1.5–2.5% faster than the for-loop. On real workloads (mujoco_warp), we saw up to ~5% throughput improvement on benchmarks that launch many small kernels per step, though many cases showed negligible gains — the benefit scales inversely with how much actual work each kernel does.

Since the grid-stride loop only matters when `max_blocks` constrains the grid (the default is effectively unlimited), disabling it for the common case is a clean win with no functional downside.

That said, if your workload is dominated by many small kernels, the best optimization advice is to try to fuse them into fewer, larger kernels where possible — that eliminates not just the grid-stride overhead but also launch overhead, memory traffic from intermediates, etc. The `grid_stride=False` option helps at the margin, but kernel fusion is where the real wins are.

## 2. `wp.tid()` cost (coordinate decomposition)

[...]

PR #1390, opened 2026-04-23
Hard-error on wp.launch() dim-rank mismatch

### LINKED ISSUE #1389: Alert when `wp.launch()` dim has more dimensions than the kernel's `wp.tid()` unpacks
### Background

[GH-1270](https://github.com/NVIDIA/warp/issues/1270) templated `launch_bounds_t<N>` on a per-kernel `kernel_dim` inferred from how `wp.tid()` is unpacked in the kernel AST. As a side effect, when the user calls `wp.launch()` with a `dim=` whose rank exceeds `kernel_dim`, Warp's [`_normalize_launch_dim`](warp/_src/context.py#L8111-L8130) silently folds the excess dimensions into the last retained slot. For scalar-`tid` kernels this collapses to a flat linear index; for 2-D kernels it collapses trailing dims into the second slot; etc.

The CHANGELOG calls this out as a breaking change and suggests `i, _ = wp.tid()` as the migration pattern for users who relied on the old unravel behavior. The behavior is deliberate and documented.

However, the silent fold is a real footgun. See [newton-physics/newton#2546](https://github.com/newton-physics/newton/pull/2546) for a concrete instance: a kernel used `tid = wp.tid()` with `dim=joint_f.shape` (a 3-tuple). Under the old semantics this dispatched `nworld * 1 * ndofs` threads with `tid = coord.i` ∈ `[0, nworld)` repeated — redundant but idempotent, so it silently "worked." Under the new semantics Warp flattens the launch to `nworld * 1 * ndofs` threads with `tid` as the flat linear index, so `joint_q[tid, 0, 0]` walks past the first-dim bound of the 3-D array. The result is intermittent access violations, heap corruption detected later inside `wp_lookup` when the damage lands on NT-heap LFH metadata, or silent numerical corruption when the damage lands in a neighbouring Warp array. This was diagnosed over several days with faulthandler traces that kept pointing at detection sites rather than the corruption site, because the overflow accumulates until the Windows heap validator notices.

The important observation from that debug session: **Newton never actually wanted the fold behavior**. It wanted first-dim indexing only, accidentally got correct but repeated indexes under old semantics via the unravel form, and had no way to notice when the semantics changed except through hard-to-reproduce memory-corruption crashes downstream. The silent fold hid the migration mistake long enough to corrupt memory.

### The four valid user intents for `dim=shape` where `shape=(3, 3)`

Given a launch shape of `(3, 3)`, there are four distinct thread-ID patterns a user might want:

| Intent (thread IDs produced) | Correct usage |
|---|---|
| `(0,0), (0,1), (0,2), (1,0), …, (2,2)` — full 2-D work | `i, j = wp.tid()` with `dim=shape` |
| `0, 0, 0, 1, 1, 1, 2, 2, 2` — first-dim only, with repetition | `i, _ = wp.tid()` with `dim=shape` |
| `0, 1, 2, …, 8` — flat linear across all 9 threads | `i = wp.tid()` with `dim=math.prod(shape)` |
| `0, 1, 2` — first-dim only, no repetition | `i = wp.tid()` with `dim=shape[0]` |

All four are expressible cleanly without the implicit fold. The ambiguous form `i = wp.tid()` + `dim=(3, 3)` currently produces the third pattern silently and is the least likely intended one if the old behavior worked correctly for them but was unintentionally slower.

### Options to address

1. **Warn, don't error.** Keep the current silent fold, but emit a `UserWarning` at launch time whenever `ndim(dim) > kernel_dim`. Zero-cost migration for code that genuinely wants the fold; users who hit it by accident see the warning and fix it. Weakest remedy — warnings are easy to miss in noisy log output, and in the Newton case the first visible symptom was a heap-corruption crash with no obvious connection to the kernel at fault.

2. **Error, with an explicit opt-in.** Raise `ValueError` by default when `ndim(dim) > kernel_dim`, and add a way to say "I actually want the fold" — either a `flatten=True` kwarg on `wp.launch()`, a helper `wp.fold_dim(shape, kernel_dim)`, or a `wp.config.allow_dim_flatten` flag. Preserves the documented behavior for users who want it; forces a deliberate choice.

3. **Error, no opt-in.** Raise `ValueError` unconditionally. Users who want the flatten behavior rewrite the call with an explicit scalar or manually-folded shape (`dim=math.prod(shape)` for `kernel_dim=1`, `dim=(shape[0], shape[1]*shape[2])` for partial folds, etc.). No new API surface; no silent footgun. Users genuinely relying on the fold pay a one-time migration cost.

### Recommendation

Option 3 — hard error, no opt-in. The Newton bug illustrates exactly why the silent fold is dangerous: a caller who thought they were indexing the first dimension of a 3-D array got the linear-index behavior instead, exceeding the shape dimensions, and the only feedback channel was a delayed memory-corruption crash. That caller did not want the fold; they wanted the first-dim-only pattern (row 4 of the table above) and had always wanted it. They accidentally got the old unravel behavior (row 2) for free, and when the semantics flipped to the fold behavior (row 3), there was no signal that anything had changed until production crashes started rolling in.

For genuine fold use cases — rare, and trivially rewritable as a scalar or partially-folded shape — the user is better served by spelling it out at the call site. The fold ambiguity for `kernel_dim ≥ 2` (which dims collapse into which) is itself a small footgun that benefits from being made explicit.

Option 2's opt-in kwarg addresses the same problem but adds API bureaucracy for a corner case. If the friction from option 3 turns out to be larger than expected, adding an opt-in later is a strictly additive change — it can be bolted on without further breakage.

shi-eric:
## Description

Replace the silent fold/pad in `_normalize_launch_dim` with a hard `ValueError` on rank mismatch.

Introduce `_prepare_launch_dim(dim, kernel, *, tiled=False)` which canonicalizes `dim`, bypasses validation for kernels with no `wp.tid()` call (flattening any shape to total thread count via `math.prod`), and raises `ValueError` with a concrete migration hint listing 2–4 valid fix options computed from the specific call.

Wire into `wp.launch()`, `Launch.set_dim()`, `wp.launch_tiled()` via `_construct_tiled_bounds`, and the experimental JAX FFI paths (`custom_call.py`, `ffi.py`). Remove `_normalize_launch_dim` entirely. `Launch.set_dim()` preserves tiled semantics on recorded tiled launches.

The `launch_tiled` rank-mismatch error type changes from `RuntimeError` to `ValueError` for consistency with the regular launch path; kernels that do not call `wp.tid()` continue to accept any `dim`.

Motivation: the silent fold introduced by the `launch_bounds_t<N>` templating (GH-1270, unreleased) caused out-of-bounds indexing and heap corruption in downstream kernels — the trigger took a few hours of debugging to isolate in Newton ([newton-physics/newton#2546](https://github.com/newton-physics/newton/pull/2546)) because the overflow accumulates until an allocator validator notices. Since GH-1270 has not shipped in a release, the hard-error replaces the silent fold as the shipping behavior for 1.13 — the CHANGELOG entry is consolidated under GH-1270.

closes #1389

## Checklist

## Test plan

- `uv run --extra dev -m warp.tests -k test_template_launch_bounds` — 37 tests pass (includes 20 unit tests for new helpers + 6 integration tests for the three call sites + 11 prior launch-bounds tests).
- `uv run --extra dev -m warp.tests -k test_launch` — 19 tests pass (regression guard on existing launch functionality).
- Representative error output for ``tid = wp.tid()`` launched with ``dim=(3, 3)``:

```
ValueError: Launch dim (3, 3) has rank 2 but kernel 'foo' unpacks wp.tid() into 1 variable (kernel_dim=1).
Pick the intended behavior:
  - flat linear index over 9 threads: launch with dim=9
  - first-dim index (0..2, no repetition): launch with dim=3
  - first-dim index with repetition: unpack as \`i, _ = wp.tid()\`
  - per-dim indexing: unpack as \`i, j = wp.tid()\`
```

## Bug fix

Kernel code that worked under the pre-1.13 unravel semantics:

\`\`\`python
@wp.kernel
def repeat_first_dim(arr: wp.array3d(dtype=float)):
    # Pre-1.13: launched with dim=arr.shape (3-tuple), tid got first-dim
    # index 0,0,0,1,1,1,... — idempotent writes, coincidentally correct.
    tid = wp.tid()
    arr[tid, 0, 0] = float(tid)

wp.launch(repeat_first_dim, dim=arr.shape, inputs=[arr])
\`\`\`

Under unreleased GH-1270 this silently folds to a flat linear `tid`, which then indexes past `arr.shape[0]` and corrupts the heap. This PR raises `ValueError` at launch time with the migration options inline; the user picks \`tid = wp.tid()\` + \`dim=arr.shape[0]\` or \`i, _, _ = wp.tid()\` + \`dim=arr.shape\`.

## Summary by CodeRabbit

* **Breaking Changes**
  * Launch-dimension rank is now strictly validated; mismatches raise ValueError (including tiled launches). Kernels that read thread indices must use a matching rank; kernels that never read thread indices accept any shape by flattening to a total-thread count. launch_tiled rank-mismatch now raises ValueError.

* **Documentation**
  * Changelog updated with migration guidance and clarified validation messages.

* **Tests**
  * Expanded coverage for rank validation, error text, migration hints, and tiled vs non-tiled behaviors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,13 +109,14 @@
   ([GH-1297](https://github.com/NVIDIA/warp/issues/1297)).
 - Template `launch_bounds_t` on dimensionality (`launch_bounds_t<N>`) so that `launch_coord()` and `wp.tid()`
   use `if constexpr` instead of runtime dimension checks, eliminating dead branches and reducing register
-  pressure for lower-dimensional launches
-  ([GH-1270](https://github.com/NVIDIA/warp/issues/1270)).
-  **Breaking:** Kernels that use fewer ``wp.tid()`` return values than launch dimensions now flatten
-  excess dimensions instead of unraveling them. For example, a kernel using ``i = wp.tid()`` launched
-  with ``dim=(3, 3)`` now yields thread IDs ``0, 1, …, 8`` instead of the previous ``0, 0, 0, 1, 1, 1,
-  2, 2, 2``. To preserve the old behavior, unpack all dimensions and discard the ones you don't need
-  (e.g. ``i, _ = wp.tid()`` to recover the previous first-dimension indexing).
+  pressure for lower-dimensional launches.
+  **Breaking:** `wp.launch()`, `Launch.set_dim()`, `wp.launch_tiled()`, and the experimental JAX FFI paths
+  now raise `ValueError` when the rank of `dim` does not match the kernel's `wp.tid()` unpack arity
+  (e.g. `i = wp.tid()` launched with `dim=(3, 3)`); the error message lists concrete migration options.
+  To recover the previous first-dimension indexing, unpack all dimensions and discard extras
+  (e.g. `i, _ = wp.tid()`). Kernels that do not call `wp.tid()` still accept any `dim`, flattened to
+  total thread count. `launch_tiled`'s rank-mismatch error changed from `RuntimeError` to `ValueError`
+  for consistency ([GH-1270](https://github.com/NVIDIA/warp/issues/1270)).
 - Allow `wp.Volume.load_from_numpy()` and `wp.Volume.allocate()` to accept a 3-element sequence for `voxel_size`,
   enabling anisotropic voxel spacing ([GH-1193](https://github.com/NVIDIA/warp/issues/1193)).
 - Include the Warp version in kernel cache paths when a custom cache path is set,
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -17,6 +17,7 @@ import inspect
 import io
 import itertools
 import json
+import math
 import operator
 import os
 import platform
@@ -7884,6 +7885,7 @@ class Launch:
         max_blocks: int = 0,
         block_dim: int = 256,
         adjoint: bool = False,
+        tiled: bool = False,
     ):
         # retain the module executable so it doesn't get unloaded
         self.module_exec = kernel.module.load(device, block_dim)
@@ -7951,14 +7953,28 @@ class Launch:
         self.adjoint: bool = adjoint
         """Whether to run the adjoint kernel instead of the forward kernel."""
 
+        self.tiled: bool = tiled
+        """True if this launch was created via :func:`wp.launch_tiled`. Used by
+        :meth:`set_dim` to route replays through the tiled dim-validation path.
+        Not the same as ``bounds.tiled``; see ``_construct_tiled_bounds`` for
+        when each is set.
+        """
+
     def set_dim(self, dim: int | list[int] | tuple[int, ...]):
         """Set the launch dimensions.
+
+        For recorded tiled launches (``self.tiled == True``), reconstructs
+        bounds via the same tiled path used at record time so the user-rank
+        semantics (block axis implicit or unpacked) are preserved.
 
         Args:
             dim: The dimensions of the launch.
         """
-        normalized = _normalize_launch_dim(dim, self.kernel.adj.kernel_dim)
-        self.bounds = launch_bounds_t(normalized)
+        if self.tiled:
+            self.bounds = _construct_tiled_bounds(dim, self.block_dim, self.kernel)
+        else:
+            normalized = _prepare_launch_dim(dim, self.kernel)
+            self.bounds = launch_bounds_t(normalized)
 
         # launch bounds always at index 0
         self.params[0] = self.bounds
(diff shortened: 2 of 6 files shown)


c0d1f1ed:
Served as a foundation for https://github.com/NVIDIA/warp/commit/e9e7cc6a4c1909023c027d5098827422764e3eca (MR !2256), which gates the optimization and error behind a `wp.config.optimize_tid` flag (default `False`). Ultimately the optimization was reverted to address the Newton issue and avoid others running into similar crashes. It remains to be seen how to re-introduce the optimization, but this preliminary PR can be closed for now. Thanks Eric!

PR #1397, opened 2026-04-27
Add cooperative GPU scalar fallbacks for tile_cholesky, tile_*_solve

daedalus5:
## Description

Add cooperative GPU scalar fallback paths for `tile_cholesky`,
`tile_cholesky_solve`, `tile_lower_solve`, `tile_upper_solve` (and their
inplace variants), and the `tile_cholesky` adjoint, so they parallelise
across `WP_TILE_BLOCK_DIM` threads when libmathdx is unavailable or
disabled. Closes a gap relative to `tile_matmul` and `tile_fft`, whose
non-libmathdx paths are already cooperative.

Also splits the relevant ops out of `warp/native/tile.h` into per-category
headers `tile_matmul.h`, `tile_solve.h`, `tile_cholesky.h`, mirroring
`tile_reduce.h`/`tile_radix_sort.h`. `tile.h` shrinks by ~1100 lines.

Adds two new module options `enable_mathdx_cholesky` and
`enable_mathdx_trsm` (parity with `enable_mathdx_gemm`) so callers can
route specific kernels through the scalar path on a libmathdx-enabled
build.

Algorithmic choices captured as inline tradeoff comments:
- Vector-RHS triangular solve uses a thread-0 gate + sync (the inner
  loop is sequential and the inner dot product alone doesn't pay for
  reduction overhead at typical n).
- Cholesky factorisation diagonal element is computed redundantly on
  all threads (race-free read-only access; only thread 0 writes the
  result), avoiding a shared-mem broadcast cell.
- Cooperative Cholesky adjoint mirrors the libmathdx LTO path's
  six-phase structure (gemm, symmetrize, two trsms with a transpose,
  accumulate) verbatim, with scalar inner kernels in place of LTO calls.

Performance notes in each new header are grounded in measured benchmarks
(L40, sm_89, fp64, block_dim=32). Headline: cuSolverDx wins at every
size measured; the cooperative scalar path is primarily a correctness
fallback for builds without libmathdx and a way to skip slow LTO
compilation during development. The matmul-style "scalar wins at small
tiles" crossover does not appear here because of the sequential
dependencies in cholesky and trsm.

Out of scope (still `assert(false)` stubs, tracked separately):
`adj_tile_cholesky_inplace`, `adj_tile_cholesky_solve` /
`_inplace`, `adj_tile_lower_solve` / `_inplace`, `adj_tile_upper_solve`
/ `_inplace`. These need actual gradient derivations independent of
mathdx availability.

## Checklist

## Test plan

- New `warp/tests/tile/test_tile_solve_no_mathdx.py` (24 tests) and
  `warp/tests/tile/test_tile_cholesky_no_mathdx.py` (16 tests),
  registered in `warp/tests/unittest_suites.py:default_suite`. Set
  `enable_mathdx_<op>=False` at module scope so the same test file runs
  the cooperative path on cuda and the CPU path on cpu without rebuilding
  the native library.
- Coverage includes block_dim==1 GPU smoke tests (single-thread-block
  codegen path) and a larger-N (n=32) cholesky-lower adjoint test that
  exercises the `__shared__ T W1[n*n], W2[n*n]` scratch budget.
- Existing `test_tile_cholesky.py` adjoint tests now cover the
  cooperative scalar adjoint on CPU (where `block_dim==1` collapses the
  cooperative loops to sequential).
- Verified locally on L40 / sm_89: full tile suite (`-k TestTile`) runs
  546/546 OK.

## New feature / enhancement

```python
import warp as wp

# Route tile_cholesky and its adjoint through the cooperative scalar path
# instead of cuSolverDx, e.g. to skip the slow LTO compile during development.
wp.set_module_options({"enable_mathdx_cholesky": False})

@wp.kernel
def cholesky_kernel(gA: wp.array2d(dtype=wp.float64), gL: wp.array2d(dtype=wp.float64)):
    A = wp.tile_load(gA, shape=(8, 8))
    L = wp.tile_cholesky(A)
    wp.tile_store(gL, L)
```

## Summary by CodeRabbit

* **New Features**
  * Added configuration options (`enable_mathdx_cholesky`, `enable_mathdx_trsm`) to control GPU execution paths for Cholesky and triangular solve operations, enabling fallback to shared-memory implementations when optimized paths are disabled.

* **Tests**
  * Added comprehensive test suites for Cholesky and triangular solve operations with fallback implementations enabled.


daedalus5:
Closing — pushed for visibility on the contributor fork only, not ready for upstream review yet.

Issue #1400, opened 2026-04-28
APIC C++ examples (02_apic_visualization, 03_apic_visualization_cpu) do not build on Linux


shi-eric:
### Bug Description

The two C++ APIC examples added in `d11116fca` (GH-1349), `warp/examples/cpp/02_apic_visualization` and `warp/examples/cpp/03_apic_visualization_cpu`, do not build on Linux as shipped.

#### Issue 2a: library naming

On Linux, `build_lib.py:215-222` produces `warp.so` (no `lib` prefix). Both examples' build files look for the conventional `lib`-prefixed name:

- `02/CMakeLists.txt:88` and `03/CMakeLists.txt:68`: `find_library(WARP_LIB NAMES warp warp.dll …)` searches for `libwarp.so` on Linux.
- `02/Makefile:58` and `03/Makefile:54`: `-lwarp` searches for `libwarp.so`.

Repro:

```
$ cd warp/examples/cpp/03_apic_visualization_cpu
$ cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Release
CMake Error at CMakeLists.txt:75 (message):
  Could not find Warp library (warp.dll / libwarp.so)
```

#### Issue 2b: hardcoded mismatched `dlopen` name in `03/main.cpp`

```cpp
// warp/examples/cpp/03_apic_visualization_cpu/main.cpp:264
const char* lib_name = "libwarp-clang.so";
```

The actual file shipped in `warp/bin/` on Linux is `warp-clang.so` (no `lib` prefix). Even after fixing 2a, the program would fail at runtime with `Failed to load libwarp-clang.so`.

#### Independent of #1399

These bugs are independent of the fastcall Python-symbol issue tracked in #1399. Even with a Python-free `warp.so`, the examples would still fail to build on Linux.

### System Information

Linux x86_64, Warp `main` at `dac22ded4`, CMake 3.20+, ninja, g++ 13. Affects both branches that contain `d11116fca`.

shi-eric:
## Update from implementation: an extra naming layer, plus a blocker on #1399

While preparing a fix on a feature branch, two things surfaced that weren't in the original issue text.

### 1. CMake auto-converts absolute paths to `-L … -l<name>`

Extending `find_library(NAMES warp warp.so warp.dll)` makes configure-time lookup resolve `WARP_LIB` to `/path/to/warp.so` correctly. But the link step still fails:

```
[ 89%] Linking CXX executable 03_apic_visualization_cpu
[...]

Issue #1399, opened 2026-04-28
warp.so/warp.dll has unresolved Python C API symbols, blocks loading from non-Python C++ apps


shi-eric:
### Bug Description

Since `4bf1ae879` ("Add METH_FASTCALL native bindings for half-float conversion", GH-1339), `fastcall.cpp` is statically embedded in `warp.so`/`warp.dll`. As a result, the resulting library has 8 unresolved Python C API symbols:

```
PyFloat_FromDouble, PyFloat_AsDouble,
PyLong_FromUnsignedLong, PyLong_AsUnsignedLongMask,
PyErr_Occurred, PyErr_SetString, PyExc_TypeError,
PyModule_Create2
```

Inside a Python process the interpreter resolves them, so `import warp` works fine. Outside a Python process the load fails. Reproduction on Linux:

```cpp
// dlopen_test.cpp
#include <dlfcn.h>
#include <cstdio>
int main() {
    void* h = dlopen("warp/bin/warp.so", RTLD_LAZY);
    if (!h) { fprintf(stderr, "%s\n", dlerror()); return 1; }
    return 0;
}
```

```
$ g++ dlopen_test.cpp -ldl -o t && ./t
warp/bin/warp.so: undefined symbol: PyExc_TypeError
```

`PyExc_TypeError` is a data symbol, so even `RTLD_LAZY` resolves it eagerly. There is no escape via lazy binding.

On Windows, `build_dll.py:694` unconditionally adds `python3.lib` to the link line, so `warp.dll` should carry an import-table dependency on `python3.dll` and fail to load via `LoadLibrary` on a system without it. Inferred from the build script, not yet verified on a Windows build.

#### Why this matters

The new APIC CPU-replay feature added in `d11116fca` (GH-1349) is positioned as "capture a `.wrp` file in Python, deploy it to a Python-less C++ application on an embedded / edge target, replay there." Today, on a Linux target without a Python interpreter, the C++ app cannot load `warp.so` at all, even though `wp_apic_cpu_replay_graph` itself is pure C ABI.

### System Information

Reproduced on Warp `main` at `dac22ded4`, Linux x86_64. Affects `1.13.0rc1` and `1.13.0rc2`. No released wheel (<= `1.12.1`) is affected.

c0d1f1ed:
Fixed by 666acf2a96b6c63cec0b79f93462acdc1f7a6a27

PR #1404, opened 2026-04-29
Fix APIC C++ examples on Linux (GH-1400)

shi-eric:
## Summary

Fixes the C++ APIC examples (`02_apic_visualization`, `03_apic_visualization_cpu`) which currently fail to build on Linux on `release-1.13`. Targets `release-1.13` first because that branch already has the `fastcall.cpp` revert from #1399 — the `main`-branch fix is blocked on that landing there. See [the GH-1400 update comment](https://github.com/NVIDIA/warp/issues/1400#issuecomment-4340025975) for context.

Two commits:

1. **`Fix APIC C++ example library lookup on Linux`** — five-site naming alignment with `build_lib.py`'s prefix-less `warp.so` / `warp-clang.so` Linux artifacts:
   - CMake `find_library(NAMES warp warp.so warp.dll …)` so the bare `warp.so` is matched.
   - Makefile linux LDFLAGS use `-l:warp.so` (GNU ld literal-name extension).
   - `03_apic_visualization_cpu/main.cpp` `dlopen("warp-clang.so")` (was `libwarp-clang.so`).
   - Reorder warp-library lookup above the `find_package(OpenGL QUIET)` gate so configure fails loudly even on hosts without GL.
   - Widened `FATAL_ERROR` enumerates all three platform variants.

2. **`Add CTest smoke coverage for APIC C++ examples`** — make the existing `linux-x86_64 cpp examples test` job actually exercise these examples:
   - `--smoke` argv branch in both `main.cu` / `main.cpp` runs the meaningful warp paths (load graph → load CPU modules → query params → replay 10×) and exits before any `glfwInit()`.
   - `add_test(NAME apic_visualization_smoke / apic_visualization_cpu_smoke …)` with the `cpp_example` label.
   - `test_examples.sh` extended to run `capture_wave.py` alongside the existing `compile_kernel.py` convention so `.wrp` inputs are produced before the C++ build step.
   - Wraps the resolved warp library in an `UNKNOWN IMPORTED` CMake target so `target_link_libraries` passes the absolute path verbatim — otherwise CMake auto-converts `/path/to/warp.so` to `-L/path/to -lwarp` and the link fails the same way as the original Makefile bug.

The `warp-cpp-test-env` image already has the OpenGL/X11 build deps (committed via `88781e9d9` on `main`; the rebuilt `12.9.1-ubuntu24.04`-tagged image was pushed to the registry, and `release-1.13` references the same tag). No CI image change is needed in this PR.

Out of scope: #1399 (Python C ABI symbols leaking into `warp.so`). That issue is independent at the cause level but is a hard prerequisite for the same fix on `main`.

## Test plan

- [x] Local: `uv run build_lib.py --quick && cd warp/examples/cpp && bash test_examples.sh` — all 4 CTest entries pass:
  - `cubin_launch_test` (existing)
  - `source_include_test` (existing)
  - `apic_visualization_smoke` (new) — `smoke OK (10 graph launches)`
  - `apic_visualization_cpu_smoke` (new) — `smoke OK (10 replay iterations)`
- [ ] CI: `linux-x86_64 cpp examples test` job picks up the two new tests.

### DIFF
--- a/warp/examples/cpp/02_apic_visualization/CMakeLists.txt
+++ b/warp/examples/cpp/02_apic_visualization/CMakeLists.txt
@@ -22,6 +22,42 @@ enable_testing()
 # Find CUDA
 find_package(CUDAToolkit REQUIRED)
 
+# Find Warp native directory
+# Default: assume repository checkout (examples are in warp/examples/cpp/)
+if(NOT DEFINED WARP_NATIVE_DIR)
+    set(WARP_NATIVE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../../native")
+endif()
+
+# Resolve to absolute path and verify
+get_filename_component(WARP_INCLUDE "${WARP_NATIVE_DIR}" ABSOLUTE)
+if(NOT EXISTS "${WARP_INCLUDE}/builtin.h")
+    message(FATAL_ERROR
+        "Could not find Warp native directory at: ${WARP_INCLUDE}\n"
+        "Set WARP_NATIVE_DIR to override (e.g., cmake -DWARP_NATIVE_DIR=/path/to/warp/native)")
+endif()
+
+# Find Warp library
+# Default: look in warp/bin directory (repository structure)
+if(NOT DEFINED WARP_BIN_DIR)
+    set(WARP_BIN_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../../bin")
+endif()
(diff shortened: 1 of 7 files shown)


shi-eric:
Closing — NVIDIA-side merges go through internal GitLab. Local branch ershi/fix-apic-examples-1.13 retained for the equivalent MR.

PR #1411, opened 2026-04-30
fix: kernel return validation

pumpkin-bit:
## Description

Kernels are architecturally prohibited from returning values. Previously, the code generation system allowed return type annotations ``(e.g., '-> float')`` for kernels, but the function body did not explicitly return a value. This PR adds a check for rejecting return type hints for kernels to prevent misleading function signatures and avoid user confusion.

This also addresses an existing ``TODO`` and includes a corresponding test.

## Checklist

## Test plan

Verified by running the codegen test suite: `test_codegen.py`

The updated `test_error_kernel_return_value` explicitly checks and passes the `WarpCodegenTypeError` assertion when a return type hint is provided without a return value.

## Bug fix

```python
import warp as wp

wp.init()

# Prior to this PR, this compiled silently without raising an error
@wp.kernel
def f(x: float) -> float:
    return

# Should raise a WarpCodegenTypeError
wp.launch(f, dim=1, inputs=[3.0], device="cpu")


## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation to reject kernels that declare or imply non-None return values, preventing erroneous launches and related runtime errors.

* **Tests**
  * Re-enabled and added tests that assert kernels with bare returns are rejected while kernels annotated to return None remain accepted.


shi-eric:
Thanks for the contribution. This addresses the existing TODO in `test_error_kernel_return_value`: kernels already raise when they actually return a value, but a kernel with a non-`None` return annotation and only a bare `return` was still accepted, which leaves a misleading signature that Warp cannot honor.

Could you please make a few cleanup changes before this can move forward?

- Add a `CHANGELOG.md` entry under the `Unreleased` section, since this changes user-visible codegen behavior from silently accepted to reported at compile time.
- Add a test confirming that `-> None` remains accepted for kernels.
- Clean up the commit history by squashing the duplicate/fixup commits and merge commit into one signed-off commit. We need this before merging because we preserve contributor authorship when merging through our internal GitLab workflow, so we cannot squash these commits on your behalf.

Also, for future non-trivial codegen behavior changes, please open an issue or discussion first so we can agree on the desired behavior before implementation.

@c0d1f1ed @nvlukasz, could you weigh in on whether a non-`None` kernel return annotation should be a hard error, as implemented here, or a warning?

pumpkin-bit:
Alright, I created a changelog entry and wrote a test regarding the commits, I'm thinking of doing a squash after all the necessary maintainers have shared their feedback, so that all changes are agreed upon if needed also, the single squash commit will be signed as I understand it, intermediate commits don't need signatures since the squash will bring them into one? And next time for similar fixes I will create an issue or a discussion. Thanks.

shi-eric:
Thanks for working on this. The validation for non-`None` kernel return annotations, including the bare-return case addressed here, has since landed on `main` in 2abd428dbee34412b6b8f98a717755d112ecf97a, together with broader checks for value-return statements and `@wp.overload` stubs.

Since this PR is now covered by that commit, I am closing it as superseded. Thanks again for the contribution!

PR #1438, opened 2026-05-07
Compile bvh.cpp and scan.cpp as individual units

hong-nvidia:
## Description

We noticed the non-idiomatic pattern of pulling `.cpp` file into other `.cpp` files. It obfuscates include-order dependencies and makes each file harder to compile or reuse on its own.

Therefore, we've isolated the two files in question into their own compilation unit, and removed the problematic `#include`s.

## Test plan

Covered by existing build tests.

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Build system updated to include additional core library components for native builds.
  * Internal compilation structure reorganized to streamline the build and reduce redundant translation-unit handling.
  * Cleaned up stray artifacts and redundant includes to improve repository hygiene and build determinism.

### DIFF
--- a/build_lib.py
+++ b/build_lib.py
@@ -488,6 +488,8 @@ def main(argv: list[str] | None = None) -> int:
         # build warp.dll
         cpp_sources = [
             "native/warp.cpp",
+            "native/bvh.cpp",
+            "native/scan.cpp",
             "native/apic.cpp",
             "native/alloc_tracker.cpp",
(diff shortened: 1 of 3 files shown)


shi-eric:
Since you're an NV developer, please open this as a merge request on GitLab. It's okay to use GitHub for agentic development though, but all merges currently must be done on GitLab.

PR #1442, opened 2026-05-08
debug: H100 tile hang reproduction (DO NOT MERGE)

shi-eric:
**This PR is for debugging only — do not merge.**

Minimal workflow to characterize the H100 tile CPU hang observed in CI timeouts.

Three-stage test on H100 runner:
1. **Tile tests serial (j=1)** — baseline. If this hangs → deterministic bug.
2. **Tile tests parallel (j=4)** — tile-only contention. If this hangs but #1 passes → contention-triggered.
3. **Full test suite parallel** — reproduce exact CI conditions. If only this hangs → cross-module contention.

Related: All 6 recent H100 timeout failures end in tile CPU tests that start but never complete.

Context: https://gist.github.com/shi-eric/cd31760f5a008dcce68381a8d8c17bad

## Summary by CodeRabbit

* **Chores**
  * Streamlined CI flows, restricted triggers, consolidated redundant jobs, and standardized artifact retention and concurrency behavior to simplify pipelines and shorten run time.
  * Removed legacy static-analysis job and several outdated workflow steps.

* **Tests**
  * Reworked GPU testing into containerized H100 tile-stack runs with two variants (stack-true and stack-false) executing the full test suite and collecting system info.
  * Cleaned up containerized doctest setup steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,421 +1,71 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-License-Identifier: Apache-2.0
-
 name: CI
 
 on:
   workflow_call:
   workflow_dispatch:
   push:
-    tags:
-      - v*
     branches:
-      - main
-      - release-*
-  schedule: # Schedule workflow only runs on the default branch
-    - cron: "45 9 * * *"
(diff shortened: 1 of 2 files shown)


shi-eric:
Closing — unable to reproduce the H100 tile hang after 5 attempts. The issue appears to require specific runner contention conditions that can't be forced. Will investigate further if it recurs organically.

PR #1446, opened 2026-05-08
Add JAX FFI Host support

loney7:
## Description

This PR adds support for running JAX FFI callbacks on the CPU (Host) in addition to CUDA.

**Changes:**
- Registered FFI targets for both `"CUDA"` and `"Host"` platforms in `register_ffi_callback`.
- Handled the `"Host"` platform case in `ffi_callback` by using the CPU device and bypassing CUDA-specific features like streams and graphs.
- Updated `FfiCallable` to reconstruct arguments and execute the function on the CPU when running on the Host platform.

## Checklist

## Test plan

You can verify these changes by running the JAX interop tests which include FFI tests. Ensure they pass on both CPU and GPU (if available).

```bash
uv run warp/tests/interop/test_jax.py


## Summary by CodeRabbit

* **New Features**
  * Separate CUDA and CPU execution paths for FFI callbacks with device-aware execution scoping.
  * CUDA graph compatibility enabled only for CUDA execution.
  * CPU/host path can take a direct Python execution route for faster host calls.
  * Separate CUDA and Host callback registrations for JAX interop.

* **Tests**
  * Added CPU/host FFI tests covering kernels and callables (add, sincos, in/out args, scale).

### DIFF
--- a/warp/_src/jax/ffi.py
+++ b/warp/_src/jax/ffi.py
@@ -235,10 +235,22 @@ class FfiKernel:
         # register the callback
         jax = _get_jax()
         FFI_CCALLFUNC = ctypes.CFUNCTYPE(ctypes.c_void_p, ctypes.POINTER(XLA_FFI_CallFrame))
-        self.callback_func = FFI_CCALLFUNC(self.ffi_callback)
-        ffi_ccall_address = ctypes.cast(self.callback_func, ctypes.c_void_p)
-        ffi_capsule = jax.ffi.pycapsule(ffi_ccall_address.value)
-        jax.ffi.register_ffi_target(self.name, ffi_capsule, platform="CUDA")
+
+        def make_callback(platform):
+            def callback(call_frame):
+                return self.ffi_callback(call_frame, platform)
+
+            return callback
+
+        self.callback_func_cuda = FFI_CCALLFUNC(make_callback("CUDA"))
+        ffi_capsule_cuda = jax.ffi.pycapsule(ctypes.cast(self.callback_func_cuda, ctypes.c_void_p).value)
+        jax.ffi.register_ffi_target(self.name, ffi_capsule_cuda, platform="CUDA")
+        try:
+            self.callback_func_host = FFI_CCALLFUNC(make_callback("Host"))
+            ffi_capsule_host = jax.ffi.pycapsule(ctypes.cast(self.callback_func_host, ctypes.c_void_p).value)
+            jax.ffi.register_ffi_target(self.name, ffi_capsule_host, platform="Host")
+        except AttributeError:
+            pass
 
     def __call__(self, *args, output_dims=None, launch_dims=None, vmap_method=None):
         jax = _get_jax()
@@ -332,7 +344,8 @@ class FfiKernel:
         # preload on the specified devices
         if self.module_preload_mode == JaxModulePreloadMode.CURRENT_DEVICE:
             device = wp.device_from_jax(get_jax_device())
-            self.kernel.module.load(device)
+            block_dim = 256 if device.is_cuda else 1
+            self.kernel.module.load(device, block_dim)
         elif self.module_preload_mode == JaxModulePreloadMode.ALL_DEVICES:
             for d in jax.local_devices():
                 try:
@@ -340,9 +353,10 @@ class FfiKernel:
                 except Exception:
                     # ignore unsupported devices like TPUs
                     pass
-                # we only support CUDA devices for now
-                if dev.is_cuda:
-                    self.kernel.module.load(dev)
+                # we only support CUDA and CPU devices for now
+                if dev.is_cuda or dev.is_cpu:
+                    block_dim = 256 if dev.is_cuda else 1
+                    self.kernel.module.load(dev, block_dim)
 
         # save launch data to be retrieved by callback
         launch_id = self.launch_id
@@ -353,7 +367,7 @@ class FfiKernel:
 
         return call(*args, launch_id=launch_id)
 
-    def ffi_callback(self, call_frame):
+    def ffi_callback(self, call_frame, platform):
         try:
             # On the first call, XLA runtime will query the API version and traits
             # metadata using the |extension| field. Let us respond to that query
@@ -365,10 +379,11 @@ class FfiKernel:
                     metadata_ext = ctypes.cast(extension, ctypes.POINTER(XLA_FFI_Metadata_Extension))
                     metadata_ext.contents.metadata.contents.api_version.major_version = 0
                     metadata_ext.contents.metadata.contents.api_version.minor_version = 1
-                    # Turn on CUDA graphs for this handler.
-                    metadata_ext.contents.metadata.contents.traits = (
-                        XLA_FFI_Handler_TraitsBits.COMMAND_BUFFER_COMPATIBLE
-                    )
+                    if platform == "CUDA":
+                        # Turn on CUDA graphs for this handler.
+                        metadata_ext.contents.metadata.contents.traits = (
+                            XLA_FFI_Handler_TraitsBits.COMMAND_BUFFER_COMPATIBLE
+                        )
                     return None
 
             # Lock is required to prevent race conditions when callback is invoked
@@ -448,11 +463,20 @@ class FfiKernel:
                 kernel_params[0] = ctypes.addressof(launch_bounds)
 
                 # get device and stream
-                device = wp.get_cuda_device(get_device_ordinal_from_callframe(call_frame.contents))
-                stream = get_stream_from_callframe(call_frame.contents)
+                if platform == "CUDA":
+                    device = wp.get_cuda_device(get_device_ordinal_from_callframe(call_frame.contents))
+                    stream = get_stream_from_callframe(call_frame.contents)
+                else:
+                    device = wp.get_device("cpu")
+                    stream = None
 
-                # get kernel hooks
-                hooks = self.kernel.module.get_kernel_hooks(self.kernel, device)
+                # get kernel hooks (ensure module is loaded on the target device,
+                # e.g. Host callbacks may run on CPU even when preloaded on CUDA).
+                # CPU tile execution is single-lane, so use block_dim=1 to compile
+                # WP_TILE_BLOCK_DIM correctly for explicitly shaped tile operations.
+                block_dim = 256 if device.is_cuda else 1
+                module_exec = self.kernel.module.load(device, block_dim)
+                hooks = module_exec.get_kernel_hooks(self.kernel)
                 assert hooks.forward, "Failed to find kernel entry point"
 
                 # reject non-cluster-aligned grids with a clear Python error instead
@@ -460,20 +484,32 @@ class FfiKernel:
                 _validate_cluster_launch(hooks.cluster_dim, launch_bounds.size, 256, 0)
 
                 # launch the kernel (cluster_dim is cached on the hooks at load time)
-                if wp._src.context.runtime.core.wp_cuda_launch_kernel(
-                    device.context,
-                    hooks.forward,
-                    launch_bounds.size,
-                    0,
-                    256,
-                    int(self.kernel.grid_stride),
-                    hooks.cluster_dim,
-                    hooks.forward_smem_bytes,
-                    kernel_params,
-                    stream,
-                    None,  # apic_info
-                ):
-                    _raise_cuda_launch_error(self.kernel, device)
+                if platform == "CUDA":
+                    if wp._src.context.runtime.core.wp_cuda_launch_kernel(
+                        device.context,
+                        hooks.forward,
+                        launch_bounds.size,
+                        0,
+                        256,
+                        int(self.kernel.grid_stride),
+                        hooks.cluster_dim,
+                        hooks.forward_smem_bytes,
+                        kernel_params,
+                        stream,
+                        None,  # apic_info
+                    ):
+                        _raise_cuda_launch_error(self.kernel, device)
+                else:
+                    args_struct, _ = wp._src.context._build_cpu_args_structs(
+                        self.kernel, hooks, [launch_bounds, *arg_refs], False
+                    )
+                    wp._src.context.runtime.core.wp_cpu_launch_kernel(
+                        ctypes.cast(hooks.forward, ctypes.c_void_p),
+                        ctypes.byref(launch_bounds),
+                        ctypes.byref(args_struct) if args_struct is not None else None,
+                        None,
+                        None,
+                    )
 
         except Exception as e:
             print(traceback.format_exc())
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by loney7** [comment]
pre-commit.ci autofix

**Comment by loney7** [comment]
pre-commit.ci autofix

File: warp/tests/interop/test_jax.py [hunk 1]
```diff
@@ -2256,6 +2419,29 @@ class TestJax(unittest.TestCase):
         # ffi callback tests
         add_function_test(TestJax, "test_ffi_callback", test_ffi_callback, devices=jax_compatible_cuda_devices)
 
+        # ffi Host (CPU) tests
+        add_function_test(TestJax, "test_ffi_jax_kernel_host_add", test_ffi_jax_kernel_host_add, devices=None)
```
shi-eric: These Host tests only use `jax.devices("cpu")`, but they are registered inside the CUDA JAX availability gate. Please move them under CPU JAX availability so CPU-only CI covers the Host backend.

warp/_src/jax/ffi.py
```diff
@@ -427,25 +434,37 @@ def ffi_callback(self, call_frame):
...
+                        launch_bounds.size,
+                        0,
+                        256,
+                        hooks.forward_smem_bytes,
+                        kernel_params,
+                        stream,
+                        None,  # apic_info
+                    )
+                else:
+                    wp._src.context.runtime.core.wp_cpu_launch_kernel(
```
shi-eric, line 506: This Host path is using the CUDA launch ABI for `wp_cpu_launch_kernel`. The CPU binding expects `(func, bounds, args, adj_args, apic_info)`, with kernel args packed into the CPU args struct. As written, Host `jax_kernel` raises `TypeError: this function takes at least 5 arguments (4 given)` instead of launching.

shi-eric:
Please add a `CHANGELOG.md` entry under `Unreleased` for this JAX Host FFI behavior change.

shi-eric:
Please squash this PR down to a single coherent commit before merge.

shi-eric:
Please rebase onto current `main` and move the fix/tests to the promoted JAX code paths. Commit `604a8961df6d40ea64ff1e740b23581e4c72c96f` promoted the JAX code from `jax_experimental` to `jax` after this PR was opened, so the final diff should target the current locations.

warp/_src/jax/ffi.py
```diff
@@ -743,6 +768,35 @@ def ffi_callback(self, call_frame):
                 assert num_inputs == self.num_inputs
                 assert num_outputs == self.num_outputs
 
+                if platform == "Host":
```
nvlukasz: Can we deduplicate this argument reconstruction code? Perhaps extract to a helper function.

warp/_src/jax_experimental/ffi.py
```diff
@@ -875,16 +929,16 @@ def ffi_callback(self, call_frame):
                     arg_list.append(arr)
 
                 # call the Python function with reconstructed arguments
-                with wp.ScopedStream(stream, sync_enter=False):
-                    if stream.is_capturing:
+                with wp.ScopedStream(stream, sync_enter=False) if stream else wp.ScopedDevice(device):
```
nvlukasz: No need to check `stream` here, since the host path returns early above. If we get here, the stream is not None.

warp/_src/jax_experimental/ffi.py
```diff
@@ -875,16 +929,16 @@ def ffi_callback(self, call_frame):
...
+                    if stream and stream.is_capturing:
                         # capturing with JAX
                         with wp.ScopedCapture(external=True) as capture:
                             self.func(*arg_list)
 
                         # keep a reference to the capture object to prevent required modules getting unloaded
                         call_desc.capture = capture
 
-                    elif self.graph_mode == GraphMode.WARP:
+                    elif self.graph_mode == GraphMode.WARP and device.is_cuda:
```
nvlukasz: No need to check `device.is_cuda` here, since the host part returns early above. Same comment on a few more lines below.

nvlukasz (commented):
Thank you for the contribution. Please address outstanding comments or let us know if you are unable to do so.

loney7:
> Thank you for the contribution. Please address outstanding comments or let us know if you are unable to do so.

@nvlukasz , please allow till the end of this week to address the outstanding comments. I apologise for the delay.

loney7:
pre-commit.ci autofix

loney7:
pre-commit.ci autofix

loney7:
pre-commit.ci autofix

btaba:
Hi @nvlukasz wondering if you could take another pass :)

loney7:
pre-commit.ci autofix

loney7:
pre-commit.ci autofix

btaba:
friendly ping to @nvlukasz or @shi-eric (sorry for the trouble)

btaba:
friendly ping again @nvlukasz or @shi-eric

shi-eric:
Hey @btaba, did you check that the tests pass before asking for a re-review?

The newly added Host FFI tests fail both locally and in GitHub CI. The RTX PRO 6000 JAX job (`test-warp-gpu-linux`, `jax[cuda13]`) fails with the same two Host FFI failures I see locally: an exact-equality failure in `test_ffi_jax_kernel_host_sincos`, and a Host callable vec2 output shape mismatch in `test_ffi_jax_callable_host_scale_constant`.

The branch is also still conflicting with `main`, and the requested squash is pending. Please fix the failing tests, rebase onto current `main`, squash the PR, and then ping us again for review.

loney7:
pre-commit.ci autofix

loney7:
/ok to test https://github.com/NVIDIA/warp/pull/1446/changes/0644b335729982472891dd91351aa0c4681027b3

loney7:
/ok to test https://github.com/NVIDIA/warp/pull/1446/changes/865d5ef32bff3b3247892baef9483ce6d4b23e7a

shi-eric:
Could you also fix this Host `jax_kernel` placement case? It fails when `jax_kernel` is traced with CUDA as the effective JAX default device, but the actual call executes on CPU.

Minimal repro:

```python
import jax
import jax.numpy as jnp
import numpy as np
import warp as wp


@wp.kernel
def add_kernel(
    a: wp.array[wp.float32],
    b: wp.array[wp.float32],
    out: wp.array[wp.float32],
):
    tid = wp.tid()
    out[tid] = a[tid] + b[tid]


jax_add = wp.jax_kernel(add_kernel)


@jax.jit
def f(a, b):
    (out,) = jax_add(a, b)
    return out


cpu = jax.devices("cpu")[0]

# Leave the effective JAX default device as CUDA, but place inputs on CPU.
a = jax.device_put(jnp.arange(16, dtype=jnp.float32), cpu)
b = jax.device_put(jnp.ones(16, dtype=jnp.float32), cpu)

out = f(a, b)
jax.block_until_ready(out)

np.testing.assert_allclose(np.asarray(out), np.arange(1, 17, dtype=np.float32))
```

On this branch, `__call__` preloads the module for the default CUDA device. Execution then enters the Host callback, selects `wp.get_device("cpu")`, and fails when fetching CPU hooks:

```text
RuntimeError: Module is not loaded on device cpu
```

The Host callback should ensure the kernel module is loaded for CPU before calling `get_kernel_hooks(..., device)`. This would be good to cover with a regression test that leaves the effective JAX default device on CUDA while passing CPU-placed inputs.

shi-eric:
Could you also fix the generic Host FFI callback buffer path? The Host target is registered, but the buffers passed to the user callback are still `FfiBuffer` objects that only expose `__cuda_array_interface__`. On CPU, passing those buffers directly to `wp.launch(..., device="cpu")` fails because Warp looks for `__array_interface__`.

Minimal repro:

```python
import warnings

import jax
import jax.numpy as jnp
import warp as wp

with warnings.catch_warnings():
    warnings.simplefilter("ignore", DeprecationWarning)
    from warp.jax_experimental import register_ffi_callback


@wp.kernel
def double_kernel(inp: wp.array[wp.float32], out: wp.array[wp.float32]):
    tid = wp.tid()
    out[tid] = 2.0 * inp[tid]


def warp_callback(inputs, outputs, attrs, ctx):
    inp = inputs[0]
    out = outputs[0]

    assert ctx.stream is None
    assert hasattr(inp, "__cuda_array_interface__")
    assert not hasattr(inp, "__array_interface__")

    wp.launch(double_kernel, dim=inp.shape[0], inputs=[inp], outputs=[out], device="cpu")


register_ffi_callback("warp_generic_host_buffer_repro", warp_callback)


@jax.jit
def f(x):
    out_type = jax.ShapeDtypeStruct(x.shape, x.dtype)
    call = jax.ffi.ffi_call("warp_generic_host_buffer_repro", out_type)
    return call(x)


with jax.default_device(jax.devices("cpu")[0]):
    x = jnp.arange(16, dtype=jnp.float32)
    out = f(x)

jax.block_until_ready(out)
```

On this branch, the Host callback is reached, but the CPU launch fails while packing the `FfiBuffer`:

```text
RuntimeError: Error launching kernel 'double_kernel', argument 'inp' expects an array of type array(ndim=1, dtype=float32), but passed value has type FfiBuffer.
```

Can the Host generic callback path expose CPU-compatible buffers, e.g. via `__array_interface__`, so the same callback pattern supported on CUDA also works with `wp.launch(..., device="cpu")`?

loney7:
@shi-eric , can you please have a look

shi-eric:
Could you hard-code `block_dim=1` for Host callbacks while keeping CUDA at 256?

`jax_kernel()` does not expose `block_dim` to users, and Warp defines CPU tile execution as single-lane. Loading the CPU module with the default value of 256 compiles `WP_TILE_BLOCK_DIM` incorrectly, so explicitly shaped tile operations can process only lane 0's portion of the tile.

Something along these lines:

```python
block_dim = 256 if device.is_cuda else 1
module_exec = self.kernel.module.load(device, block_dim)
hooks = module_exec.get_kernel_hooks(self.kernel)
```

Could you also add a Host regression test that stores an explicitly shaped tile, such as `tile_ones(shape=64)`, and verifies that all elements are written?

warp/_src/jax/ffi.py
```diff
@@ -906,7 +963,7 @@ def ffi_callback(self, call_frame):
 
                 # call the Python function with reconstructed arguments
                 with wp.ScopedStream(stream, sync_enter=False):
```
shi-eric, line 965: On the Host platform `stream` is `None` here, and `ScopedStream(None)` is a no-op (its
`__enter__`/`__exit__` do nothing when the stream is `None`), so no device scope is
established around the callback. Device-less `wp.launch()` calls inside the callback then
resolve to Warp's default device, which is `cuda:0` on any machine with a visible GPU. The
kernels launch asynchronously on CUDA with CPU buffer pointers, producing silently wrong
output (or an async `CUDA error 700: illegal memory access`, depending on whether the GPU
can access pageable host memory).

This is the failure on the GPU runners in
[test-warp-gpu-linux](https://github.com/NVIDIA/warp/actions/runs/28826192211/job/85490549815?pr=1446)
and
[test-warp-gpu-linux-mgpu-python310](https://github.com/NVIDIA/warp/actions/runs/28826192211/job/85490549722?pr=1446):
`test_ffi_jax_callable_host_in_out` fails with stale/partially written output (8 of 1M
elements stale in one job, 100% in the other). The other host `jax_callable` tests only
pass by timing luck; the same pattern as `scale_func` produced wrong results on my machine.

```suggestion
                # ScopedStream makes the stream's device current, but there is
                # no stream on the Host platform, so use ScopedDevice to route
                # device-less launches in the callback to the CPU
                if stream is not None:
                    exec_scope = wp.ScopedStream(stream, sync_enter=False)
                else:
                    exec_scope = wp.ScopedDevice(device)

                with exec_scope:
```

Verified locally on 2x RTX 3090 (CPU jaxlib and `jax[cuda13]`): before this change, the
host `jax_callable` tests fail with GPUs visible and pass with `CUDA_VISIBLE_DEVICES=`;
after it, all host and CUDA FFI tests pass. The CUDA branch is unchanged.

A regression test can pin this down deterministically by asserting on the resolved device
instead of comparing outputs (output comparisons can pass by timing luck on drivers where
the GPU can dereference pageable host memory):

```python
@unittest.skipUnless(_jax_version() >= (0, 5, 0), "Jax version too old")
@unittest.skipUnless(wp.is_cpu_available(), "Requires CPU backend")
def test_ffi_jax_callable_host_default_device(test, device):
    """Verify that Host callbacks route device-less launches to the CPU.

    Regression test for the Host (CPU) FFI path: the callback must run with
    the CPU device current so that ``wp.launch()`` calls without an explicit
    device run on the CPU even when Warp's default device is a CUDA device.
    """
    import jax.numpy as jp  # noqa: PLC0415

    jax_callable = wp.jax_callable

    # devices resolved by device-less launches inside the callback
    resolved_devices = []

    def record_device_func(a: wp.array[float], b: wp.array[float]):
        resolved_devices.append(wp.get_device())
        wp.launch(scale_kernel, dim=a.shape, inputs=[a, 2.0], outputs=[b])

    jax_func = jax_callable(record_device_func)

    @jax.jit
    def f():
        a = jp.arange(ARRAY_SIZE, dtype=jp.float32)
        return jax_func(a, output_dims={"b": a.shape})

    # force a non-CPU default device to expose launches that don't inherit the
    # callback's device scope
    device_scope = wp.ScopedDevice("cuda:0") if wp.is_cuda_available() else contextlib.nullcontext()

    with device_scope, jax.default_device(jax.devices("cpu")[0]):
        (b,) = f()
        jax.block_until_ready(b)

    test.assertGreater(len(resolved_devices), 0)
    for d in resolved_devices:
        test.assertEqual(d, wp.get_device("cpu"))

    assert_np_equal(np.asarray(b), 2 * np.arange(ARRAY_SIZE, dtype=np.float32))
```

(plus an `add_function_test(TestJax, "test_ffi_jax_callable_host_default_device", ..., devices=None)`
registration next to the other host tests)

shi-eric:
Hey @loney7 we decided to move ahead with our own CPU support for JAX FFI because the review cycle here had taken too long and we wanted to get the feature into 1.16.0 instead of iterating on it indefinitely. It is now on `main` in dfec89720be3e2a0047499126e12861248c21ce9 and tracked in #1661.

The version we merged is not the same patch. It covers `wp.jax_kernel()` and `wp.jax_callable()` on CPU and CUDA, loads modules on the device that actually runs the callback, and compiles CPU kernels with a single-lane block dimension. It also defines how graph modes behave on CPU and has broader CPU-only and mixed-device testing. We did not carry over Host support for the generic `register_ffi_callback()` path.

Please try the version on `main` with your use case. If you still need the generic callback path or anything else from this PR, open a focused follow-up with a small reproducer and we can look at it separately.

The core functionality is now on `main`, so we are closing this PR as redundant.

PR #1475, opened 2026-05-22
Reject non-None kernel return annotations

### LINKED ISSUE #1471: Kernel return annotations are silently ignored
### Bug Description

Warp kernels map to CPU/CUDA kernel entry points and do not return values directly. Outputs are written through explicit output arguments or side effects. The docs already describe this model, and Warp generally treats type annotations as part of the typed codegen contract rather than as cosmetic Python metadata.

Today, a kernel can declare a non-`None` return annotation and Warp accepts it silently when no value is returned:

```python
import warp as wp

@wp.kernel
def annotated_kernel(x: float) -> float:
    return

wp.launch(annotated_kernel, dim=1, inputs=[1.0], device="cpu")
```

Observed behavior on current `main`: the kernel definition succeeds and the launch succeeds. The `-> float` annotation is ignored.

A value-returning kernel does fail today:

```python
import warp as wp

@wp.kernel
def returning_kernel(x: float):
    return x

wp.launch(returning_kernel, dim=1, inputs=[1.0], device="cpu")
```

Observed behavior: the kernel definition succeeds, then launch fails with:

```text
WarpCodegenTypeError: Error, kernels can't have return values
```

For comparison, these valid forms both define and launch successfully:

```python
@wp.kernel
def kernel_with_none_annotation() -> None:
    return

@wp.kernel
def kernel_without_return_annotation():
    return
```

Expected behavior: non-`None` return annotations on kernels should be reported as invalid, because kernels do not return values. Value-returning kernels should continue to fail, with a diagnostic that points users at the same rule: kernels do not return values; write results to output arguments, and either omit the return annotation or use `-> None`.

This is confusing because annotations elsewhere in Warp affect codegen and specialization. Silently ignoring a kernel return annotation makes `-> float` look like a meaningful contract even though Warp cannot honor it.

### System Information

Observed on current `main` at `882c8d391`.

```text
Warp version: 1.14.0.dev0
Device used for repro: cpu
```

snico432:
## Description

Raises an error when a kernel has a non-`None` return annotation. Closes #1471.

## Checklist

## Test plan

Tested on CPU with:

```bash
uv run warp/tests/test_codegen.py TestCodeGen.test_error_kernel_return_value_cpu
```

## Bug fix

Previously, the `float` return annotation was silently ignored by Warp:

```python
import warp as wp

@wp.kernel
def annotated_kernel(x: float) -> float:
    return

wp.launch(annotated_kernel, dim=1, inputs=[1.0], device="cpu")
```

This PR makes that code raise:

```text
WarpCodegenTypeError: 'annotated_kernel': Error, kernels can't have return values; write results to output arguments and either omit the return annotation or use '-> None'
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Kernels with non-None return type annotations now raise a clear error, with guidance to write results to output arguments or use a None return.

* **Documentation**
  * Changelog updated with an Unreleased entry describing the behavior change.

* **Tests**
  * Tests updated to cover and assert the new error behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/warp/pull/1475?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,7 @@
   ([GH-1466](https://github.com/NVIDIA/warp/issues/1466)).
 - Fix tile byte-offset overflow for arrays larger than 2 GiB
   ([GH-1422](https://github.com/NVIDIA/warp/issues/1422)).
+- Raise an error when a kernel is defined with a non-`None` return annotation ([GH-1471](https://github.com/NVIDIA/warp/issues/1471)).
 
 ### Documentation
 
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -2288,6 +2288,12 @@ class ModuleBuilder:
         if kernel.adj.return_var is not None:
             raise WarpCodegenTypeError(f"'{kernel.key}': Error, kernels can't have return values")
 
+        if kernel.adj.arg_types.get("return") is not None:
+            raise WarpCodegenTypeError(
+                f"'{kernel.key}': Error, kernels can't have return values; "
+                "write results to output arguments and either omit the return annotation or use '-> None'"
+            )
+
     def build_function(self, func):
         if func in self.functions:
             return
--- a/warp/tests/test_codegen.py
+++ b/warp/tests/test_codegen.py
@@ -565,13 +565,16 @@ def test_error_kernel_return_value(test, device):
     with test.assertRaisesRegex(wp.WarpCodegenTypeError, r".*Error, kernels can't have return values"):
         wp.launch(f3, dim=1, inputs=[3.0], device=device)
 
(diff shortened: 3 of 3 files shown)


shi-eric:
Thanks for picking up #1471 and putting together this focused fix. The issue has since been resolved on `main` in 2abd428dbee34412b6b8f98a717755d112ecf97a with a broader implementation that covers the same non-`None` return annotation case, along with value-return statements and overloaded kernel stubs.

Since #1471 and the behavior addressed here are now covered, I am closing this PR as superseded. Thanks again for the contribution!

PR #1503, opened 2026-06-01
Report Clang CUDA compile errors (GH-1325)

### LINKED ISSUE #1325: Clang CUDA compilation return value not checked in Python
`warp/_src/build.py` ~L53 calls `wp_compile_cuda` but discards the return value. The function returns `ctypes.c_int` (0 on success, -1 on failure), but the caller does not check it. The NVRTC path at ~L97 checks `if err != 0: raise Exception(...)`.

When compilation fails, the NVRTC path raises immediately with the compilation error. The Clang CUDA path silently continues, and the missing PTX file causes a downstream exception that doesn't indicate the root cause.

**Repro:**

```python
import sys
import warp as wp

mode = sys.argv[1] if len(sys.argv) > 1 else "nvrtc"
wp.config.llvm_cuda = (mode == "clang")

@wp.func_native("""
    INVALID SOURCE;
""")
def bad_func(x: float) -> float: ...

@wp.kernel
def test_kernel(a: wp.array(dtype=float)):
    i = wp.tid()
    a[i] = bad_func(1.0)

a = wp.zeros(10, dtype=float, device="cuda:0")
wp.launch(test_kernel, dim=10, inputs=[a], device="cuda:0")
```

**NVRTC (`uv run repro.py nvrtc`):**
```
Exception: CUDA kernel build failed with error code 6
```

**Clang CUDA (`uv run repro.py clang`):**
```
Exception: Failed to load CUDA module '__main__'
```

Both paths print the Clang/NVRTC compiler error to stderr, but only the NVRTC path raises a Python exception that identifies the failure as a compilation error.

snico432:
## Description

A failed Clang CUDA compilation now raises an exception immediately instead of silently continuing and causing a downstream module-load exception that doesn't indicate the root cause. Closes #1325.

## Checklist

## Test plan

Added `test_invalid_native_func_compile_error` in `warp/tests/cuda/test_clang_cuda.py`.

Ran:

```bash
uv run warp/tests/cuda/test_clang_cuda.py -k invalid_native_func_compile_error
```

Result: test collected and skipped locally because I don't have a CUDA device available.

## Bug fix

**Repro from #1325:**

```python
import sys
import warp as wp

mode = sys.argv[1] if len(sys.argv) > 1 else "nvrtc"
wp.config.llvm_cuda = (mode == "clang")

@wp.func_native("""
    INVALID SOURCE;
""")
def bad_func(x: float) -> float: ...

@wp.kernel
def test_kernel(a: wp.array(dtype=float)):
    i = wp.tid()
    a[i] = bad_func(1.0)

a = wp.zeros(10, dtype=float, device="cuda:0")
wp.launch(test_kernel, dim=10, inputs=[a], device="cuda:0")
```

**NVRTC (uv run repro.py nvrtc):**

```
Exception: CUDA kernel build failed with error code 6
```

**Clang CUDA (uv run repro.py clang):**

```
Exception: Failed to load CUDA module '__main__'
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Clang/LLVM CUDA compilation failures are now reported immediately during compilation rather than surfacing later as module load errors, improving error clarity and faster failure detection.

* **Tests**
  * Added test coverage that verifies native-function/CUDA compile failures are detected and reported at build time to prevent silent downstream errors.

* **Documentation**
  * Changelog updated to note the improved immediate reporting of CUDA compilation failures.

### DIFF
--- a/warp/tests/cuda/test_clang_cuda.py
+++ b/warp/tests/cuda/test_clang_cuda.py
@@ -127,6 +127,23 @@ def test_bf16_arithmetic(test, device):
     np.testing.assert_allclose(out.numpy(), a_np + b_np, rtol=1e-2)
 
 
+def test_invalid_native_func_compile_error(test, device):
+    @wp.func_native("""
+        INVALID SOURCE;
+    """)
+    def bad_func(x: float) -> float: ...
+
+    @wp.kernel(module="unique")
+    def invalid_native_func_kernel(a: wp.array[float]):
+        i = wp.tid()
+        a[i] = bad_func(1.0)
+
+    a = wp.zeros(10, dtype=float, device=device)
+
+    with test.assertRaisesRegex(Exception, "CUDA kernel build failed with error code"):
+        wp.launch(invalid_native_func_kernel, dim=10, inputs=[a], device=device)
+
+
 class TestClangCUDA(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
@@ -146,7 +163,9 @@ add_function_test(TestClangCUDA, "test_conditional_kernel", test_conditional_ker
 bf16_devices = [d for d in devices if d.arch >= 80]
 add_function_test(TestClangCUDA, "test_bf16_round_trip", test_bf16_round_trip, devices=bf16_devices)
 add_function_test(TestClangCUDA, "test_bf16_arithmetic", test_bf16_arithmetic, devices=bf16_devices)
-
+add_function_test(
+    TestClangCUDA, "test_invalid_native_func_compile_error", test_invalid_native_func_compile_error, devices=devices
+)
 
 if __name__ == "__main__":
     unittest.main(verbosity=2)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,6 +234,9 @@
 - Fix `Tape.record_scope_end()` to raise a clear error for unmatched
   scope ends and preserve nested non-empty tape visualization scopes
   ([GH-1515](https://github.com/NVIDIA/warp/issues/1515)).
+- Report Clang CUDA compilation failures immediately instead of raising a
+  downstream CUDA module load error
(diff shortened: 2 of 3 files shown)


c0d1f1ed:
I think this looks good. @snico432 what prompted wanting to fix this? The Clang compilation path for CUDA is just in a prototype state in case we have a use for it later or need another compiler to compare to. Just curious if you have a specific use for it.

snico432:
@c0d1f1ed I don't have a specific use case. An engineer from your team came to one of my lectures and gave a presentation about the project and I found it interesting. I'm just trying to learn more about compilers while hopefully making some open-source contributions.

warp/tests/cuda/test_clang_cuda.py
```diff
@@ -50,6 +50,18 @@ def conditional_kernel(
...
+""")
+def bad_func(x: float) -> float: ...
+
+
+@wp.kernel(module="unique")
+def invalid_native_func_kernel(a: wp.array[float]):
+    i = wp.tid()
+    a[i] = bad_func(1.0)
+
+
```
shi-eric: These declarations can probably be moved into `test_invalid_native_func_compile_error` since the kernel is using `module="unique"`

shi-eric:
Thanks for the contribution, @snico432! I appreciate that you kept the branch consistently at one commit that was up-to-date with `main` which made it easy to merge this on our internal GitLab repository. The addition of tests, a changelog entry, and DCO signoff also saved some time.

Some minor feedback for future pull requests in the repo:

- Communicate in the original GitHub issue that you would like to work on it
- Allow the original commenter of a thread to decide whether to resolve it rather than resolving it yourself

There is overhead in reviewing pull requests, and our team is pretty small, so we would prefer to spend our time reviewing more substantive pull requests. But everyone has to start somewhere, and your pull request was already in a great shape :smiley:

snico432:
Thanks for the feedback @shi-eric! I'll make those adjustments in future PRs.

PR #1507, opened 2026-06-02
Fix BVH refit root leaves (GH-1506)

### LINKED ISSUE #1506: Bottom-up bvh refit bug
### Bug Description

There's an out-of-bounds access bug in bvh_refit_kernel for cases where there's only a single node in the tree. The fix is to add a check for the special case like so:
        if (leaf) {
            BVHPackedNodeHalf& lower = node_lowers[index];
            BVHPackedNodeHalf& upper = node_uppers[index];
            // update the leaf node

            // only need to compute bound when this is a valid leaf node. Be 
            // careful to handle the special case of only a single node in
            // the hierarchy, which won't have a valid parent index
            if ((index == 0) || (parent != -1 && !node_lowers[parent].b)) {

### System Information

_No response_

shi-eric:
## Description

Closes #1506.

Fix CUDA `wp.Bvh.refit()` for root-leaf cases. The device refit kernel now treats `parent == -1` and `parent_parent == -1` as valid leaf/root cases before reading parent node metadata, avoiding out-of-bounds reads for single-node BVHs and packed LBVH roots.

The bug was specific to the plain BVH CUDA refit path. The solid-angle mesh refit path already had equivalent guards.

## Checklist

## Test plan

- `uv run build_lib.py --quick`
- `WARP_CACHE_PATH=/tmp/warp-cache-fix-bvh-refit-single-node-final WARP_CACHE_ROOT=/tmp/warp-cache-fix-bvh-refit-single-node-final uv run --extra dev -m warp.tests -s autodetect -k TestBvh.test_bvh_refit_single_leaf -k TestBvh.test_bvh_refit_packed_root`
- `WARP_CACHE_PATH=/tmp/warp-cache-fix-bvh-refit-single-node-bvhfile WARP_CACHE_ROOT=/tmp/warp-cache-fix-bvh-refit-single-node-bvhfile uv run --extra dev -m warp.tests -s autodetect -k TestBvh`
- `uvx pre-commit run --files CHANGELOG.md warp/native/bvh.cu warp/tests/geometry/test_bvh.py`

## Bug fix

The out-of-bounds read is most directly observed with CUDA memory checking tools. In normal execution it may not surface as a deterministic Python exception because the invalid read can land in accessible allocator padding.

```python
import warp as wp

lowers = wp.array([wp.vec3(0.0, 0.0, 0.0)], dtype=wp.vec3, device="cuda")
uppers = wp.array([wp.vec3(1.0, 1.0, 1.0)], dtype=wp.vec3, device="cuda")
bvh = wp.Bvh(lowers, uppers, constructor="lbvh")
bvh.refit()
```

## New feature / enhancement

Not applicable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an out-of-bounds memory access in CUDA BVH refit affecting single-node and packed-root tree refits, improving stability on CUDA devices.

* **Tests**
  * Added a consolidated regression test validating BVH refit updates bounds and restores query results for single-leaf and packed-root scenarios.

* **Documentation**
  * Updated changelog to note the BVH refit fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -216,6 +216,9 @@
   ``f`` is such a local now compile and call correctly instead of failing during codegen.
   Rebinding a function-valued local to a different function or to a non-function value raises a clear error
   ([GH-1423](https://github.com/NVIDIA/warp/issues/1423)).
+- Fix CUDA BVH refit out-of-bounds access for single-node and packed-root
+  trees ([GH-860](https://github.com/NVIDIA/warp/issues/860),
+  [GH-1506](https://github.com/NVIDIA/warp/issues/1506)).
 
 ### Documentation
 
--- a/warp/native/bvh.cu
+++ b/warp/native/bvh.cu
@@ -71,8 +71,8 @@ __global__ void bvh_refit_kernel(
             BVHPackedNodeHalf& upper = node_uppers[index];
             // update the leaf node
 
-            // only need to compute bound when this is a valid leaf node
-            if (!node_lowers[parent].b) {
+            // Single-node trees store the root leaf without a parent.
+            if (parent == -1 || !node_lowers[parent].b) {
                 const int start = lower.i;
                 const int end = upper.i;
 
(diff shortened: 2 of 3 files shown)


AnkaChan:
I will take a look this week thank you for fixing this!

Issue #549, opened 2025-02-24
PEP 484 Type Hint Support


shi-eric:
This is meant to be a parent issue for all kinds of issues regarding the use of type annotations in Warp.

References

- https://numpy.org/devdocs/reference/typing.html#ndarray
- https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#decorators

PR #1539, opened 2026-06-08
Infer wp.array DType from argument in constructor

thomasbbrunner:
## Description

_NOTE: this is a draft. Opening this PR for comments/feedback._

Currently, `wp.array` constructors & factories (zeros/ones/empty/full/from_numpy/...) infer the type as `array[Any, int]` because their signatures don't expose the dtype as a generic parameter.

For instance, notice how the `DType` of the `class array(Array[DType, NDim])` is always `Any`:

```python
a = wp.array([1, 2, 3], dtype=wp.int32)
reveal_type(a)
# output: array[Any, int]

f = wp.full(10, 7, dtype=wp.float32)
reveal_type(f)
# output: array[Any, int]
```

This causes `mypy` to complain when in strict mode: `error: Need type annotation for "a"  [var-annotated]`

To fix this, users must either explicitly set the type or ignore the error. Either fix is a workaround and imo suboptimal.

Here, I'm proposing to set the array's `DType` according to the `dtype` argument in the constructor. With this, the type is correctly set:

```python
a = wp.array([1, 2, 3], dtype=wp.int32)
reveal_type(a)
# output: array[int32, int]

f = wp.full(10, 7, dtype=wp.float32)
reveal_type(f)
# output: array[float32, int]
```

There are no open issues specifically describing this. This is somewhat of an extension to https://github.com/NVIDIA/warp/issues/549.

Is there interest in fixing this? If so, I'd update this MR to:
- add tests for this (potentially calling mypy to assert that the revealed types match the expectations), and,
- extend this pattern to other constructors (zeros_like/*_like, from_torch/paddle/jax/dlpack).

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved type annotations for array creation functions (`zeros`, `ones`, `empty`, `full`, `from_numpy`) with support for generic type parameters, enabling better type inference and IDE support.

### DIFF
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -66,7 +66,7 @@ import warp.config
 from warp._src.codegen import WarpCodegenTypeError, synchronized
 from warp._src.logger import LOG_DEBUG, LOG_WARNING, get_logger, log_debug, log_error, log_info, log_warning
 from warp._src.texture import Texture1D, Texture2D, Texture3D, texture1d_t, texture2d_t, texture3d_t
-from warp._src.types import LAUNCH_MAX_DIMS, Array, LaunchBounds, launch_bounds_t, type_repr
+from warp._src.types import LAUNCH_MAX_DIMS, Array, DType, LaunchBounds, launch_bounds_t, type_repr
 
 _wp_module_name_ = "warp.context"
 
@@ -7647,13 +7647,13 @@ class RegisteredGLBuffer:
 
 def zeros(
     shape: int | tuple[int, ...] | list[int] | None = None,
-    dtype: type = float,
+    dtype: type[DType] | None = None,
     device: DeviceLike = None,
     requires_grad: bool = False,
     pinned: bool = False,
     retain_grad: bool = False,
     **kwargs,
-) -> warp.array:
+) -> warp.array[DType]:
     """Return a zero-initialized array.
 
     Args:
@@ -7667,6 +7667,8 @@ def zeros(
     Returns:
         A warp.array object representing the allocation
     """
+    if dtype is None:
+        dtype = float
 
     arr = empty(
         shape=shape,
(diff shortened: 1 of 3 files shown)


warp/_src/context.py
```diff
@@ -7643,13 +7647,13 @@ def unmap(self):
 
 def zeros(
     shape: int | tuple[int, ...] | list[int] | None = None,
-    dtype: type = float,
+    dtype: type[DType] | None = None,
```
thomasbbrunner, line 7650: The default was changed to `None`.

This is to avoid a type error, because the concrete default float would pin the free DType TypeVar.

There is precedent to this (`wp.full` already uses this pattern).

This changes the behavior of the code. `wp.zeros(10, dtype=None)` is now a legal "use the default". Before it raised a `TypeError` in `type_size_in_bytes`.

thomasbbrunner:
@shi-eric could you take a look into this and let me know if there's interest in fixing this? If so, I'd work on a proper PR.

Would be interested in your opinion on how to add a test for this.

warp/_src/context.py
christophercrouzet: It's better to have `float` set as the default parameter value since the expectations are clear from the signature but, other than that, I don't think it's too much of an issue since it's a common Python idiom to consider `None` as a way to express “no value has been passed” and define the default behaviour inside the function's body.

christophercrouzet:
Thanks @thomasbbrunner!

At first glance, I'm not seeing any obvious drawback to have these changes added, so let's give it a try.

Could you please:

- Open a new issue.
- Wrap up the other changes you planned on adding.
- Add a CHANGELOG entry.
- Reference the issue in this PR's title, the commit message, and the CHANGELOG.
- Edit your commit message to include the DCO.

See https://nvidia.github.io/warp/stable/user_guide/contribution_guide.html for details.

As for the tests you suggested, we currently don't cover typing behaviour with our tests, but @shi-eric suggested extending the [`type-check-stubs` CI job](https://github.com/NVIDIA/warp/blob/efeca6ca87421a08477a2d8cda4c48aae927a28d/.github/workflows/ci.yml#L837-L857) with a new step that would check the `reveal_type()` output (maybe using a `python -c` inline command if it's small enough?).

Issue #1561, opened 2026-06-16
Allow passing a key/name to the `@wp.kernel`


thomasbbrunner:
### Description

`@wp.kernel` derives a kernel's key from `make_full_qualified_name(func)`, which in turn uses the function's `__qualname__`. There's no way to override it.

The `@wp.func` decorator already exposes a `name` argument for the same purpose.

It would be great if we would extend this pattern to the `wp.kernel`, enabling overriding the kernel's name through an argument: `@wp.kernel(key="custom_name")`

### Context

This is relevant when generating kernels using [kernel closures](https://nvidia.github.io/warp/stable/deep_dive/codegen.html#kernel-closures). In this case, different kernels have the same key and only differ by their hashes:

```python
import warp as wp

def make_scaler(factor: float) -> wp.Kernel:
    @wp.kernel
    def scale(x: wp.array(dtype=wp.float32)):
        i = wp.tid()
        x[i] = x[i] * factor
    return scale

double = make_scaler(2.0)
triple = make_scaler(3.0)

print(double.key)  # make_scaler__locals__scale
print(triple.key)  # make_scaler__locals__scale  <-- identical!
```

In our case, we export these kernels with `compile_aot_module` with `strip_hash=True`. This results in a collision of the kernel names. We strip the hashes to enable more flexible access to the exported code.

### Proposal

Expose a `wp.kernel(key=...)` or `wp.kernel(name=...)` (to be consistent with `wp.func`) argument, which overrides the kernel's name and skips the sanitation of the name with `make_full_qualified_name`.

A potential pitfall is if the user passes a name that is not a valid C name. We should make it clear in the docstring which requirements the name needs to satisfy. Alternatively, we could check the user's custom name and raise if it is not valid.

Happy to implement this, if there's interest!

thomasbbrunner:
Addressed in https://github.com/NVIDIA/warp/pull/1570

c0d1f1ed:
Thanks for filing this issue and providing a PR!

I'll take it for internal review and try to get it merged for the 1.16 release. I think I'll make the following refinements:
- If `strip_hash=True` causes distinct kernels to have the same exported native name, fail early with a clear message.
- Clarify the difference between "key" (implementation detail), kernel name, and exported symbol name.
- Revisit the `wp.overload` handling.

EDIT: Let's get the PR merged. I can follow up with the refinements.

shi-eric:
[...]

PR #1570, opened 2026-06-19
Add name argument to @wp.kernel (GH-1561)

thomasbbrunner:
## Description

Addresses https://github.com/NVIDIA/warp/issues/1561.

`@wp.kernel` derives a kernel's key from `make_full_qualified_name(func)`, which in turn uses the function's `__qualname__`. There's no way to override it.

The `@wp.func` decorator already exposes a `name` argument for the same purpose.

## Changes

Here I'm adding support for a `name` argument for the `wp.kernel` decorator, enabling overriding the kernel's name.

I chose `name` to be consistent with the `wp.func` API. However, internally this is called `key`.

A potential pitfall is if the user passes a name that is not a valid C name. I've added an explicit check for this, otherwise users are faced with a somewhat cryptic compiler error.

## New feature / enhancement

This is especially useful when generating kernels with the [closure pattern](https://nvidia.github.io/warp/stable/deep_dive/codegen.html#kernel-closures). In this case, different kernels have the same key and only differ by their hashes:

```python
import warp as wp

def make_scaler(factor: float) -> wp.Kernel:
    @wp.kernel
    def scale(x: wp.array(dtype=wp.float32)):
        i = wp.tid()
        x[i] = x[i] * factor
    return scale

double = make_scaler(2.0)
triple = make_scaler(3.0)

print(double.key)  # make_scaler__locals__scale
print(triple.key)  # make_scaler__locals__scale  <-- identical!
```

With these changes, users can now specify a name for the kernel:

```python
def make_scaler(factor: float) -> wp.Kernel:
    @wp.kernel(name=f"scale{factor}".replace(".", "_"))
    def scale(x: wp.array(dtype=wp.float32)):
        i = wp.tid()
        x[i] = x[i] * factor
    return scale
```

## Checklist

## Validation summary

Ran all tests and linter.

## Summary by CodeRabbit

* **New Features**
  * `@wp.kernel` now accepts an optional `name` to control the kernel identifier, with C-identifier validation.
  * Added a public utility to check whether a string is a valid C identifier.
* **Bug Fixes**
  * Improved kernel lookup to correctly resolve renamed kernels (preferring the most recently registered match).
* **Tests**
  * Added unit coverage for `@wp.kernel` name override/default/invalid cases, identifier validation, and renamed-kernel behavior with `@wp.overload`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
   ([GH-1378](https://github.com/NVIDIA/warp/issues/1378))
 - **Experimental:** Record `wp.utils.array_sum()` and `wp.utils.array_inner()` in APIC capture so saved and replayed
   graphs recompute them from current inputs on CPU and CUDA ([GH-1663](https://github.com/NVIDIA/warp/issues/1663)).
+- Add a `name` argument to `@wp.kernel` to override the kernel's generated key
+  ([GH-1561](https://github.com/NVIDIA/warp/issues/1561)).
 
 ### Removed
 
--- a/warp/_src/codegen.py
+++ b/warp/_src/codegen.py
@@ -6317,6 +6317,16 @@ def make_full_qualified_name(func: str | Callable) -> str:
     return re.sub("[^0-9a-zA-Z_]+", "", func.replace(".", "__"))
 
 
+# matches a valid C identifier: a letter or underscore followed by letters, digits, or underscores
+_C_IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+
+def is_valid_c_identifier(name: str) -> bool:
+    # Note: This only validates the lexical shape. It does not reject reserved
+    # C/C++/CUDA keywords.
+    return _C_IDENTIFIER_RE.match(name) is not None
+
+
 def codegen_struct(struct, device="cpu", indent_size=4, include_tile_helpers=False):
     name = struct.native_name
 
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -1041,8 +1041,13 @@ class Kernel:
         else:
             self.module = module
 
+        # Keep track of the qualified name of the Python function used to create
+        # this kernel, so it can still be identified even when the kernel is given
+        # a custom name (key).
+        self.qualname = warp._src.codegen.make_full_qualified_name(func)
+
         if key is None:
-            self.key = warp._src.codegen.make_full_qualified_name(func)
+            self.key = self.qualname
         else:
             self.key = key
 
@@ -1610,6 +1615,7 @@ def grad(func: Callable) -> GradWrapper:
 def kernel(
     f: Callable | None = None,
     *,
+    name: str | None = None,
     enable_backward: bool | None = None,
     launch_bounds: tuple[int, ...] | int | None = None,
     cluster_dim: int | None = None,
@@ -1660,6 +1666,9 @@ def kernel(
 
     Args:
         f: The function to be registered as a kernel.
+        name: Overrides the kernel's key. If ``None``, the function's
+            qualified name will be used as the key. The name must have the
+            shape of a valid C identifier; raises a ``ValueError`` otherwise.
         enable_backward: If False, the backward pass will not be
             generated.
         launch_bounds: CUDA ``__launch_bounds__`` attribute for the
@@ -1702,6 +1711,13 @@ def kernel(
 
     def wrapper(f, *args, **kwargs):
         kernel_options = {}
+
+        if name is None:
+            key = warp._src.codegen.make_full_qualified_name(f)
+        else:
+            if not warp._src.codegen.is_valid_c_identifier(name):
+                raise ValueError(f"@wp.kernel for '{f.__name__}': name '{name}' is not a valid C identifier.")
+            key = name
 
         if enable_backward is not None:
             kernel_options["enable_backward"] = enable_backward
@@ -1755,7 +1771,7 @@ def kernel(
         # Create the kernel object and register it with the module
         k = Kernel(
             func=f,
-            key=warp._src.codegen.make_full_qualified_name(f),
+            key=key,
             module=m,
             options=kernel_options,
         )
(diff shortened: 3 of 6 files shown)


warp/_src/context.py
```diff
@@ -1448,6 +1452,13 @@ def my_kernel_fast(a: wp.array[float], b: wp.array[float]):
     def wrapper(f, *args, **kwargs):
         kernel_options = {}
 
+        if name is None:
+            key = warp._src.codegen.make_full_qualified_name(f)
+        else:
+            if not warp._src.codegen.is_valid_c_identifier(name):
+                raise ValueError(f"@wp.kernel for '{f.__name__}': name '{name}' is not a valid C identifier.")
+            key = name
```
shi-eric, line 1720: Please fix decorator-form `@wp.overload` with renamed generic kernels.

This branch registers a kernel only under the explicit `name=` key, but the documented overload-stub path still resolves the kernel from the stub's Python qualified name via `Module._find_kernel()`. That means this now fails:

```python
from typing import Any

import warp as wp


@wp.kernel(name="custom_scale")
def scale(x: wp.array[Any], s: Any):
    i = wp.tid()
    x[i] = s * x[i]


@wp.overload
def scale(x: wp.array[wp.float32], s: wp.float32):
    ...
```

Observed on this branch:

```text
RuntimeError: Failed to find a kernel named 'scale' in module __main__
```

Direct overload registration still works:

```python
wp.overload(scale, [wp.array[wp.float32], wp.float32])
```

so the regression is specifically in the decorator-form lookup. I think the fix should preserve lookup by the original Python qualified name, or otherwise teach `_find_kernel()` / `@wp.overload` how to find kernels whose key was overridden by `@wp.kernel(name=...)`.

Please also add a regression test covering `@wp.kernel(name="...")` plus decorator-form `@wp.overload`.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Add a `name` argument to `@wp.kernel` to override the kernel's key, mirroring `@wp.func`.
+  ([GH-1561](https://github.com/NVIDIA/warp/issues/1561)).
```
shi-eric: Nitpick: Please revise the `CHANGELOG.md` entry to avoid saying this mirrors `@wp.func`.

`@wp.func(name=...)` is useful precedent for the API shape, but the changelog should describe the kernel behavior directly rather than compare it to a different decorator.

Suggested wording:

```markdown
Add a `name` argument to `@wp.kernel` to override the kernel's generated key.
```

thomasbbrunner: Agreed, done!

warp/_src/context.py
thomasbbrunner: Indeed, thanks for spotting this issue.

From looking at the code, there are a couple ways to address this. I took the liberty of implementing one of them.

For completeness, these are the options considered:
1. Key self.kernels by kernel.qualname instead of kernel.key. This makes the _find_kernel() implementation pretty simple, but messes with the AOT and causes collisions deeper in the code. Not really a valid solution.
1. Store the kernel.qualname next to kernel.key. If _find_kernel fails to find a kernel from the func's qualname, assume it was renamed. Then iterate over the kernels in the module, and compare the qualnames. **This is the one implemented.**
1. Keep another dict self.kernels_by_qualname next to self.kernels in the Module class. _find_kernel() uses this other dict. Drawback is that we need to keep these two data structures in sync.

Open to iterating over this.

thomasbbrunner:
Added a fix for the `wp.overload` edge-case you highlighted, @shi-eric. Thanks for spotting it.

Now, it can happen that when using the closure pattern with wp.overload, we get multiple possible matches in _find_kernel().

For instance, in code below, the overload would find two kernels with the qualname `scale` when running `make("b")`.

```python
    def make(suffix):
        @wp.kernel(name=f"kernel_{suffix}")
        def scale(x: Any):
            return

        @wp.overload  # <-- this overload will match the first registered `scale` kernel
        def scale(x: float): ...

        return scale

    first = make("a")
    second = make("b")
```

I've added some logic to address this, namely, picking the last match, which should correspond to the correct target kernel.

Admittedly, this check is fragile. It assumes that the `@wp.overload` is defined just after the overloaded kernel, `self.kernels` preserves insertion order and that the `self.kernels` keys aren't overwritten.

Alternatively, we can raise an error if multiple matches are found. In this case, users would need to pass a custom name that is different for each closure case.

_Note: I've added the fix on a separate commit for easier review. I'd squash everything before merging._

c0d1f1ed:
@thomasbbrunner we've decided that if you address the above review comments and rebase we can probably take this PR as-is. 🙏 I can do a minor follow-up if needed.

warp/_src/codegen.py
```diff
@@ -5131,6 +5131,14 @@ def make_full_qualified_name(func: str | Callable) -> str:
     return re.sub("[^0-9a-zA-Z_]+", "", func.replace(".", "__"))
 
 
+# matches a valid C identifier: a letter or underscore followed by letters, digits, or underscores
+_C_IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+
+def is_valid_c_identifier(name: str) -> bool:
+    return _C_IDENTIFIER_RE.match(name) is not None
```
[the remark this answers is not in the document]
thomasbbrunner: Fair, but it's not easy to have a complete check for this. I could look into it, but I think the complexity is not warranted.

warp/_src/context.py
thomasbbrunner: I've looked into this again and still think that the current implementation is the best approach.

Would leave as is.

thomasbbrunner:
@c0d1f1ed happy to push this over the finish line.

I've addressed the remaining points, squashed the commits and rebased.

There are some design questions regarding the solution to @shi-eric's comment in https://github.com/NVIDIA/warp/pull/1570#discussion_r3448204618. I've implemented a valid solution, happy to iterate over it.

From my side, this should be ready!

c0d1f1ed (changes requested):
https://github.com/NVIDIA/warp/commit/ffce87624195f2b7a899074d910edf1c2632b91b included support for naming kernels. Please verify if it satisfies GH-1561 and close this PR, or rework it if more changes are needed.

thomasbbrunner:
Hi @c0d1f1ed, thanks for the update. The changes seem to fit our use-case, so this PR shouldn't be needed anymore. I'll wait for the next release to test it out though. If needed, I can re-open this.

Thanks!

PR #1571, opened 2026-06-20
Fix NVRTC include path option storage

fallintoplace:
## Description

`wp_cuda_compile_program()` built the main NVRTC include option in a fixed-size stack buffer. The guard checked only `include_dir`, but the buffer also had to hold the `--include-path=` prefix and the null terminator. That allowed paths that passed the check to overflow the stack buffer while constructing the option.

This changes the main include option to use the same dynamic string storage pattern used for the other generated NVRTC options, while reserving capacity up front so the `c_str()` pointers passed to NVRTC stay stable until compilation.

## Changes

- Store the primary `--include-path=` option in `std::vector<std::string>` instead of a fixed-size `char` buffer.
- Reserve storage for all generated option strings before taking `c_str()` pointers.
- Remove the incorrect length check and the `strcpy()`/`strcat()` construction.

## Checklist

## Validation summary

I verified the focused source diff and formatting locally:

- `git diff --check`
- `uvx pre-commit run --files warp/native/warp.cu`
- `uv run build_lib.py --quick`

The quick build succeeded on macOS arm64 in CPU-only mode. This machine does not have a CUDA toolchain available, so I could not run an NVRTC/CUDA compile-path regression test locally.

## Bug fix

The issue can be triggered when Warp is installed or checked out under a very long path and a CUDA kernel compile uses that path as `warp_home/native` for the include directory. Without this change, the fixed stack buffer can overflow while prepending `--include-path=`.

```python
import warp as wp


@wp.kernel
def noop():
    pass


wp.init()
wp.launch(noop, dim=1, inputs=[], device="cuda:0")
```

## New feature / enhancement

Not applicable.

## Release Notes

* **Refactor**
  * Improved CUDA compilation memory management by transitioning include path handling from fixed-size buffers to dynamic allocation. This enhancement increases system reliability, removes previous capacity constraints, and provides better flexibility for projects requiring extensive include path configurations.

### DIFF
--- a/warp/native/warp.cu
+++ b/warp/native/warp.cu
@@ -4439,13 +4439,6 @@ size_t wp_cuda_compile_program(
     bool use_ptx = output_ext && strcmp(output_ext + 1, "ptx") == 0;
     const bool print_debug = (std::getenv("WARP_DEBUG") != nullptr);
 
-    // check include dir path len (path + option)
-    const int max_path = 4096 + 16;
-    if (strlen(include_dir) > max_path) {
-        fprintf(stderr, "Warp error: Include path too long\n");
-        return size_t(-1);
-    }
-
     if (print_debug) {
         // Not available in all nvJitLink versions
         // unsigned major = 0;
@@ -4458,9 +4451,10 @@ size_t wp_cuda_compile_program(
         printf("NVRTC version %d.%d\n", major, minor);
     }
 
-    char include_opt[max_path];
-    strcpy(include_opt, "--include-path=");
-    strcat(include_opt, include_dir);
+    // Vector to store dynamically created option strings
+    std::vector<std::string> stored_options;
+    stored_options.reserve(static_cast<size_t>(std::max(num_cuda_include_dirs, 0)) + 3);
+    stored_options.push_back(std::string("--include-path=") + include_dir);
 
     const int max_arch = 128;
     char arch_opt[max_arch];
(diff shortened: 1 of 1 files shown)


shi-eric:
Thanks for the PR. Before we consider this further, can you clarify what real circumstance led you to this change?

Was this based on an actual Warp failure you encountered, or was it found by static analysis / code inspection? If it was an observed failure, please share the platform, install path layout, CUDA environment, reproduction steps, and validation showing that this change fixes it.

For non-trivial changes, especially in CUDA runtime internals, we want contributions to be driven by demonstrated user issues, production failures, or clearly validated feature work. We generally do not want to establish a pattern of accepting speculative hardening PRs for hypothetical edge cases, even when the patch itself is small.

If this is only a theoretical issue involving an unusually long Warp install path, we are unlikely to prioritize it. If there is a concrete real-world case behind it, please add those details so we can evaluate the impact.

shi-eric:
Closing this PR for now. The underlying concern is plausible, but we do not have
a reported Warp failure or validation that exercises the CUDA/NVRTC path.

Changes to native CUDA runtime code require maintainer time to reproduce, review,
and test across supported configurations. We prioritize demonstrated bugs that
affect users or block real Warp workflows over speculative edge cases. Given the
narrow path-length conditions and the absence of observed user impact, we do not
plan to take this change.

If this failure is encountered in actual use, please open an issue with the
affected environment, installation path, reproduction steps, and CUDA/NVRTC
validation so we can reassess it.

PR #1590, opened 2026-06-26
Fix kernel hook cache leak for closure-created kernels

### LINKED ISSUE #1589: `@wp.kernel` factory leaks one `Adjoint` per launch (default-module path)
### Description

Repeatedly launching an identical kernel created by a factory/closure (a `@wp.kernel` defined inside a function and re-created each call) grows memory without bound on the default-module path.

The live-kernel `WeakSet` correctly drops the re-created `Kernel` objects, so `Module` kernel/module counts stay stable. But `ModuleExec.kernel_hooks` is keyed by `kernel.adj`, and each re-created kernel has a fresh `Adjoint`. The hooks dict holds those `Adjoint` objects as strong keys, so one `Adjoint`/AST/`Var` graph is retained per launch — a linear leak that's easy to miss because counts look flat.

This is the pattern used by kernel factories that specialize on a static loop count / dtypes, e.g.:

```python
def get_kernel(max_bones_count, dtype):
    @wp.kernel
    def k(...):
        for i in range(wp.static(max_bones_count)):
            ...
    return k
```

### Repro

```python
import warp as wp

wp.init()

def make_kernel():
    @wp.kernel
    def k(a: wp.array(dtype=float), b: wp.array(dtype=float)):
        i = wp.tid()
        b[i] = a[i] + 1.0
    return k

a = wp.zeros(8, dtype=float)
b = wp.zeros(8, dtype=float)

k0 = make_kernel()
wp.launch(k0, dim=8, inputs=[a], outputs=[b])
module = k0.module
del k0

for i in range(1, 2001):
    k = make_kernel()                 # identical kernel, recreated each iteration
    wp.launch(k, dim=8, inputs=[a], outputs=[b])
    del k
    if i % 500 == 0:
        hooks = sum(len(e.kernel_hooks) for e in module.execs.values())
        print(f"after {i:5d} launches: live_kernels={len(module._get_live_kernels())}, kernel_hooks={hooks}")
```

Output (reproduces on `main`, also `1.15.0.dev20260612`):

```
after   500 launches: live_kernels=1, kernel_hooks=501
after  1000 launches: live_kernels=1, kernel_hooks=1001
after  1500 launches: live_kernels=1, kernel_hooks=1501
after  2000 launches: live_kernels=1, kernel_hooks=2001
```

`live_kernels` stays at 1, but `kernel_hooks` (and the `Adjoint` graphs it pins) grows one per launch. Device-agnostic — the cache keying is shared by the CPU and CUDA paths.

### Expected

Relaunching an identical recreated kernel should not grow `kernel_hooks`; the entry-point cache should collapse to one entry per distinct compiled kernel.

### Workaround

`@wp.kernel(module="unique")` already dedups identical factory kernels (returns the cached kernel object), so it doesn't exhibit the growth.

adenzler-nvidia:
## Description

Closes #1589. `ModuleExec.get_kernel_hooks` keyed its entry-point cache by `kernel.adj`. The live-kernel `WeakSet` drops a re-created `Kernel` object, but its adjoint stays pinned as a dict key, so a kernel factory that rebuilds an identical `@wp.kernel` each step retains one `Adjoint`/AST/`Var` graph per launch — unbounded memory growth even though kernel and module counts stay flat. This keys the cache by the mangled name (the compiled-symbol identity) instead. The kernel hash folds in the adjoint and all per-kernel options, so within a `ModuleExec` the resolved hooks are fully determined by the name; identical specializations collapse to a single entry and no adjoint is retained.

## Changes

- `ModuleExec.get_kernel_hooks` keys `kernel_hooks` by `kernel.get_mangled_name()` rather than `kernel.adj`.
- Add `test_create_kernel_loop_hooks_bounded` to `test_codegen_instancing.py`, asserting the per-`ModuleExec` hook count does not grow when an identical kernel is recreated and relaunched.

## Checklist

## Validation summary

The new test recreates an identical factory kernel in a loop and asserts the hook cache does not grow past its baseline. Verified red/green with the equivalent keying on `1.15.0.dev20260612`: with the current `kernel.adj` keying the hook count grows linearly with launches (the leak), and with name-keying it stays flat at the baseline. The change is device-agnostic — the cache keying is shared by the CPU and CUDA paths, so the test registers once with `devices=None`. Full native test-suite run is left to CI.

## Bug fix

```python
import warp as wp

wp.init()

def make_kernel():
    @wp.kernel
    def k(a: wp.array(dtype=float), b: wp.array(dtype=float)):
        i = wp.tid()
        b[i] = a[i] + 1.0
    return k

a = wp.zeros(8, dtype=float)
b = wp.zeros(8, dtype=float)

k0 = make_kernel()
wp.launch(k0, dim=8, inputs=[a], outputs=[b])
module = k0.module
del k0

for i in range(1, 2001):
    k = make_kernel()
    wp.launch(k, dim=8, inputs=[a], outputs=[b])
    del k
    if i % 500 == 0:
        hooks = sum(len(e.kernel_hooks) for e in module.execs.values())
        print(f"after {i} launches: live_kernels={len(module._get_live_kernels())}, kernel_hooks={hooks}")
```

Without this PR, `kernel_hooks` grows one entry per launch while `live_kernels` stays at 1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed unbounded memory growth when repeatedly launching the same kernel created from a factory or closure.
  - Improved kernel hook handling so repeated launches reuse existing cached hooks more reliably.

- **Tests**
  - Added a regression test to verify kernel hook usage stays bounded across repeated kernel creation and execution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,8 @@
 
 ### Fixed
 
+- Fix unbounded memory growth when repeatedly launching identical kernels created by a factory or closure
+  ([GH-1589](https://github.com/NVIDIA/warp/issues/1589)).
 - Fix `wp.load_module()` and `wp.ScopedCapture(force_module_load=True)` after CPU launches so CUDA graph capture
   precompiles the correct CUDA kernel variant. On CUDA drivers older than 12.3 this previously raised
   `CUDA_ERROR_STREAM_CAPTURE_UNSUPPORTED`; on newer drivers it silently recompiled inside the capture window
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -2977,14 +2977,13 @@ class ModuleExec:
 
     # lookup and cache kernel entry points
     def get_kernel_hooks(self, kernel) -> KernelHooks:
-        # Use kernel.adj as a unique key for cache lookups instead of the kernel itself.
-        # This avoids holding a reference to the kernel and is faster than using
-        # a WeakKeyDictionary with kernels as keys.
-        hooks = self.kernel_hooks.get(kernel.adj)
+        # Key by the mangled name (compiled-symbol identity), not kernel.adj, which pinned one
+        # Adjoint per closure-recreated kernel -- a leak the live-kernel WeakSet can't reclaim.
+        name = kernel.get_mangled_name()
+
+        hooks = self.kernel_hooks.get(name)
         if hooks is not None:
             return hooks
-
-        name = kernel.get_mangled_name()
 
         options = kernel.module.options | kernel.options
 
@@ -3082,7 +3081,7 @@ class ModuleExec:
 
             hooks = KernelHooks(forward, backward)
 
-        self.kernel_hooks[kernel.adj] = hooks
+        self.kernel_hooks[name] = hooks
         return hooks
 
 
--- a/warp/tests/test_codegen_instancing.py
+++ b/warp/tests/test_codegen_instancing.py
@@ -1308,6 +1308,46 @@ def test_garbage_collection(test, device):
 
 # =======================================================================
 
+
+def test_create_kernel_loop_hooks_bounded(test, device):
+    """
+    Relaunching an identical recreated kernel must not grow the per-ModuleExec hook cache
+    (regression for an Adjoint leak when hooks were keyed by kernel.adj).
+    """
+
+    def make():
+        @wp.kernel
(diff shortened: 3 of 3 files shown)


adenzler-nvidia:
Closing in favor of an MR on the internal GitLab Warp repo. The bug stays tracked in #1589.

PR #1599, opened 2026-06-29
optim: don't reallocate Adam fp16 moment buffers on every set_params

### LINKED ISSUE #1593: Adam.set_params re-zeros fp16 optimizer moment state on every call
### Bug

`warp.optim.Adam.set_params()` silently reallocates (zeros) its moment buffers `m`/`v` on every call when the parameters are `float16`, discarding any accumulated optimizer state. For `float32`/`vec3` it correctly reuses the existing buffers, so the reset only affects `fp16` params.

### Cause

Moment buffers are intentionally kept in `float32` even when the params are `fp16` ([`adam.py:128`](https://github.com/NVIDIA/warp/blob/main/warp/_src/optim/adam.py#L128)):

```python
elif param.dtype == wp.float16:
    dtype = wp.float32  # we always use fp32 for moments, even if params are fp16
```

But the reuse/realloc guard compares the existing buffer dtype against the **param** dtype rather than the moment-buffer dtype ([`adam.py:132,134`](https://github.com/NVIDIA/warp/blob/main/warp/_src/optim/adam.py#L132-L135)):

```python
if self.m[i] is None or self.m[i].shape != param.shape or self.m[i].dtype != param.dtype:
    self.m[i] = wp.zeros(shape=param.shape, dtype=dtype, device=param.device)
```

For an fp16 param this condition is `float32 != float16` → always `True`, so the moment buffer is reallocated to zeros on every `set_params` call. `SGD.set_params` ([`sgd.py:83`](https://github.com/NVIDIA/warp/blob/main/warp/_src/optim/sgd.py#L83)) does not hit this because its buffer dtype matches the param dtype (`wp.zeros_like(param)`).

### Repro (reproduces on both CPU and CUDA)

```python
import warp as wp
import warp.optim
wp.init()

p = wp.zeros(8, dtype=wp.float16, requires_grad=True)
opt = warp.optim.Adam([p], lr=1e-3)
opt.m[0].fill_(1.0)          # pretend we have accumulated moment state
opt.set_params([p])          # same params -> should be a no-op
print((opt.m[0].numpy() == 1.0).all())   # -> False (state wiped); with fp32 params -> True
```

### Suggested fix

Compare against the moment-buffer dtype (`dtype`) instead of `param.dtype` on the two guard lines. This leaves `float32`/`vec3` behavior unchanged (there `dtype == param.dtype`) and stops the spurious reset for `fp16`.

Happy to open a PR — I have a one-line fix plus a regression test covering `fp32`/`fp16`/`vec3`.

EylonKrause:
## Description

`warp.optim.Adam` keeps its first/second moment buffers in fp32 even when the optimized parameters are fp16 (`set_params` forces `dtype = wp.float32` for an fp16 param). But the reallocation guard compares the *existing* moment's dtype against `param.dtype`:

```python
if self.m[i] is None or self.m[i].shape != param.shape or self.m[i].dtype != param.dtype:
    self.m[i] = wp.zeros(shape=param.shape, dtype=dtype, device=param.device)
```

For an fp16 param this is `float32 != float16`, which is always true, so the moment buffers are reallocated — and therefore zeroed — on **every** `set_params` call. Any code that re-sets the params on a persistent optimizer (e.g. to swap in new tensors between iterations) silently wipes Adam's accumulated `m`/`v` state, breaking fp16 convergence. Closes #1593.

## Changes

- `warp/_src/optim/adam.py`: compare the existing moment dtype against the intended moment `dtype` (fp32 for fp16 params) rather than `param.dtype`. For fp32 / vec3 params `dtype == param.dtype`, so their behavior is unchanged; for fp16 params the moments are now preserved across repeated `set_params` calls.
- `warp/tests/test_adam.py`: add `test_adam_fp16_moments_preserved_on_set_params`, which writes a sentinel into the moments and asserts a second `set_params` neither reallocates nor zeros them.

## Testing

- `python -m py_compile` clean on both files.
- The change is a self-contained control-flow fix; the added regression test exercises it on whatever devices CI runs. (I develop on a single-GPU WSL2 box without a full Warp native build, so I could not run the suite locally — flagging that for the reviewer.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved optimizer parameter refresh behavior so moment buffers keep the correct precision and are only recreated when needed.
  * Fixed an issue where moment buffers for half-precision parameters could be unnecessarily reallocated or reset.

* **Tests**
  * Added a regression test covering half-precision parameters to ensure optimizer moment buffers remain stable across parameter updates.

### DIFF
--- a/warp/_src/optim/adam.py
+++ b/warp/_src/optim/adam.py
@@ -129,9 +129,9 @@ class Adam:
                 else:
                     raise RuntimeError(f"Unsupported dtype for Warp Adam optimizer: {param.dtype}")
 
-                if self.m[i] is None or self.m[i].shape != param.shape or self.m[i].dtype != param.dtype:
+                if self.m[i] is None or self.m[i].shape != param.shape or self.m[i].dtype != dtype:
                     self.m[i] = wp.zeros(shape=param.shape, dtype=dtype, device=param.device)
-                if self.v[i] is None or self.v[i].shape != param.shape or self.v[i].dtype != param.dtype:
+                if self.v[i] is None or self.v[i].shape != param.shape or self.v[i].dtype != dtype:
                     self.v[i] = wp.zeros(shape=param.shape, dtype=dtype, device=param.device)
 
     def reset_internal_state(self):
--- a/warp/tests/test_adam.py
+++ b/warp/tests/test_adam.py
@@ -133,6 +133,31 @@ def test_adam_solve_two_inputs(test, device):
                 test.assertLessEqual(v, tol)
 
 
+def test_adam_fp16_moments_preserved_on_set_params(test, device):
+    # Regression test for #1593: for fp16 params, set_params() reallocated (and
+    # therefore zeroed) the fp32 moment buffers on every call, because the
+    # realloc guard compared the moment dtype against param.dtype (fp32 != fp16)
+    # instead of the intended moment dtype.
+    with wp.ScopedDevice(device):
+        params = wp.array(np.array([0.1, 0.2], dtype=np.float16), dtype=wp.float16, requires_grad=True)
+        opt = warp.optim.Adam([params], lr=0.02)
+
+        m0, v0 = opt.m[0], opt.v[0]
+        # Moments are kept in fp32 even when the params are fp16.
+        test.assertEqual(m0.dtype, wp.float32)
+        # Write a sentinel so a spurious reset is observable.
+        m0.fill_(1.0)
+        v0.fill_(1.0)
+
+        # Re-setting the same params must not reallocate or zero the moments.
+        opt.set_params([params])
+
+        test.assertIs(opt.m[0], m0)
+        test.assertIs(opt.v[0], v0)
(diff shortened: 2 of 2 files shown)


shi-eric:
Hi @EylonKrause, this PR duplicates #1595. Both PRs make the same two-line production change, while #1595 was already open and includes broader regression coverage and the required changelog entry.

Issue #1593 had already been assigned to @Nas01010101 after they coordinated with the team. Our contribution guidelines require contributors to check existing issues and coordinate in the issue thread when work is already assigned.

We’re closing this pull request in favor of #1595 so the implementation and review remain in one place. In the future, please check for assigned issues and existing pull requests before beginning implementation.

EylonKrause:
Thanks @shi-eric, and apologies for the duplicate — you're right I should have checked that #1593 was assigned to @Nas01010101 and that #1595 was already open before sending this. Closing in favor of #1595 makes sense; I'll coordinate in the issue thread before picking up assigned work going forward. Appreciate the clear explanation.

PR #1647, opened 2026-07-11
Fix custom grad with extra shared memory

### LINKED ISSUE #1646: custom @wp.func_grad that needs more shared memory than its forward
### Bug Description

```python
"""A custom @wp.func_grad that needs more shared memory than its forward (here a
tile_matmul vs an elementwise op) makes Warp under-size the backward kernel's shared
memory -> out-of-bounds __shared__ writes on CUDA (error 700; CPU tolerates it). Fix:
fold custom_grad_func.get_total_required_shared() into the kernel roofline in
codegen.py::Adjoint.add_call (and rm -rf ~/.cache/warp/* to drop stale kernels)."""

import numpy as np

import warp as wp

wp.init()
M = wp.constant(32)


@wp.func
def scale2x(x: wp.array2d(dtype=float), y: wp.array2d(dtype=float), i: int):
    # forward: y = 2x elementwise -> tiny shared footprint
    wp.tile_store(y, wp.tile_load(x, shape=(M, M), offset=(i * M, 0)) * 2.0, offset=(i * M, 0))


@wp.func_grad(scale2x)
def adj_scale2x(x: wp.array2d(dtype=float), y: wp.array2d(dtype=float), i: int):
    # backward is dL/dx = 2*dL/dy, but routed through a shared-heavy tile_matmul (the
    # trigger); *0.0 keeps the value at 2 while the matmul still feeds the scatter (no DCE)
    g = wp.tile_load(wp.adjoint[y], shape=(M, M), offset=(i * M, 0))
    x0 = wp.tile_load(x, shape=(M, M), offset=(i * M, 0))
    wp.tile_atomic_add(wp.adjoint[x], g * 2.0 + wp.tile_matmul(g, x0) * 0.0, offset=(i * M, 0))


@wp.kernel(module="unique")
def run(x: wp.array2d(dtype=float), y: wp.array2d(dtype=float)):
    scale2x(x, y, wp.tid())


def main(dev="cuda:0"):
    m, b = int(M), 128
    x = wp.array(np.ones((b, m), np.float32), device=dev, requires_grad=True)
    y = wp.zeros_like(x)
    with wp.Tape() as tape:
        wp.launch_tiled(run, dim=(b // m,), inputs=[x], outputs=[y], block_dim=64, device=dev)
    tape.backward(grads={y: wp.ones_like(y)})
    print("no OOB; x.grad all == 2 ?", np.allclose(x.grad.numpy(), 2.0))  # CUDA 700 here without the fix


if __name__ == "__main__":
    main()
```

### System Information

```bash
Software
  Warp (package):   1.15.0
  Warp (core):      1.15.0
  warp-clang:       1.15.0 (LLVM 18.1.3)
  NumPy:            2.5.1
  Python:           3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]
  Platform:         Linux-6.17.0-35-generic-x86_64-with-glibc2.39

CUDA
  Enabled:          True
  Toolkit:          12.9
  Driver:           13.2
  NVRTC:            12.9
  Forward compat:   True

Libraries
  MathDx:           0.3.1
  cuBQL:            enabled
  NanoVDB:          32.8.0

Build
  Compiler:         GCC 14.2.1
  Debug:            False
  Verify FP:        False
  Fast math:        False
  Sanitizer:        none

CPU
  Model:            arrowlake-s
  ISA features:     61
                      64bit, adx, aes, avx, avx2, avxifma, avxneconvert, avxvnni, ...

Devices
  cpu
    Name:             x86_64
  cuda:0
    Name:             NVIDIA GeForce RTX 5090
    Memory:           31.4 GiB
    Arch:             sm_120
    SMs:              170
    PCI:              00000000:02:00
    Mempool:          enabled
    GPU->CPU mem:     True
    CPU->GPU mem:     False
    CPU/GPU atomics:  False
````

etaoxing:
Fix https://github.com/NVIDIA/warp/issues/1646

## Description

Warp's shared-memory roofline for a kernel only accounted for the shared memory used by each @wp.func's forward body. When a function carries a custom gradient (@wp.func_grad) or custom replay whose backward pass needs more shared memory than the forward (e.g. a tile_matmul in the grad vs. an elementwise op in the forward), the backward kernel's __shared__ allocation was under-sized, causing out-of-bounds shared writes — an illegal memory access (CUDA error 700) on GPU, silently tolerated on CPU. This PR folds the custom grad/replay functions' shared requirements into the roofline so the backward kernel is sized correctly.

## Changes

  - codegen.py (Adjoint.add_call): when computing the shared-memory roofline for a non-builtin call, also build and account for the function's custom_grad_func and custom_replay_func, taking the max of the forward and custom-grad/replay requirements.call, also build and account for the function's custom_grad_func and custom_replay_func, taking the max of the forward and custom-grad/replay requirements.
- context.py (ModuleBuilder.build_function): track in-flight builds in a building_functions set and skip re-entrant builds, breaking the build recursion that occurs when a custom grad calls its own forward during the roofline pass.
- Tests: add test_tile_custom_grad_extra_shared — a custom func_grad whose backward routes through an EXTRA×EXTRA shared scratch tile that dwarfs the M×M forward tile.

## Validation summary

- Added test_tile_custom_grad_extra_shared, which exercises a custom grad whose backward shared footprint exceeds its forward. On the target branch (without this fix) the backward under-allocates shared memory and the launch fails with CUDA error 700 / corrupts x.grad; with the fix the kernel is sized correctly and dL/dx == 2.0 everywhere.
- Verified the reproducer (tile_matmul backward over an elementwise forward) runs clean after the fix, and confirmed the recursion guard in build_function handles a custom grad that re-enters its own forward mid-build.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GPU shared-memory sizing for backward execution, ensuring custom gradient/replay paths that need more shared memory than the forward pass no longer risk illegal memory access.
  * Improved module build behavior to prevent recursive, in-progress builds and to keep build/import ordering consistent.
* **Tests**
  * Added a tiled-kernel test that validates backward correctness when a custom gradient uses a larger shared-memory footprint than the forward pass.


etaoxing:
merged on gitlab

PR #1702, opened 2026-07-27
Fix/hip mempool enable for graph capture

### LINKED ISSUE #1234: [REQ] Extend texture interop capabilities
### Description

Allow textures to be populated from warp arrays or other data sources at runtime without going though host side memory.

### Context

Allows very efficient interop with rendering frameworks like Vulkan etc.

zhihuidu-amd:
## Description

Three bug fixes in warp/_src/types.py and warp/_src/context.py required for correct GPU graph capture on AMD ROCm (HIP) devices. These make the standard upstream graph capture pattern work on AMD MI300X/MI325X (ROCm 7.x) without affecting NVIDIA CUDA behavior.
Tested on AMD MI325X (gfx942), ROCm 7.2 with mujoco_warp + PyTorch-ROCm.
Companion PR: google-deepmind/mujoco_warp#1556

Fix 1 — types.py: hipMalloc(0) returns nullptr on ROCm
Unlike CUDA where cudaMalloc(0) returns a unique non-null pointer, hipMalloc(0) on ROCm returns nullptr. For zero-capacity arrays (e.g. models with no tendons, no welds), downstream code that uses the pointer causes GPU page faults.
python# Before: hipMalloc(0) → nullptr → GPU page fault
ptr = allocator.allocate(capacity)

# After: 1-byte minimum on HIP, behavior unchanged on CUDA
alloc_size = 1 if (device.is_cuda and getattr(device, "is_hip", False) and capacity == 0) else capacity
if alloc_size > 0:
    ptr = allocator.allocate(alloc_size)

Fix 2 — types.py: zero_() raises hipErrorInvalidValue on zero-size arrays
hipMemsetAsync(ptr, 0, 0) returns hipErrorInvalidValue on ROCm 7.x. CUDA silently ignores size=0. Added early return after the docstring to match CUDA no-op behavior:
pythondef zero_(self):
    """Zero out the array entries."""
    if self.size == 0:   # hipMemsetAsync(size=0) errors on ROCm; no-op matches CUDA
        return
    ...

Fix 3 — context.py: Opt-out for mempool conflict with PyTorch ROCm allocator
On AMD ROCm, Warp auto-enabling the HIP memory pool at device init conflicts with PyTorch's ROCm allocator when both run in the same process (e.g. physics simulation + RL policy training on the same GPU). This causes null pointer faults during wp.array() allocation.
New environment variable WARP_DISABLE_MEMPOOLS_FOR_TORCH=1 allows users to opt out:
python# NVIDIA CUDA: env var not set → behavior unchanged (is_mempool_enabled = is_mempool_supported)
# AMD ROCm: set WARP_DISABLE_MEMPOOLS_FOR_TORCH=1 → mempool disabled
if os.environ.get("WARP_DISABLE_MEMPOOLS_FOR_TORCH") == "1":
    self.is_mempool_enabled = False
else:
    self.is_mempool_enabled = self.is_mempool_supported
This approach requires no HIP-specific device attribute detection and preserves NVIDIA CUDA behavior exactly.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://nvidia.github.io/warp/latest/project/contribution_guide.html).

## Bug fix

```python
import warp as wp
# Code that demonstrates the bug
```

## New feature / enhancement

```python
import warp as wp
# Code that demonstrates the new capability
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of zero-capacity arrays on HIP/ROCm to prevent invalid allocation behavior.
  * Made zeroing zero-size arrays on HIP/ROCm a safe no-op (matching CUDA behavior).
  * Updated graph capture to ignore zero-size input/output arrays, avoiding unnecessary nodes and edges.
* **Behavior Changes**
  * CUDA memory-pool auto-enablement on HIP/ROCm is now opt-in only to prevent allocator conflicts.
* **Tests**
  * Adjusted assertions for empty array capture to validate `size == 0`.

### DIFF
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -5550,7 +5550,6 @@ class Device:
                 # enable if supported
                 self.is_mempool_enabled = self.is_mempool_supported
             else:
-                # disable by default
                 self.is_mempool_enabled = False
 
             uuid_buffer = (ctypes.c_char * 16)()
--- a/warp/_src/tape.py
+++ b/warp/_src/tape.py
@@ -1149,7 +1149,7 @@ def visit_tape(
         for id, x in enumerate(launch.inputs):
             name = kernel.adj.args[id].label
             if isinstance(x, wp.array):
-                if x.ptr is None:
+                if x.size == 0:  # skip zero-size arrays (ptr may be non-None on HIP)
                     continue
                 # if x.ptr in array_to_launch and len(array_to_launch[x.ptr]) > 1:
                 #     launch_arg_i = array_to_launch[x.ptr]
@@ -1166,6 +1166,8 @@ def visit_tape(
             elif wp._src.types.is_struct(x):
                 for varname, var in get_struct_vars(x).items():
                     if isinstance(var, wp.array):
+                        if var.size == 0:  # skip zero-size nested arrays (ptr may be non-None on HIP)
+                            continue
                         if not hide_readonly_arrays or var.ptr in computed_nodes or var.ptr in input_output_ptr:
                             add_array_node(var, f"{name}.{varname}", active_scope_stack)
                             input_arrays.append(var.ptr)
@@ -1176,7 +1178,7 @@ def visit_tape(
         output_arrays = []
         for id, x in enumerate(launch.outputs):
             name = kernel.adj.args[id + len(launch.inputs)].label
-            if isinstance(x, wp.array) and x.ptr is not None:
+            if isinstance(x, wp.array) and x.size > 0:  # skip zero-size arrays
                 add_array_node(x, name, active_scope_stack)
                 output_arrays.append(x.ptr)
                 computed_nodes.add(x.ptr)
@@ -1184,6 +1186,8 @@ def visit_tape(
             elif wp._src.types.is_struct(x):
                 for varname, var in get_struct_vars(x).items():
                     if isinstance(var, wp.array):
+                        if var.size == 0:  # skip zero-size nested arrays (ptr may be non-None on HIP)
+                            continue
                         add_array_node(var, f"{name}.{varname}", active_scope_stack)
                         output_arrays.append(var.ptr)
                         computed_nodes.add(var.ptr)
--- a/warp/_src/types.py
+++ b/warp/_src/types.py
@@ -4358,6 +4358,13 @@ class array(Array[DType, NDim]):
 
     def zero_(self):
         """Zero out the array entries."""
+        if self.size == 0:
+            # Skip the zero-byte memset, but still record the initialization:
+            # otherwise verify_autograd_array_access keeps a stale "read" mark
+            # on the empty array and a later guarded write reports a false
+            # write-after-read.
+            self.mark_init()
+            return
         self._apic_ensure_tracked()
         if self.is_contiguous:
             # simple memset is usually faster than generic fill
--- a/warp/config.py
+++ b/warp/config.py
@@ -14,6 +14,7 @@ setting documentation for details.
 For information on module-level and kernel-level settings, see :doc:`/user_guide/configuration`.
 """
 
+import os as _os
 import sys as _sys
 import types as _types
 from enum import IntEnum as _IntEnum
@@ -386,8 +387,13 @@ enable_graph_capture_module_load_by_default: bool = True
 Only affects systems with CUDA driver versions below 12.3.
 """
 
-enable_mempools_at_init: bool = True
-"""Enable CUDA memory pools during device initialization when supported."""
+enable_mempools_at_init: bool = _os.environ.get("WARP_ENABLE_MEMPOOLS_AT_INIT", "1").lower() not in ("0", "false", "no")
+"""Enable CUDA memory pools during device initialization when supported.
+
+Set ``WARP_ENABLE_MEMPOOLS_AT_INIT`` to ``0``, ``false``, or ``no`` before
+importing Warp to disable automatic memory pool initialization (e.g. when
+sharing the process with a framework that owns its own GPU allocator).
+"""
 
 track_memory: bool = False
 """Enable tracking of memory allocations at initialization.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -321,6 +321,7 @@ autosummary_filename_map = {
 }
 
 AUTOSUMMARY_ANNOTATION_OVERRIDES = {
+    "warp.config.enable_mempools_at_init": ": bool = True",
     "warp.config.launch_array_access_mode": (
         ": warp.config.LaunchArrayAccessMode = warp.config.LaunchArrayAccessMode.RELAXED"
     ),


warp/_src/context.py
```diff
@@ -4927,10 +4927,19 @@ def __init__(self, runtime, alias, ordinal=-1, is_primary=False, context=None):
...
-                self.is_mempool_enabled = self.is_mempool_supported
+                # HIP/ROCm: auto-enabling the memory pool conflicts with PyTorch's
+                # ROCm allocator when used in the same process. Apply opt-in only
+                # for HIP devices; CUDA mempool behavior is completely unchanged.
+                # On HIP/ROCm, auto-enabling the memory pool conflicts with PyTorch's
+                # On AMD ROCm (HIP), this auto-enabled pool can conflict with
+                # PyTorch's ROCm allocator when both run in the same process.
+                # Set WARP_DISABLE_MEMPOOLS_FOR_TORCH=1 to opt out.
+                # On NVIDIA CUDA, this variable is not set and behavior is unchanged.
+                if os.environ.get("WARP_DISABLE_MEMPOOLS_FOR_TORCH") == "1":
```
shi-eric: This comment block is much longer than necessary, and several lines are duplicated or stale. More importantly, nothing about `WARP_DISABLE_MEMPOOLS_FOR_TORCH` is specific to PyTorch. When set, it affects all supported GPU devices, including NVIDIA CUDA.

Could we instead add `WARP_ENABLE_MEMPOOLS_AT_INIT` and use it to initialize the existing `warp.config.enable_mempools_at_init` setting? Setting `WARP_ENABLE_MEMPOOLS_AT_INIT=0` before importing Warp would provide the needed opt-out, while `Device.__init__` would continue to depend only on the existing config setting. This also works if Warp initialization ever happens during import.

warp/_src/types.py
```diff
@@ -3735,12 +3735,18 @@ def _init_new(self, dtype, shape, strides, device, pinned):
         # cannot leak a freshly-allocated pointer between allocate() and self.deleter assignment.
         deleter = allocator.deallocate
 
-        if capacity > 0:
+        # HIP/ROCm: hipMalloc(0) returns nullptr (unlike CUDA which returns a unique
+        # non-null pointer). Use 1-byte minimum so ptr is always valid on HIP.
+        # HIP/ROCm: hipMalloc(0) returns nullptr; use 1-byte minimum for zero-capacity arrays.
+        # getattr guards against non-HIP CUDA devices that do not define is_hip.
+        is_hip = device.is_cuda and getattr(device, "is_hip", False)
+        alloc_size = 1 if (is_hip and capacity == 0) else capacity
```
shi-eric: [Warp 1.13.0 added pluggable CUDA allocators](https://nvidia.github.io/warp/v1.13/user_guide/changelog.html#added) through `wp.Allocator` and `wp.set_cuda_allocator()`.

Could the zero-size behavior live in a custom allocator for the HIP fork? That allocator could translate a zero-byte request into a one-byte allocation without putting `Device.is_hip` logic in `array._init_new()`.

`_init_new()` currently skips the allocator when `capacity == 0`, so we may need a small generic hook that lets an allocator opt into receiving zero-size requests. Existing allocators would keep the current `ptr = None` behavior, while the HIP fork would get the customization it needs.

warp/_src/types.py
```diff
@@ -4180,6 +4186,10 @@ def _apic_ensure_tracked(self):
 
     def zero_(self):
         """Zero out the array entries."""
+        # HIP/ROCm: hipMemsetAsync(size=0) returns hipErrorInvalidValue.
+        # Early-return to match CUDA no-op behavior for zero-size arrays.
+        if self.size == 0:
```
shi-eric, line 4361: The early return looks reasonable, but it isn’t HIP-specific: zeroing an empty array should be a no-op on every backend, consistent with `fill_()`. Please remove the HIP/ROCm-specific comment; the code is self-explanatory.

warp/tests/test_apic.py
```diff
@@ -1325,8 +1325,9 @@ def test_capture_with_empty_array_input(test, device):
     out = wp.zeros(n_out, dtype=float, device=device)
     empty = wp.zeros(0, dtype=float, device=device)
 
-    # Sanity: zero-length arrays carry arr.ptr == None (not 0)
-    test.assertTrue(empty.ptr is None, f"expected None, got {empty.ptr!r}")
+    # Sanity: zero-length arrays have size == 0. On HIP, ptr may be non-null
+    # (1-byte allocation to avoid hipMalloc(0)=nullptr GPU page faults).
+    test.assertTrue(empty.size == 0, f"expected size 0, got {empty.size!r}")
```
shi-eric: I don’t think we should change this upstream test. It was added specifically to exercise APIC’s `ptr is None` path; replacing that assertion with `size == 0` removes the regression coverage, since `wp.zeros(0)` already guarantees a size of zero.

Warp upstream does not currently support HIP, and the non-null representation for empty arrays is a policy introduced by the HIP fork, so adapting this assertion belongs in that fork for now. We’re open to a backend-neutral upstream proposal that preserves coverage of both pointer representations, but this change on its own weakens the upstream test.

zhihuidu-amd:
Thank you for the review @shi-eric! Updated per your feedback (commit 67868a5):

**Renamed  -> ** and wired it into the existing  setting in . The HIP device init in  now simply honours  (no separate HIP-specific env var path). Users running alongside PyTorch-ROCm set  before importing Warp to avoid the HIP allocator conflict. Also added docstrings to the modified  methods.

warp/config.py
```diff
@@ -387,6 +387,14 @@ class DeterministicMode(_IntEnum):
...
 
 enable_mempools_at_init: bool = True
+
+# Allow opt-out via environment variable before importing Warp.
+# Set WARP_ENABLE_MEMPOOLS_AT_INIT=0 to disable memory pool initialization
+# (e.g. when sharing the process with a framework that owns its own GPU allocator).
+import os as _os
+if _os.environ.get("WARP_ENABLE_MEMPOOLS_AT_INIT", "").lower() in ("0", "false", "no"):
+    enable_mempools_at_init = False
+del _os
```
shi-eric: The config-based environment opt-out looks like the right approach. Beyond moving `os` to the top-level import block, can you derive the value directly in the assignment?

```python
import os as _os

# ...

enable_mempools_at_init: bool = _os.environ.get(
    "WARP_ENABLE_MEMPOOLS_AT_INIT", "1"
).lower() not in ("0", "false", "no")
"""Enable CUDA memory pools during device initialization when supported.

Set ``WARP_ENABLE_MEMPOOLS_AT_INIT`` to ``0``, ``false``, or ``no`` before
importing Warp to disable automatic memory pool initialization.
"""
```

`del _os` does not provide much value because the underscored alias is already private. More importantly, the current `if` and `del` separate the assignment from its attribute docstring, so Sphinx produces an empty description for this setting. Keeping the assignment and string literal adjacent fixes that without requiring custom docstring handling.

There is one Sphinx-specific detail: autodoc renders the imported runtime value. If the documentation process has `WARP_ENABLE_MEMPOOLS_AT_INIT=0`, the API page reports `False` as the default. Can you also add the following existing-style override to `AUTOSUMMARY_ANNOTATION_OVERRIDES` in `docs/conf.py`?

```python
"warp.config.enable_mempools_at_init": ": bool = True",
```

This keeps the documented default stable while the attribute docstring explains the environment override.

shi-eric:
Thanks @zhihuidu-amd just a few more things before we can get it merged:

- Fix the pre-commit failure
- Squash commits on this pull request to a single one
- Address the failing DCO check

Thanks!

zhihuidu-amd:
Thanks @shi-eric — updates on your list:

- **DCO** — passing.
- **pre-commit** — fixed. The `ruff format` failure was `warp/config.py`: the `enable_mempools_at_init` assignment was hand-wrapped across three lines, but the joined expression is exactly 120 chars, which is within the configured `line-length = 120`, so ruff collapses it. Formatting only, no behaviour change (verified identical for the env var unset, `1`, `0`, `false`, `FALSE`, `no`, `yes`). All four checks are green now.
- **Squash** — this currently sits at 2 commits, since I added the format fix on top rather than force-pushing over the reviewed commit. Happy to squash back to one whenever you'd like — just say the word and I'll do it, or feel free to use "Squash and merge".

Let me know if anything else is needed.

warp/_src/types.py
```diff
@@ -4175,6 +4180,8 @@ def _apic_ensure_tracked(self):
 
     def zero_(self):
         """Zero out the array entries."""
+        if self.size == 0:
+            return
```
shi-eric, line 4367: When `wp.config.verify_autograd_array_access` is enabled, this returns before `mark_init()`, so `zero_()` leaves an empty array marked as read. This reproduces by by passing an empty array to a kernel where the read is guarded by if `a.shape[0] > 0`. No runtime read occurs, but the access verifier marks the argument as read. After `zero_()`, a similarly guarded write produces a false write-after-read warning.

Warp emits a false write-after-read warning even though the empty array is never accessed. `mark_init()` should be called before returning so the zero-byte memset is skipped without leaving stale access state.

warp/_src/tape.py
shi-eric, line 1188: This struct-output path is missing the zero-size check used for direct outputs above and struct inputs. For a struct containing an empty array: `Tape.visualize()` emits a spurious out.values node and  edge, while the equivalent direct empty output is skipped.

zhihuidu-amd:
Thanks for the detailed review @shi-eric — both items are fixed in f626f2d.

**`warp/_src/types.py`** — `mark_init()` is now called before the early return for `size == 0`, so the zero-byte memset is still skipped but the array no longer keeps a stale read flag:

```python
if self.size == 0:
    # Skip the zero-byte memset, but still record the initialization:
    # otherwise verify_autograd_array_access keeps a stale "read" mark
    # on the empty array and a later guarded write reports a false
    # write-after-read.
    self.mark_init()
    return
```

**`warp/_src/tape.py`** — you're right, and that was an inconsistency in my own change: struct *inputs* had the zero-size guard but struct *outputs* did not. Added using the same form as the existing checks, so an empty array nested in an output struct is now skipped like the equivalent direct output.

**`warp/tests/test_apic.py`** — agreed, and this one is already resolved: that file is no longer part of this PR. The branch was rescoped after that earlier review, so the upstream APIC test and its `ptr is None` coverage are untouched. The current diff is five files: `docs/conf.py`, `warp/_src/context.py`, `warp/_src/tape.py`, `warp/_src/types.py`, `warp/config.py`.

All four checks are green on f626f2d. One caveat worth stating: NVIDIA CI has not run on this branch yet (copy-pr-bot vetting), so these two paths have not been exercised by your test suite — I verified them by reading the code and reasoning about the flow rather than by executing the autograd verifier and `Tape.visualize()`. If you can trigger CI, that would be the real confirmation.

shi-eric:
Great. Now please squash changes to a single commit and rebase with main.

zhihuidu-amd:
Done — squashed to a single commit and rebased onto current `main` (53ca1b76f).

The rebase applied cleanly with no conflicts, and the net diff is unchanged: 5 files, +22/-5. The commit message now covers all of the changes together, including the `mark_init()` and struct-output fixes from your review above.

All four checks are green.

One thing worth flagging: NVIDIA CI still has not run on this branch (copy-pr-bot vetting), so the `zero_()` and `Tape.visualize()` paths have not been exercised by your test suite — I verified them by reading the code rather than by running the autograd verifier. If you are able to trigger CI, that would be the real confirmation before merge.

PR #1710, opened 2026-07-29
Fix parallel CPU module loading (GH-1705)

### LINKED ISSUE #1705: Concurrent CPU module loads can fail kernel lookup
## Bug description

When four CPU modules are loaded concurrently in a fresh process, `wp.force_load(..., max_workers=4)` sometimes returns normally even though a kernel from one of those modules cannot be launched afterward. The load is reported as successful, but the later kernel lookup fails.

### Expected behavior

If `wp.force_load()` returns without an error, kernels from every loaded module should be available for launch.

### Actual behavior

The failure is observed intermittently, and a different module may be reported as missing in each run. When the failure occurs, the following output is printed:

```text
Failed to find module: wp_<module>_1
RuntimeError: Failed to find forward kernel 'transform' from module '<module>' for device 'cpu'
```

Success is reported by all four module-load messages before the lookup fails.

## Reproduction

The following script can be saved as `repro.py`:

```python
import importlib.util
import sys
import tempfile
import uuid
from pathlib import Path

import warp as wp


MODULE_SOURCE = """
import warp as wp


@wp.kernel
def transform(values: wp.array[wp.int32]):
    tid = wp.tid()
    values[tid] = values[tid] * 2 + 1
"""


def import_module(path: Path, name: str):
    spec = importlib.util.spec_from_file_location(name, path)
    if spec is None or spec.loader is None:
        raise RuntimeError(f"Could not create an import spec for {path}")
    module = importlib.util.module_from_spec(spec)
    sys.modules[name] = module
    spec.loader.exec_module(module)
    return module


def main() -> None:
    with tempfile.TemporaryDirectory() as directory:
        root = Path(directory)
        modules = []
        for index in range(4):
            name = f"_parallel_load_{index}_{uuid.uuid4().hex}"
            path = root / f"{name}.py"
            path.write_text(MODULE_SOURCE)
            modules.append(import_module(path, name))

        wp.force_load(
            device="cpu",
            modules=[wp.get_module(module.__name__) for module in modules],
            block_dim=1,
            max_workers=4,
        )

        values = wp.array([0, 1, 2, 3], dtype=wp.int32, device="cpu")
        for module in modules:
            wp.launch(
                module.transform,
                dim=values.shape,
                inputs=[values],
                device="cpu",
                block_dim=1,
            )

        actual = values.numpy().tolist()
        expected = [15, 31, 47, 63]
        if actual != expected:
            raise RuntimeError(f"Expected {expected}, got {actual}")
        print(actual)


if __name__ == "__main__":
    main()
```

It can be run repeatedly in fresh processes and caches:

```bash
for attempt in $(seq 1 20); do
  cache_dir=$(mktemp -d)
  WARP_CACHE_PATH="$cache_dir" python repro.py || break
done
```

On the system described below, a failure was produced in 1 of 12 fresh processes.

## Other runs

The same pattern was also tested with a cold-start harness that uses four fixed modules:

- fresh parallel CPU load: 3 failures in 12 processes;
- fresh serial CPU load: 0 failures in 4 processes; and
- serially load one CPU module to initialize the JIT, then load the remaining modules in parallel: 0 failures in 12 processes.

Four more failures were captured by the retained study: three parallel CPU cold samples and one parallel CPU warm sample. The same missing-module and missing-forward-kernel messages were produced in each failure.

Different modules were reported as missing in different runs. No failure was observed in the serial runs or in the runs where one CPU module was loaded before the remaining parallel loads were started.

## Temporary workaround

Until the issue is resolved, `max_workers <= 1` should be used for module loads that can target the CPU. This includes `device="cpu"`, mixed device lists, and `device=None`.

The default value of `wp.config.load_module_max_workers` is `0`, under which modules are loaded serially. If it is set to `None`, up to four workers may be selected and the failure may be triggered for CPU loads. The problem was not observed in CUDA-only runs.

## System information

```text
Warp package/core/warp-clang: 1.17.0.dev0
Local source revision: 35cbcee4
Relevant runtime/native files: unmodified from the main branch at test time
Python: 3.12.13
Platform: Linux 6.8.0-1051-gke x86_64, glibc 2.35
warp-clang LLVM: 18.1.3
Build compiler: GCC 11.4.0
Build type: Release; no sanitizer
```

arko92:
## Description

Fix an intermittent race during parallel CPU module loading. Concurrent first-time calls to `wp_load_obj()` could create separate LLVM JIT instances and overwrite the process-global JIT pointer. A module load could therefore report success while its kernels remained unavailable during lookup.

Closes #1705.

## Changes

- Serialize CPU JIT initialization and module-registry operations. Specifically, added a mutex around CPU JIT module loading, unloading, and symbol lookup. The mutex allows only one thread at a time to access the shared LLVM JIT instance and its module registry, preventing concurrent cold-start loads from creating competing JIT instances or making previously loaded modules unreachable.
- Preserve parallel source generation and native compilation.
- Add a cold-process regression test that launches kernels from every parallel-loaded CPU module.
- Document the fix in the unreleased changelog.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://nvidia.github.io/warp/latest/project/contribution_guide.html).

## Validation summary

- Reproduced the original issue in 6 of 20 fresh-process runs. Each failure produced the reported missing-module and missing-forward-kernel errors.
- Verified that the new cold-start regression test fails without the native mutex.
- Verified that the regression test passes after rebuilding with the fix.
- Ran the complete parallel module-loading test module. All CPU tests passed; two CUDA-only tests were skipped because the local build was CPU-only.
- Ran the full test suite. All executed tests passed, with only the expected CUDA-dependent tests skipped in the CPU-only environment.
- Verified that CPU debug and release builds succeed.
- Verified the legacy RTDyld CPU-linker test with the synchronization in place.

## Bug fix

The issue can be reproduced intermittently by running this example repeatedly in fresh processes:

```python
import warp as wp


def make_kernel():
    @wp.kernel(module="unique")
    def transform(values: wp.array[wp.int32]):
        tid = wp.tid()
        values[tid] = values[tid] * 2 + 1

    return transform


kernels = [make_kernel() for _ in range(4)]

wp.force_load(
    device="cpu",
    modules=[kernel.module for kernel in kernels],
    block_dim=1,
    max_workers=4,
)

values = wp.array([0, 1, 2, 3], dtype=wp.int32, device="cpu")

for kernel in kernels:
    wp.launch(
        kernel,
        dim=values.shape,
        inputs=[values],
        device="cpu",
        block_dim=1,
    )

assert values.numpy().tolist() == [15, 31, 47, 63]
```

Without the fix, a launch may fail even though every module load reported success:

```text
Failed to find module: wp_<module>_1
RuntimeError: Failed to find forward kernel 'transform'
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed intermittent missing CPU modules when loading in parallel with `wp.force_load(max_workers > 1)`.
  * Improved reliability and consistency during concurrent module loading, unloading, and lookup operations.
* **Tests**
  * Added regression coverage for parallel CPU module loading during a cold start, including validation in an isolated process and synchronized loading scenarios.

### DIFF
--- a/warp/tests/test_module_parallel_load.py
+++ b/warp/tests/test_module_parallel_load.py
@@ -6,12 +6,19 @@ wp.force_load() and wp.load_module().
 """
 
 import os
+import subprocess
+import sys
 import tempfile
+import threading
 import unittest
 import uuid
 from importlib import util
+from unittest import mock
+
+import numpy as np
 
 import warp as wp
+import warp._src.context as context
 
 # Chain of shared ``@wp.func`` helpers used by
 # ``TestParallelLoadSharedHelper`` below. Multiple helpers calling each
@@ -113,6 +120,62 @@ def _assert_modules_loaded_on_cpu(test, modules):
         test.assertTrue(has_cpu_exec, f"Module {m.name} was not loaded on CPU")
 
 
+def _make_parallel_cpu_kernel():
+    """Create a kernel in its own module for the parallel CPU load regression."""
+
+    module_name = f"parallel_cpu_load_{uuid.uuid4().hex}"
+
+    @wp.kernel(module=module_name)
+    def transform(values: wp.array[wp.int32]):
+        tid = wp.tid()
+        values[tid] = values[tid] * 2 + 1
+
+    return transform
+
+
+def _run_parallel_cpu_cold_start():
+    """Load and launch four CPU modules concurrently in a fresh subprocess."""
+    wp.init()
+
+    kernels = [_make_parallel_cpu_kernel() for _ in range(4)]
+    modules = [kernel.module for kernel in kernels]
+
+    # This is a fresh process, so none of the calls below should find an
+    # existing CPU JIT instance. Hold each worker at the native loading
+    # boundary until all four are ready, maximizing concurrent first entry
+    # into wp_load_obj() without changing production code.
+    load_barrier = threading.Barrier(len(modules))
+    original_load_obj = context.runtime.llvm.wp_load_obj
+
+    def synchronized_load_obj(*args):
+        load_barrier.wait(timeout=120)
+        return original_load_obj(*args)
+
+    with mock.patch.object(
+        context.runtime.llvm,
+        "wp_load_obj",
+        side_effect=synchronized_load_obj,
+    ):
+        wp.force_load(device="cpu", modules=modules, block_dim=1, max_workers=len(modules))
+
+    # Checking Module.execs is insufficient for this regression: the Python
+    # bookkeeping can contain a module whose native JIT instance is no longer
+    # reachable. Launch every kernel to force native lookup.
+    values = wp.array([0, 1, 2, 3], dtype=wp.int32, device="cpu")
+    for kernel in kernels:
+        wp.launch(
+            kernel,
+            dim=values.shape,
+            inputs=[values],
+            device="cpu",
+            block_dim=1,
+        )
+
+    actual = values.numpy()
+    expected = [15, 31, 47, 63]
+    np.testing.assert_array_equal(actual, expected)
+
+
 class TestModuleParallelLoad(unittest.TestCase):
     def test_force_load_serial(self):
         """Verify that serial compilation (max_workers=0) loads modules correctly."""
@@ -157,6 +220,39 @@ class TestModuleParallelLoad(unittest.TestCase):
         modules = _generate_modules(1)
         wp.force_load(device="cpu", modules=modules, max_workers=2)
         _assert_modules_loaded_on_cpu(self, modules)
+
+    def test_force_load_parallel_cpu_cold_start_kernel_lookup(self):
+        """Verify that modules loaded in parallel remain launchable on the CPU.
+
+        Run in a subprocess because the CPU JIT is process-global and the
+        initialization race can only occur on the first native module loads.
+        Give the subprocess a fresh filesystem cache without clearing Warp's
+        shared test cache, which is unsafe under the parallel test runner.
+        """
+        with tempfile.TemporaryDirectory() as cache_path:
+            env = os.environ.copy()
+            env["WARP_CACHE_PATH"] = cache_path
+            result = subprocess.run(
+                [
+                    sys.executable,
+                    "-c",
+                    (
+                        "from warp.tests.test_module_parallel_load import _run_parallel_cpu_cold_start; "
+                        "_run_parallel_cpu_cold_start()"
+                    ),
+                ],
+                check=False,
+                capture_output=True,
+                text=True,
+                env=env,
+                timeout=180,
+            )
+
+        self.assertEqual(
+            result.returncode,
+            0,
+            f"Parallel CPU cold-start subprocess failed.\nstdout:\n{result.stdout}\nstderr:\n{result.stderr}",
+        )
 
 
 def _assert_modules_loaded_on_cuda(test, modules, device):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,8 @@
   instead of node coordinates in Warp FEM kernels ([GH-1685](https://github.com/NVIDIA/warp/issues/1685)).
 - Reject malformed APIC `.wrp` memory sections containing duplicate region IDs
   or out-of-bounds initial data during graph loading.
+- Fix intermittent missing CPU modules after parallel loading with `wp.force_load(max_workers > 1)`
+  ([GH-1705](https://github.com/NVIDIA/warp/issues/1705)).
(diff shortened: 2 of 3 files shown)


warp/tests/test_module_parallel_load.py
```diff
@@ -113,6 +117,62 @@ def _assert_modules_loaded_on_cpu(test, modules):
...
+            kernel,
+            dim=values.shape,
+            inputs=[values],
+            device="cpu",
+            block_dim=1,
+        )
+
+    actual = values.numpy().tolist()
+    expected = [15, 31, 47, 63]
+    if actual != expected:
```
shi-eric: Please replace the manual list comparison and `RuntimeError` with `np.testing.assert_array_equal()`. These are exact integer results, and the NumPy assertion will provide more useful element-level diagnostics when the test fails. Please also add `import numpy as np`.

warp/tests/test_module_parallel_load.py
```diff
@@ -158,6 +218,39 @@ def test_force_load_single_module(self):
         wp.force_load(device="cpu", modules=modules, max_workers=2)
         _assert_modules_loaded_on_cpu(self, modules)
 
+    def test_force_load_parallel_cpu_cold_start_kernel_lookup(self):
+        """All modules reported as loaded in parallel must remain launchable on CPU.
```
shi-eric: Please rewrite the docstring summary in imperative mood and state what the test verifies. For example: `Verify that modules loaded in parallel remain launchable on the CPU.`

warp/tests/test_module_parallel_load.py
```diff
@@ -158,6 +218,39 @@ def test_force_load_single_module(self):
...
+
+        Run in a subprocess because the CPU LLJIT is process-global and the
```
shi-eric: Please use the established Warp term "CPU JIT" instead of the LLVM implementation type `LLJIT`. Update both this docstring and the earlier helper comment for consistency.

warp/tests/test_module_parallel_load.py
```diff
@@ -113,6 +117,62 @@ def _assert_modules_loaded_on_cpu(test, modules):
...
+    # boundary until all four are ready, maximizing concurrent first entry
+    # into wp_load_obj() without changing production code.
+    load_barrier = threading.Barrier(len(modules))
+    original_load_obj = context.runtime.llvm.wp_load_obj
+
+    def synchronized_load_obj(*args):
+        load_barrier.wait(timeout=120)
+        return original_load_obj(*args)
+
+    context.runtime.llvm.wp_load_obj = synchronized_load_obj
```
shi-eric: Please use `mock.patch.object()` as a context manager instead of assigning and restoring the native binding manually. This makes the monkeypatch scope explicit and handles restoration automatically. Please also add `from unittest import mock`.

```python
with mock.patch.object(
    context.runtime.llvm,
    "wp_load_obj",
    side_effect=synchronized_load_obj,
):
    wp.force_load(
        device="cpu",
        modules=modules,
        block_dim=1,
        max_workers=len(modules),
    )
```

shi-eric:
Please keep this PR squashed to a single commit and rebase it onto the latest `main`. We need to merge the change into our internal GitLab repository first, so having one up-to-date commit makes that integration straightforward. Thanks!

warp/tests/test_module_parallel_load.py
arko92: Updated to use np.testing.assert_array_equal() and added the NumPy import in commit d87782b

warp/tests/test_module_parallel_load.py
arko92: Updated the docstring to use imperative mood and clearly state what the test verifies in commit d87782b

warp/tests/test_module_parallel_load.py
arko92: Replaced “LLJIT” with “CPU JIT” in the helper comment and test docstring in commit d87782b

warp/tests/test_module_parallel_load.py
arko92: Replaced the manual binding assignment and restoration with mock.patch.object() as a context manager in commit d87782b

arko92:
> Please keep this PR squashed to a single commit and rebase it onto the latest `main`. We need to merge the change into our internal GitLab repository first, so having one up-to-date commit makes that integration straightforward. Thanks!

Done—I've squashed the changes into a single DCO-signed commit and rebased it onto the latest main. Thank you for the comments!

shi-eric:
Thanks @arko92, this looks good to me. I will get it merged shortly.

shi-eric (approved):

PR #1733, opened 2026-08-05
Stabilize callable hashes

### LINKED ISSUE #1696: Salted hash() in create_warp_function makes kernel hashing nondeterministic, defeating and eventually poisoning the kernel cache
## Bug Description

`warp._src.utils.create_warp_function()` names lambda-generated functions using Python's builtin `hash()`:

```python
def unique_name(code: str):
    return "func_" + hex(hash(code))[-8:]
```

`hash()` on strings is salted per process (PYTHONHASHSEED), and this name becomes the function key and is embedded in the generated source. It therefore feeds `ModuleHasher`, so **any kernel referencing a lambda-generated function gets a different kernel/module hash in every process**, with two consequences:

**1. The kernel cache never hits across processes.** Identical code recompiles on every run, and a shared/persistent cache directory grows without bound — one new module directory per process per closure kernel. In newton-physics CI this grew the shared kernel cache to 2.3 GB, nearly all of it one-off salted duplicates of the same few kernels.

**2. Rare wrong-binary cache hits.** On-disk module directories are named `wp_{module.name}_{module_hash.hex()[:7]}` (`get_module_identifier`), i.e. 28 bits of discrimination, while kernel symbols embed a *different* hash (`{key}_{kernel.hash.hex()[:8]}`, `Kernel.get_mangled_name`) that is never validated on a cache hit. Once a cache accumulates many salted variants of the same module, a truncated-hash collision loads a stale cubin that lacks the expected symbol:

```
Warp CUDA error 500: named symbol not found (in function wp_cuda_get_kernel, warp/native/warp.cu:5459)
Warp CUDA error: Failed to lookup kernel function <key>_<hash8>_cuda_kernel_forward in module
...
  File ".../warp/_src/context.py", line 3015, in get_kernel_hooks
KeyError: '<key>_<hash8>_cuda_kernel_forward_smem_bytes'
```

This failed a newton-physics main-branch CI run: https://github.com/newton-physics/newton/actions/runs/30091930217/job/89477357077

`wp.map()` with a plain callable goes through the same path and is equally affected.

## Reproduction — nondeterminism (no GPU needed)

```python
import warp as wp

def make_kernel():
    threshold = 0.05
    fn, _ = wp.utils.create_warp_function(lambda q, qd: q[2] > threshold)

    @wp.kernel
    def test_fn_kernel(body_q: wp.array(dtype=wp.transform),
                       body_qd: wp.array(dtype=wp.spatial_vector),
                       failures: wp.array(dtype=bool)):
        tid = wp.tid()
        failures[tid] = not wp.bool(fn(body_q[tid], body_qd[tid]))
    return test_fn_kernel

k = make_kernel()
wp.launch(k, dim=1,
          inputs=[wp.zeros(1, dtype=wp.transform), wp.zeros(1, dtype=wp.spatial_vector)],
          outputs=[wp.zeros(1, dtype=bool)], device="cpu")
print(k.hash.hex()[:8])
```

Run twice with a fixed `WARP_CACHE_PATH`: the hash differs every run and the module always logs `(compiled)`, never `(cached)`, adding a new cache directory each time. With `PYTHONHASHSEED=0` the hash is stable and the second run logs `(cached)`.

## Reproduction — wrong-binary failure signature (CUDA)

Run the script above once on `cuda:0` to populate the cache and record `k.module.get_module_hash().hex()`. In a second process, pin the module digest to the first process's value (simulating the 28-bit directory collision that occurs naturally against a large cache):

```python
from warp._src import context as ctx
orig = ctx.ModuleHasher.__init__
def patched(self, kernels, options):
    orig(self, kernels, options)
    self.hash = bytes.fromhex(FIRST_PROCESS_MODULE_HASH)
ctx.ModuleHasher.__init__ = patched
```

The second process gets a module cache hit, loads the first process's cubin, and fails with exactly the CUDA error 500 + `KeyError ..._smem_bytes` above.

## System Information

Warp 1.16.0.dev20260716 (also verified in 1.14.0.dev20260518 and 1.15.0.dev20260612), Python 3.12, Linux x86_64, CUDA 12.9/13.x driver. GPU-independent (hashing is host-side); wrong-binary repro run on RTX PRO 6000 Blackwell (sm_120).

shi-eric:
## Description

Closes #1696.

`create_warp_function()` generated function names using Python's process-salted `hash()`. Because those names contribute to kernel and module identity, identical callables received different hashes across processes, preventing persistent kernel cache reuse and accumulating duplicate artifacts. At sufficient cache sizes, truncated module-directory hash collisions could also load stale binaries.

This replaces the salted hash with a deterministic 64-bit suffix derived from SHA-256. Callable behavior remains unchanged, while generated identities are stable across processes. The development version is bumped from `1.17.0.dev0` to `1.17.0.dev1` so previously salted artifacts do not share the corrected cache namespace.

## Changes

- Generate deterministic `func_<16 hex>` names for dynamically created Warp functions.
- Add a cross-process regression test using different `PYTHONHASHSEED` values.
- Bump the Python and native development versions together to isolate the corrected kernel cache.
- Add a changelog fragment describing the user-facing fix.

## Checklist

## Validation summary

The focused regression test passed and confirmed that separate processes using different `PYTHONHASHSEED` values produce the same generated function key with the expected `func_<16 hex>` format.

Pre-commit checks passed for every changed file, including generated-file consistency, version consistency, Ruff linting and formatting, typos, and ClangFormat.

## Bug fix

Before this change, running this script in separate processes with different `PYTHONHASHSEED` values produces different generated function keys:

```python
import warp as wp

def make_callable():
    return lambda q, qd: q[0] == 0.0

created_fn, _ = wp.utils.create_warp_function(make_callable())
print(created_fn.key)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-process kernel cache reuse for lambdas and other Python callables used with mapping and generated Warp functions.
  * Generated function identifiers are now consistent across different Python hash seed settings.

* **Chores**
  * Updated the development version to `1.17.0.dev1`.


eric-heiden (approved):
Thanks, LGTM! Then we don't need https://github.com/newton-physics/newton/pull/3643 anymore.

shi-eric:
Merged with 79a147ced02cf110131fbf6f71d9c546133124b9

PR #1788, opened 2026-08-13
capture_end: add skip_leaf_join to opt out of the broad leaf-join pass

### LINKED ISSUE #1787: wp.capture_end() unconditionally joins all graph leaves, corrupting externally-owned captures
## Problem

`wp.capture_end()` runs a broad leaf-join pass before ending the capture: every leaf node of the captured graph that is not already in the capture stream's dependency set is appended via `cuStreamUpdateCaptureDependencies`. This exists to rescue Warp-forked capture streams that were never joined back, so `cudaStreamEndCapture` does not fail with `cudaErrorStreamCaptureUnjoined`.

That pass is harmful when the Warp capture window is embedded inside an **externally-owned** capture (`torch.cuda.graph`, a CUDASTF context, or any capture created outside Warp and adopted with `wp.capture_begin(external=True)`). At `capture_end` time, the outer graph's leaf set includes nodes belonging to *other, concurrent branches* that the owning integration intends to extend. The broad join adopts those foreign leaves into this window's dependency set and injects false serialization edges: independent branches of the outer graph become ordered and concurrency is silently lost.

## Reproduction sketch

Inside an externally created capture, open two independent Warp capture windows in sequence (`wp.capture_begin(external=True)` / `wp.capture_end()`). After the second `capture_end()`, the resulting graph contains an edge serializing the two windows even though they share no data dependency.

## Proposed fix

Add an opt-out for integrations that own the capture frontier discipline themselves:

```python
wp.capture_end(..., skip_leaf_join=True)
```

Default behavior is unchanged. Unit tests should pin down both semantics:

- the default pass rescues an unjoined forked capture stream,
- `skip_leaf_join=True` refuses to adopt a dangling leaf (CUDA reports the unjoined capture),
- `skip_leaf_join=True` with a caller-managed join captures and replays cleanly.

caugonnet:
## Summary

`wp.capture_end()` joins every unjoined leaf of the captured graph into the capture stream's dependency set, so dangling forked capture streams still end the capture cleanly. That broad pass is harmful when the Warp capture window is embedded inside an **externally-owned** capture (`torch.cuda.graph`, CUDASTF, any capture adopted with `wp.capture_begin(external=True)`): the outer graph's leaf set includes nodes from other concurrent branches, and joining them injects false serialization edges into a frontier Warp does not own.

This PR adds an opt-out for integrations that manage the capture frontier themselves:

```python
wp.capture_end(..., skip_leaf_join=True)
```

Default behavior is unchanged.

Closes #1787.

## Changes

- `wp.capture_end()` gains a keyword-only `skip_leaf_join: bool = False`; plumbed through `wp_cuda_graph_end_capture()` (with a defaulted parameter, so existing native callers are unaffected).
- `warp/tests/cuda/test_graph_skip_leaf_join.py`: three unit tests pinning down the semantics:
  1. the default leaf-join pass rescues an unjoined forked capture stream (functional replay of both branches),
  2. `skip_leaf_join=True` refuses to adopt the dangling leaf, so CUDA rejects the unjoined capture and `capture_end` raises,
  3. `skip_leaf_join=True` with a caller-managed join captures and replays cleanly.
- Towncrier fragment `changelog/1787.added.md`.

## Testing

Built and ran the new tests on an NVIDIA GB300 (CUDA 13.4): 3/3 pass. The suite is registered in `unittest_suites.py` and requires no optional dependencies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an optional `skip_leaf_join` setting when ending CUDA graph captures or using scoped capture.
  * External capture workflows can omit automatic joining of unconnected graph leaves.
  * Existing behavior remains unchanged by default.

* **Documentation**
  * Added changelog documentation for the new capture option.

* **Tests**
  * Added coverage for default, skipped, and explicitly joined graph-capture scenarios, including cleanup after failed termination.


warp/_src/context.py
```diff
@@ -12097,12 +12098,16 @@ def capture_begin(
     _register_capture(device, stream, graph, capture_id)
 
 
-def capture_end(device: DeviceLike = None, stream: Stream | None = None) -> Graph:
+def capture_end(device: DeviceLike = None, stream: Stream | None = None, *, skip_leaf_join: bool = False) -> Graph:
```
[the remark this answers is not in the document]
caugonnet: Good catch — addressed in 316c9ea49. `ScopedCapture` now accepts `skip_leaf_join` and forwards it to `capture_end()`, and a new test (`test_scoped_capture_windows_preserve_owner_frontier`) drives the externally-owned pattern end to end: a capture begun directly through the CUDA runtime, adopted from forked streams by `ScopedCapture(external=True)` guest windows, with the false serialization edge asserted via `cudaGraphGetEdges` (exactly one extra edge with the default join, none with the opt-out).

warp/native/warp.h
```diff
@@ -764,7 +764,7 @@ WP_API void wp_cuda_event_synchronize(void* event);
 WP_API float wp_cuda_event_elapsed_time(void* start_event, void* end_event);
 
 WP_API bool wp_cuda_graph_begin_capture(void* context, void* stream, int external, int mode);
-WP_API bool wp_cuda_graph_end_capture(void* context, void* stream, void** graph_ret);
+WP_API bool wp_cuda_graph_end_capture(void* context, void* stream, void** graph_ret, bool skip_leaf_join = false);
```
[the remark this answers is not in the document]
caugonnet: The Warp native library is not a standalone stable C ABI: it ships pinned inside the Python package, and the Python layer (whose ctypes argtypes are updated in lockstep in this PR) is its only supported caller. Adding a defaulted trailing parameter to a `WP_API` function follows existing precedent in this codebase (e.g. `wp_free_device_async()` gained `void** dbg_node_ret = nullptr` the same way). A hypothetical third-party binary linking `warp.so` across versions is not a supported configuration, so keeping the same symbol with a defaulted parameter is preferable to forking a second entry point.

warp/native/warp.cu
```diff
@@ -3560,28 +3560,30 @@ bool wp_cuda_graph_end_capture(void* context, void* stream, void** graph_ret)
...
+                    CU_STREAM_ADD_CAPTURE_DEPENDENCIES
+                ));
+                // ensure graph is still valid
+                if (get_capture_graph(cuda_stream) != graph) {
+                    clean_up();
+                    return false;
+                }
             }
         }
     }
```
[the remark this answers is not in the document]
caugonnet: The allocation cleanup on this failure path is already handled, by design, through the graph user-object mechanism rather than `clean_up()`: when the capture has unfreed or temporary allocations, a `cudaUserObject` carrying them is retained on the raw graph *before* `cudaStreamEndCapture()` (deliberately, so the same mechanism also covers external captures). If `cudaStreamEndCapture()` then fails, CUDA destroys the invalidated graph, the user object is released, and `on_graph_destroy()` queues the allocations for the deferred pass that marks `graph_destroyed` and frees the unreferenced ones. The two paths cannot double-free: when `clean_up()` handles an allocation synchronously it erases it from `g_graph_allocs`, so the callback's lookup misses. This PR does not change that path — `skip_leaf_join` only makes the failure reachable on purpose in the new test, which performs no allocations during capture. Extracting the unreference loop would run it a second time on allocations the user object already owns, which is exactly the double handling the current structure avoids.

warp/tests/cuda/test_graph_skip_leaf_join.py
```diff
@@ -0,0 +1,243 @@
...
+
+        try:
+            num_edges = ctypes.c_size_t()
+            # CUDA 13 promoted the edge-data variant of cudaGraphGetEdges:
+            # (graph, from, to, edgeData, numEdges). Try it first, then fall
+            # back to the classic CUDA 12 signature (graph, from, to, numEdges).
+            err = cudart.cudaGraphGetEdges(graph, None, None, None, ctypes.byref(num_edges))
+            if err != 0:
+                num_edges = ctypes.c_size_t()
+                err = cudart.cudaGraphGetEdges(graph, None, None, ctypes.byref(num_edges))
```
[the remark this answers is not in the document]
caugonnet: Fair point — the error-based fallback was unsound: a CUDA 13 call can validly return `cudaErrorLossyQuery`, and the fallback would then pass `num_edges` in the `edgeData` position. Fixed in b41aeab8c: the test now queries `cudaRuntimeGetVersion()` and selects the CUDA 12 four-argument or CUDA 13 five-argument signature explicitly. The NULL edge-data count query stays lossless for this graph since kernel launches and event ordering carry default edge data.

warp/native/warp.h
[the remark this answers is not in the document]
caugonnet: Yes, that phrasing is accurate — please remember it. (Maintainers can refine it later via `.greptile/rules.md` if they want a different scope.)

warp/tests/cuda/test_graph_skip_leaf_join.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    wp.capture_begin(device=device, stream=main, force_module_load=False)
+    forked.wait_event(main.record_event())
+    wp.launch(_write_val, dim=N, inputs=[a, 1], stream=main)
+    wp.launch(_write_val, dim=N, inputs=[b, 2], stream=forked)
+
+
+def test_leaf_join_rescues_unjoined_fork(test, device):
+    """Default capture_end: the leaf-join pass adopts the dangling forked
+    branch, the capture succeeds, and the branch's work is in the graph."""
```
shi-eric: Nitpick: Please ensure docstrings here follow Google-style docstring conventions, e.g. short description should fit on the first line of the docstring here

warp/tests/cuda/test_graph_skip_leaf_join.py
```diff
@@ -0,0 +1,252 @@
...
+import numpy as np
+
+import warp as wp
+from warp.tests.unittest_utils import *
+
+N = 1024
+
+
+@wp.kernel
+def _write_val(arr: wp.array(dtype=int), value: int):
```
shi-eric: Nitpick: Please avoid reintroducing factory-style kernel annotations here. Use the newer subscript-style annotation instead, e.g. `wp.array[int]`)

warp/tests/cuda/test_graph_skip_leaf_join.py
```diff
@@ -0,0 +1,252 @@
...
+    forked = wp.Stream(device)
+    a = wp.zeros(N, dtype=int, device=device)
+    b = wp.zeros(N, dtype=int, device=device)
+    wp.load_module(device=device)
+
+    _begin_forked_capture(device, main, forked, a, b)
+    with test.assertRaises(RuntimeError):
+        wp.capture_end(device=device, stream=main, skip_leaf_join=True)
+
+    wp.synchronize_device(device)
```
shi-eric: Seems like this synchronize isn't needed?

warp/tests/cuda/test_graph_skip_leaf_join.py
caugonnet: Done in be0417bfc — all function docstrings now carry a one-line short description (Google style).

warp/tests/cuda/test_graph_skip_leaf_join.py
caugonnet: Done in be0417bfc — switched to the subscript-style annotation (`wp.array[int]`). Also applied the same convention to #1789.

warp/tests/cuda/test_graph_skip_leaf_join.py
caugonnet: Correct — nothing was executed (all work was captured, and the failed capture_end leaves nothing in flight). Removed in be0417bfc; the following tests stay green without it.

PR #1819, opened 2026-08-19
Refresh Trimesh side topology after in-place connectivity changes

### LINKED ISSUE #1804: FEM example: Trimesh2D side topology goes stale after in-place edge flips
### Description

`Example.remesh()` in `warp/examples/fem/example_elastic_shape_optimization.py` flips
edges directly in `_tri_vertex_indices` and deliberately does not rebuild the
`fem.Trimesh2D`. The comment there justifies this:

```python
# We intentionally avoid recreating the Trimesh2D objects: edge flips
# only affect interior edges, so boundary structures and projectors
# remain valid; ...
```

That premise does not hold: a flip can move an *unflipped* edge from one triangle to
another, which invalidates the geometry's cached side-to-triangle mapping.

### Why

For the flipped pair `t = (c, a, b)` and `n = (d, b, a)` sharing edge `a-b`, the result
is `t = (c, a, d)` and `n = (d, b, c)`. Triangle `t` used to hold edges `a-b`, `b-c` and
`c-a`; it now holds `a-d`, `d-c` and `c-a`. Edge `b-c` was never flipped, but it left `t`
and now belongs to `n`. Any flipper that preserves the cyclic slot layout does this — see
`warp/_src/geometry.py` in the `_apply_won_flips` kernel (`TT[n, jn] = n_bc`), and the
NumPy flipper it replaced had the same effect via `tri[t0] = [a, d, c]` / `tri[t1] = [b, c, d]`.

`Trimesh2D` builds `_edge_vertex_indices` and `_edge_tri_indices` once in `__init__`, so
after the flip `_edge_tri_indices` still claims that `b-c` belongs to `t`.
`Trimesh2D._edge_to_tri_coords()` (`warp/_src/fem/geometry/trimesh.py`) then maps a point
on that side into the triangle by matching the edge's endpoints against the triangle's
vertices:

```python
if edge_vidx[0] == v0:
    cx = side_coords.dtype(1.0) - side_coords[0]
elif edge_vidx[0] == v1:
    cy = side_coords.dtype(1.0) - side_coords[0]
else:
    cz = side_coords.dtype(1.0) - side_coords[0]
```

With `b` no longer present in `t`, the lookup falls through to the `else` branch and
silently assigns the coordinate to the wrong vertex slot. There is no error, just a
quadrature point evaluated at the wrong location in the wrong triangle.

### Impact

If the migrated edge is on the boundary, boundary integration over it is wrong — in this
example that is the Dirichlet projectors on the left/right subdomains and the load on the
bottom. The stale interior edge list also lists edges that no longer exist, which would
matter for any interior-side integration.

Separately, the triangle BVH is built from the original triangles' bounds
(`build_bvh=True`), so after flips it no longer bounds the actual triangles and
`fem.lookup()` (used by the `mirror_val` form) can return the wrong cell.
`Geometry.update_bvh()` only refits an existing BVH, so it is not sufficient on its own
for a topology change.

### Not a regression

This behavior predates the Warp Delaunay flipper (GH-1797): the previous host-side NumPy
flipper produced the same topology change and `remesh()` already skipped the rebuild. The
Warp port swapped the flipper without changing the caching around it.

### Suggested fix

Rebuild the `Trimesh2D` (or refresh its side topology and BVH) after `remesh()` applies
flips, and construct the spaces from the refreshed geometry.

The complication is noted in that same comment: re-running `_build_topology()` gives a
nondeterministic edge ordering that perturbs the boundary projector enough to visibly
change the displacement on this ill-conditioned system. So the fix will move the example's
output, and whoever does it should confirm the new output is correct rather than just
different. A `Trimesh2D` API for refreshing topology in place after a connectivity change
(preserving edge ordering where possible) would be the cleaner fix.

### System Information

Not environment specific — this is a logic issue in the example and the cached topology it
relies on.

nileshpatil6:
Fixes #1804.

`Trimesh` builds `_edge_vertex_indices`, `_edge_tri_indices` and `_boundary_edge_indices` from `tri_vertex_indices` once, in the constructor. `example_elastic_shape_optimization.py` rewrites that array in place after Delaunay edge flips and deliberately skips the rebuild, with a comment explaining that flips only affect interior edges so boundary structures and projectors stay valid.

That premise does not hold. Flipping the edge shared by `(c, a, b)` and `(d, b, a)` gives `(c, a, d)` and `(d, b, c)`, which moves the untouched edge `b-c` out of the first triangle and into the second. The cached map still assigns `b-c` to the first triangle, so `_edge_to_tri_coords()` looks for a vertex that is no longer in that triangle, falls through to the `else` branch, and places the quadrature point at the wrong location in the wrong triangle. Nothing raises. The BVH is bounded per triangle and is stale for the same reason, and `Geometry.update_bvh()` only refits an existing tree, so it is not sufficient after a connectivity change.

Reproduced on the two-triangle case above, checking each cached edge against the triangles actually present:

```
stale side->tri entries WITHOUT update_topology(): 6
stale side->tri entries WITH    update_topology(): 0
matches a freshly built Trimesh2D: True
```

`edge (1,2)` is the interesting one: that is `b-c`, an edge no flip selected, and the stale map points it at a triangle that no longer contains vertex 1.

## Change

`Trimesh.update_topology()` rebuilds the side topology, reorients the edges, invalidates the cached `CellArg`/`SideArg` structs that hold references to the old arrays, and rebuilds the BVH on each device that has one. The orientation launch the constructor ran inline moves into `_orient_edges()` so both paths share it.

The example calls it on the affected geometries after applying flips, before rebuilding the spaces and fields that depend on the side ordering. `_start_geo` and `_geo` share the same triangle array, so both go stale; in the deformed-mesh mode `_geo` is derived from `_start_geo` and is rebuilt by `_rebuild_fem_structures()`.

The issue suggests preserving edge ordering where possible. This does not attempt that: the ordering comes out of `compress_node_indices` and preserving it across a connectivity change would be a much larger change for a path that already has to rebuild its spaces and fields. The docstring states that sides may be reordered.

## Verification

New test `test_trimesh_update_topology` in `warp/tests/fem/test_fem_geometry.py` asserts a freshly built mesh has no stale sides, that flipping in place introduces some, that `update_topology()` removes them, and that the result matches a `Trimesh2D` built from the flipped triangles.

- `test_trimesh_update_topology`: passes on `cpu` and `cuda:0`. Errors with `AttributeError` on unmodified main, and the stale-entry assertion is what fails if the rebuild is removed.
- `warp.tests.fem.test_fem_geometry`: 27 tests, OK (1 skipped), on `cpu` and `cuda:0`.
- `fem.example_elastic_shape_optimization` example test: OK.
- `ruff check` and `ruff format --check` clean on the touched files.

One note on the concern in the issue that this would move the example's output. I could not get the flipper to produce a flip in that example: at `resolution=16` over 80 iterations and at `resolution=25` over 150 iterations, remeshing every 5 or 10 steps, `delaunay_edge_flip` returned 0 flips every time and `remesh()` returned before touching the connectivity. So I have no measurement showing the output moving, and the example test result is unchanged. That also explains how the stale topology went unnoticed. If you have a configuration that does flip, I am happy to run it and report the before and after.

Tested on Windows 11, RTX 3050, CUDA 12.9.

## Summary by CodeRabbit

* **New Features**
  * Added a public operation to refresh mesh topology after in-place triangle connectivity changes.
  * Mesh connectivity, cached data, boundary-edge information, and acceleration structures are rebuilt consistently after updates.

* **Bug Fixes**
  * Fixed elastic shape optimization remeshing so geometry and topology remain synchronized after edge flips.
  * Improved reliability of refreshed mesh data across supported devices.

### DIFF
--- a/changelog/1804.added.md
+++ b/changelog/1804.added.md
@@ -1,0 +1,3 @@
+`fem.Trimesh.update_topology()`, which rebuilds the edge list, the edge-to-triangle map, the boundary edges and the
+BVH after `tri_vertex_indices` has been modified in place, so that meshes whose connectivity changes at runtime no
+longer integrate over stale sides.
--- a/changelog/1804.fixed.md
+++ b/changelog/1804.fixed.md
@@ -1,0 +1,3 @@
+The elastic shape optimization example now refreshes its geometry after Delaunay edge flips. A flip can move an edge
+it did not select from one triangle to another, so the cached side topology and the triangle BVH no longer described
+the mesh being integrated over.
--- a/warp/_src/fem/geometry/trimesh.py
+++ b/warp/_src/fem/geometry/trimesh.py
@@ -292,9 +292,11 @@ class Trimesh(Geometry):
         self._build_topology(temporary_store)
         self._orient_edges()
 
-        # Arg structs hold references to the previous edge arrays.
+        # Arg structs hold references to the previous edge arrays. side_index_arg_value caches
+        # the boundary edge list, which _build_topology() reallocates, so it has to go too.
         Geometry.cell_arg_value.invalidate(self)
         Geometry.side_arg_value.invalidate(self)
+        Geometry.side_index_arg_value.invalidate(self)
 
         # Cell bounds are derived from the triangles, so a connectivity change invalidates the tree
         # itself and not just its bounds; refitting through update_bvh() is not enough here.
--- a/warp/_src/fem/geometry/trimesh.py
+++ b/warp/_src/fem/geometry/trimesh.py
@@ -114,17 +114,7 @@ class Trimesh(Geometry):
         self._edge_tri_indices: wp.array = None
         self._build_topology(temporary_store)
 
-        # Flip edges so that normals point away from inner cell
-        if self.dimension == 2:
-            orient_kernel = Trimesh._orient_edges_2d
-        else:
-            orient_kernel = Trimesh._orient_edges_3d
-        wp.launch(
-            kernel=orient_kernel,
-            device=positions.device,
-            dim=self.side_count(),
-            inputs=[self._edge_vertex_indices, self._edge_tri_indices, self.tri_vertex_indices, self.positions],
-        )
+        self._orient_edges()
 
         # Process all dynamic attributes (Trimesh primitives + Geometry dependents)
         cache.setup_dynamic_attributes(self)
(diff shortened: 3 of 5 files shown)


nileshpatil6:
Good catch, that one is real and I have pushed a fix.

`fill_side_index_arg()` reads `_boundary_edge_indices`, and `_build_topology()` reallocates that array, but `side_index_arg_value` is cached per device and I was only invalidating `cell_arg_value` and `side_arg_value`. Any geometry whose side-index arg had already been built kept handing out the pre-flip boundary edge list, so `BoundarySides` selected the wrong edges and boundary integration stayed wrong across the refresh. That is the same class of defect the PR is fixing, one level up.

Added `Geometry.side_index_arg_value.invalidate(self)` alongside the other two.

You were also right that the test did not exercise a previously cached boundary-side argument. It now builds all three Arg structs before the flip, so the cache is populated and has to be invalidated. That assertion is what fails without the change:

```
AssertionError: Lists differ: [0, 2, 3, 4] != [0, 1, 2, 3]
FAILED (failures=2)
```

on both `cpu` and `cuda:0`, and passes with it. Full `test_fem_geometry` is still 27 tests OK (1 skipped).

PR #1822, opened 2026-08-20
Limit usd-exchange below version 3

LouRohanNV:
## Description

Keep Warp’s Linux aarch64 benchmark and example extras on `usd-exchange` 2.x.

`usd-exchange` 3 is a major-version jump. It should enter these environments only after compatibility tests.

## Changes

- Add the `<3` bound to the benchmark dependency.
- Add the `<3` bound to the examples dependency.
- Update the matching lock metadata.
- Add a changelog fragment.

## Checklist

- [ ] I added a changelog fragment if this change affects users.

## Validation summary

I confirmed that the package and lock files contain the same upper bound.

The lock file still selects `usd-exchange` 2.3.0. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility on Linux aarch64 systems by restricting the USD exchange dependency to supported versions.
  * Updated benchmark and example installations to use compatible USD exchange releases.


shi-eric:
Thanks, Lou! There is no need to cap the `usd-exchange` for Warp, though. I've updated the `uv.lock` file already in 442eb7b5cba726dca8d1e36c4ac336b03b45e21a

Issue #679, opened 2025-04-24
[BUG] `wp.svd2` returns NaN in `U` and `V`


akrivx:
### Bug Description

There seems to be an issue with the `wp.svd2` function, which consistently returns NaN values in the `U` and `V` matrices. Even for a simple diagonal input, `U` and `V` are filled with NaNs, although singular values appear correctly computed.

#### Minimal Reproducible Example
```python
import warp as wp
wp.init()

@wp.kernel
def test_svd2(U: wp.array(dtype=wp.mat22), s: wp.array(dtype=wp.vec2), V: wp.array(dtype=wp.mat22)):
    U2 = wp.mat22()
    V2 = wp.mat22()
    s2 = wp.vec2()
    wp.svd2(wp.diag(wp.vec2(2.0)), U2, s2, V2)
    U[0] = U2
    s[0] = s2
    V[0] = V2

U = wp.zeros(1, dtype=wp.mat22)
V = wp.zeros(1, dtype=wp.mat22)
s = wp.zeros(1, dtype=wp.vec2)

wp.launch(kernel=test_svd2, dim=1, inputs=[U, s, V])

print("U:", U)
print("V:", V)
print("s:", s)
```

#### Observed Output
```bash
U: [[[nan nan]
     [nan nan]]]
V: [[[nan nan]
     [nan nan]]]
s: [[2. 2.]]
```

#### Expected Behaviour
`U` and `V` should contain valid orthogonal matrices (no NaNs) resulting from the decomposition of a simple diagonal matrix.

### System Information

- NVIDIA Warp Version: 1.6.2
- CUDA Version: CUDA Toolkit 12.8, Driver 12.9
- GPU Model: NVIDIA GeForce RTX 5080
- Python Version: 3.12.

shi-eric:
@gdaviet added improvements to `wp.svd2` which should address the issue reported here. It's available in the `main` branch already and will be available in future nightly builds. See 6f0b1c25b28609ee851592bc0ecc8d8b34cb33e5

PR #1831, opened 2026-08-24
Fix svd2 U/V for scaled orthogonal inputs (GH-1734)

### LINKED ISSUE #1734: wp.svd2 returns wrong singular vectors for scaled rotations and reflections
### Bug Description

For any scaled orthogonal input (`A = s * R`, with `R` a rotation or reflection), `wp.svd2` returns
`U = V = identity`, which reconstructs `A` only when `A` is a non-negative multiple of the identity.
The singular values are correct in every case; only `U` and `V` are wrong. There is no error, no
warning, and no NaN.

Affected inputs are ordinary ones for 2D simulation — exact rotations, `-I`, `diag(1, -1)` — so
polar decomposition, co-rotational strain, and rotation extraction silently get a wrong rotation.

It is a cliff, not a drift: a matrix a hair off a rotation reconstructs to ~1e-07, while the exact
case is wrong by up to 2.0. It is also device-dependent, because whether an intermediate lands
exactly on zero decides which path an input takes:

| `A` | CPU | CUDA |
| --- | --- | --- |
| `rot(90°)`, `-I`, `diag(1, -1)` | wrong | **wrong** |
| `rot(0.5)`, `rot(pi/4)`, `2 * rot(0.5)` | wrong | correct (~1e-07) |
| `I`, `5I` | correct | correct |

The path involved is the repeated-singular-value special case added by the fix for #679 (`svd2`
returning NaN on `diag(2, 2)`). Its comment records the assumption that a duplicate eigenvalue means
a scaled identity; the condition that actually reaches it is `Aᵀ A = lambda * I`, the whole
scaled-orthogonal family.

#### Reproducer

```python
import numpy as np
import warp as wp


@wp.kernel
def decompose(a: wp.array[wp.mat22], rec: wp.array[wp.mat22], sigma: wp.array[wp.vec2]):
    tid = wp.tid()
    U, s, V = wp.svd2(a[tid])
    sigma[tid] = s
    rec[tid] = U * wp.diag(s) * wp.transpose(V)


a = wp.array([wp.mat22(0.0, -1.0, 1.0, 0.0)], dtype=wp.mat22)  # 90-degree rotation
rec = wp.zeros(1, dtype=wp.mat22)
sigma = wp.zeros(1, dtype=wp.vec2)
wp.launch(decompose, dim=1, inputs=[a], outputs=[rec, sigma])

print(sigma.numpy()[0])                              # [1. 1.]  <- correct
print(rec.numpy()[0])                                # [[1. 0.] [0. 1.]]  <- should be the input
print(np.abs(rec.numpy()[0] - a.numpy()[0]).max())   # 1.0
```

#### What was measured

float32, on CPU (x86_64) and CUDA 12.6.3 / NVIDIA L40:

- `max |U diag(sigma) Vᵀ - A|`: 1.0 for `rot(90°)`, 2.0 for `-I` and `diag(1, -1)`, 0.96 for
  `2 * rot(0.5)`, 0.48 for `rot(0.5)`. `sigma` matches NumPy throughout.
- `det(V) = +1`, and `det(U) = -1` exactly when `det(A) < 0`, hold over a 72-matrix sample — the sign
  conventions are intact, only the degenerate family is wrong.
- `test_svd_2D` already asserts this identity at a 3e-05 tolerance, including a hand-built list of
  degenerate cases. That list has no rotation and no reflection; `I` and `5I` are the two members of
  the affected family the special case gets right, which is why the suite is green.

Found while documenting the matrix decomposition built-ins for #1498.

### System Information

- Warp 1.17.0.dev1 (built from source at 79a147ced)
- Python 3.12.13, Linux 5.15.0-113-generic, x86_64
- CUDA 12.6.3, NVIDIA L40 (sm_89), driver 570.158.01

azrabano23:
> Continuation of #1810 — that PR was closed by mistake during a repo-management sweep on my account, and GitHub refuses to reopen it. The branch and diff are identical and unchanged.

## Description

`wp.svd2()` returns `U = V = identity` for any scaled orthogonal input `A = s * R` (rotations,
reflections, negative multiples of the identity), which reconstructs `A` only when `A` is a
non-negative multiple of the identity. The singular values are correct; only the singular vectors
are wrong, silently — no error, no NaN. Downstream users of 2D polar decomposition, co-rotational
strain, and rotation extraction get a wrong rotation with no diagnostic.

Root cause: the repeated-singular-value special case added by the fix for #679 (`svd2` returning
NaN on `diag(2, 2)`) assumes that a duplicate eigenvalue of `AᵀA` means `A` is a scaled identity.
The condition that actually reaches that branch is `AᵀA = λI` (`discriminant == 0` requires both
`ATA11 == ATA22` and `ATA12 == 0`), which characterizes the whole scaled-orthogonal family: columns
of `A` orthogonal with equal norm `σ = sqrt(λ)`.

The fix uses exactly that property: for `λ > 0`, `A / σ` is exactly orthogonal, so returning
`U = A / σ`, `Σ = σI`, `V = I` reconstructs `A` while keeping `U` orthogonal and preserving the
existing sign convention `det(U) = sign(det(A))`, `det(V) = +1`. `A = 0` keeps `U = V = I`, `Σ = 0`.
The #679 case (`A = sI`, `s > 0`) returns identical results to before by construction
(`A / σ = I`).

Closes #1734.

## Changes

- `warp/native/svd.h`: in `_svd_2`'s duplicate-eigenvalue branch, return `U = A / σ` (with an
  `A = 0` guard) instead of unconditionally `U = I`; `V = I` unchanged. Header code shared by all
  scalar types and both devices.
- `warp/tests/matrix/test_mat.py`: add five scaled-orthogonal matrices (rot 90°, rot 0.5,
  `2·rot 0.5`, `-I`, reflection `diag(1, -1)`) to `test_svd_2D`'s hand-built edge-case list, which
  already asserts `U·diag(σ)·Vᵀ = A`, orthogonality of `U` and `V`, and gradients.
- `changelog/1734.fixed.md`: changelog fragment.

## Checklist

## Validation summary

Validated on macOS arm64, CPU device, native library built from source with `build_lib.py`
(CPU-only build; formatting via `uvx pre-commit run`, all hooks pass).

- **Red:** with only the new test matrices applied, `warp/tests/matrix/test_mat.py` fails at
  `test_svd_2Dfloat16_cpu` — the mismatching elements are exactly the new rotation/reflection
  entries (max absolute error 2.0, the `-I` case).
- **Green:** with the fix, the full module passes: `Ran 69 tests ... OK`, including
  `test_svd_2D{float16,float32,float64}_cpu` with the new cases and the existing gradient checks.
- **Reproducer sweep** (float32 and float64 kernels, CPU), max `|U·diag(σ)·Vᵀ − A|`:

  | Input | Before | After (f32) | After (f64) |
  | --- | --- | --- | --- |
  | `rot(90°)` | 1.0 | 0.0 | 0.0 |
  | `-I` | 2.0 | 0.0 | 0.0 |
  | `diag(1, -1)` | 2.0 | 0.0 | 0.0 |
  | `rot(0.5)` | 0.479 | 0.0 | 0.0 |
  | `2·rot(0.5)` | 0.959 | 0.0 | 0.0 |
  | `rot(π/4)` | 0.707 | 0.0 | 0.0 |
  | `I`, `5I`, `diag(2, 2)` (#679) | 0.0 | 0.0 | 0.0 |
  | zero matrix | 0.0 | 0.0 | 0.0 |
  | `I + 1e-7·offset` (near-repeated) | 7.1e-15 | 7.1e-15 | 1.1e-16 |

  `U` stays orthogonal to ≤ 1.4e-7 (f32) / 2.2e-16 (f64), singular values match NumPy throughout,
  no NaNs, and the near-repeated-singular-value input still takes the general path unchanged.

- **Not run:** CUDA. No NVIDIA GPU is available on this machine. The changed code is
  device-independent header code (`_svd_2` template in `svd.h`) with no device-specific branches,
  compiled identically for both backends; the issue reports the same wrong output on CUDA for the
  inputs whose intermediates land exactly on zero, which this change corrects by the same math.

## Bug fix

Minimal reproducer without this PR (from #1734):

```python
import numpy as np
import warp as wp


@wp.kernel
def decompose(a: wp.array[wp.mat22], rec: wp.array[wp.mat22], sigma: wp.array[wp.vec2]):
    tid = wp.tid()
    U, s, V = wp.svd2(a[tid])
    sigma[tid] = s
    rec[tid] = U * wp.diag(s) * wp.transpose(V)


a = wp.array([wp.mat22(0.0, -1.0, 1.0, 0.0)], dtype=wp.mat22)  # 90-degree rotation
rec = wp.zeros(1, dtype=wp.mat22)
sigma = wp.zeros(1, dtype=wp.vec2)
wp.launch(decompose, dim=1, inputs=[a], outputs=[rec, sigma])

print(sigma.numpy()[0])                              # [1. 1.]  <- correct
print(rec.numpy()[0])                                # [[1. 0.] [0. 1.]]  <- should be the input
print(np.abs(rec.numpy()[0] - a.numpy()[0]).max())   # 1.0 without this PR, 0.0 with it
```

---

Noting that #1734 is self-assigned — if a fix is already underway in-house, happy to close this in its favor or rework it to match; the added test matrices may be useful either way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed 2D singular value decomposition for scaled rotations, reflections, negative identity matrices, and other orthogonal inputs.
  * Improved stability for very small and zero-valued matrices.
  * Improved handling of repeated singular values and gradient calculations.
  * Ensured singular vectors are returned correctly in affected edge cases.

* **Tests**
  * Added coverage for low-magnitude inputs, orthogonal matrices, repeated singular values, gradients, and related edge cases.

### DIFF
--- a/changelog/1734.fixed.md
+++ b/changelog/1734.fixed.md
@@ -1,0 +1,3 @@
+Fix `wp.svd2()` returning an incorrect `U` matrix for scaled orthogonal
+inputs, including rotations, reflections, and negative multiples of the
+identity.
--- a/warp/native/svd.h
+++ b/warp/native/svd.h
@@ -508,6 +508,27 @@ inline CUDA_CALLABLE void _svd_2( // input A
     // output V
     Type& v11, Type& v12, Type& v21, Type& v22)
 {
+    // Step 0: Scale by the largest magnitude entry so the squared terms below
+    // cannot underflow (which would fake a zero discriminant for low-magnitude
+    // inputs, most visibly in float16). U and V are scale-invariant; the
+    // singular values are rescaled at the end.
+    Type amax = max(max(abs(a11), abs(a12)), max(abs(a21), abs(a22)));
+    if (amax == Type(0)) {
+        // A = 0
+        s1 = s2 = Type(0);
+        u11 = v11 = Type(1);
+        u12 = v12 = Type(0);
+        u21 = v21 = Type(0);
+        u22 = v22 = Type(1);
+        return;
+    }
+    // Divide directly rather than multiplying by 1 / amax: the reciprocal
+    // overflows to infinity in float16 for amax below ~1.5e-5.
+    a11 /= amax;
+    a12 /= amax;
+    a21 /= amax;
+    a22 /= amax;
+
     // Step 1: Compute ATA
     Type ATA11 = a11 * a11 + a21 * a21;
     Type ATA12 = a11 * a12 + a21 * a22;
@@ -520,12 +541,21 @@ inline CUDA_CALLABLE void _svd_2( // input A
 
     // Step 3: Singular values
     if (discriminant == Type(0)) {
-        // Duplicate eigenvalue, A ~ s Id
-        s1 = s2 = sqrt(Type(0.5) * trace);
-        u11 = v11 = Type(1);
-        u12 = v12 = Type(0);
-        u21 = v21 = Type(0);
-        u22 = v22 = Type(1);
+        // Duplicate eigenvalue, ATA = lambda Id, so A = s Q with Q orthogonal
+        // (any scaled rotation or reflection, not just a scaled identity).
+        // A / s is exactly orthogonal, so U = A / s, V = Id reconstructs A.
+        // The scaling above guarantees s > 0 here.
+        Type s = sqrt(Type(0.5) * trace);
+        Type inv_sigma = Type(1) / s;
+        s1 = s2 = s * amax;
+        v11 = Type(1);
+        v12 = Type(0);
+        v21 = Type(0);
+        v22 = Type(1);
+        u11 = a11 * inv_sigma;
+        u12 = a12 * inv_sigma;
+        u21 = a21 * inv_sigma;
+        u22 = a22 * inv_sigma;
         return;
     }
 
@@ -559,9 +589,9 @@ inline CUDA_CALLABLE void _svd_2( // input A
     u12 = -u21 * det_sign;
     u22 = u11 * det_sign;
 
-    // Step 6: Set S
-    s1 = sigma1;
-    s2 = sigma2;
+    // Step 6: Set S (undo the input scaling)
+    s1 = sigma1 * amax;
+    s2 = sigma2 * amax;
 }
 
 template <typename Type>
@@ -655,6 +685,53 @@ inline CUDA_CALLABLE void adj_svd2(
 )
 {
     const Type epsilon = _svd_config<Type>::SVD_EPSILON;
+
+    // Recompute _svd_2's branch predicate with the same arithmetic (including the
+    // underflow-avoiding scaling) so forward and adjoint always agree on the path
+    // taken: equal sigmas alone can also arise from rounding on the general path,
+    // where V is not the identity.
+    Type amax = max(max(abs(A.data[0][0]), abs(A.data[0][1])), max(abs(A.data[1][0]), abs(A.data[1][1])));
+    if (amax == Type(0)) {
+        // A = 0 has no well-defined direction
+        return;
+    }
+    Type a11 = A.data[0][0] / amax;
+    Type a12 = A.data[0][1] / amax;
+    Type a21 = A.data[1][0] / amax;
+    Type a22 = A.data[1][1] / amax;
+    Type ATA11 = a11 * a11 + a21 * a21;
+    Type ATA12 = a11 * a12 + a21 * a22;
+    Type ATA22 = a12 * a12 + a22 * a22;
+    Type diff = ATA11 - ATA22;
+    Type discriminant = diff * diff + Type(4) * ATA12 * ATA12;
+
+    if (discriminant == Type(0)) {
+        // _svd_2's duplicate-eigenvalue branch sets sigma = |A|_F / sqrt(2),
+        // U = A / sigma, and V = Id. U and V are not individually unique there,
+        // so a convention is required: recover the cotangent C of the smooth
+        // reconstruction U diag(sigma) V^T symmetrically from adj_U and adj_V
+        // (for a loss <C, U diag(sigma) V^T>, adj_U = sigma C and
+        // adj_V = sigma C^T U, so C = (adj_U + U adj_V^T) / (2 sigma)), then
+        // remove the part already carried by adj_sigma. Exact for losses of
+        // the reconstruction and of the singular values; the pure-gauge
+        // response is dropped.
+        Type inv_two_sigma = Type(0.5) / sigma[0];
+        Type c00
+            = (adj_U.data[0][0] + U.data[0][0] * adj_V.data[0][0] + U.data[0][1] * adj_V.data[0][1]) * inv_two_sigma;
+        Type c01
+            = (adj_U.data[0][1] + U.data[0][0] * adj_V.data[1][0] + U.data[0][1] * adj_V.data[1][1]) * inv_two_sigma;
+        Type c10
+            = (adj_U.data[1][0] + U.data[1][0] * adj_V.data[0][0] + U.data[1][1] * adj_V.data[0][1]) * inv_two_sigma;
+        Type c11
+            = (adj_U.data[1][1] + U.data[1][0] * adj_V.data[1][0] + U.data[1][1] * adj_V.data[1][1]) * inv_two_sigma;
+        Type d0 = adj_sigma[0] - (U.data[0][0] * c00 + U.data[1][0] * c10);
+        Type d1 = adj_sigma[1] - (U.data[0][1] * c01 + U.data[1][1] * c11);
+        adj_A.data[0][0] += c00 + U.data[0][0] * d0;
+        adj_A.data[0][1] += c01 + U.data[0][1] * d1;
+        adj_A.data[1][0] += c10 + U.data[1][0] * d0;
+        adj_A.data[1][1] += c11 + U.data[1][1] * d1;
+        return;
+    }
 
     Type s1_squared = sigma[0] * sigma[0];
     Type s2_squared = sigma[1] * sigma[1];
--- a/warp/tests/matrix/test_mat.py
+++ b/warp/tests/matrix/test_mat.py
@@ -933,6 +933,19 @@ def test_svd_2D(test, device, dtype, register_kernels=False):
                 np.array([[1.0, 1.0], [-1.0, -1.0]]),
                 np.array([[3.0, 0.0], [4.0, 5.0]]),
                 np.eye(2) + tol * np.array([[1.0, 1.0], [-1.0, -1.0]]),
+                # scaled orthogonal inputs hit the repeated-singular-value path (GH-1734)
+                np.array([[0.0, -1.0], [1.0, 0.0]]),  # rotation by 90 degrees
+                np.array([[np.cos(0.5), -np.sin(0.5)], [np.sin(0.5), np.cos(0.5)]]),  # rotation by 0.5 rad
+                2.0 * np.array([[np.cos(0.5), -np.sin(0.5)], [np.sin(0.5), np.cos(0.5)]]),  # scaled rotation
+                -np.eye(2),  # negative scaled identity
+                np.array([[1.0, 0.0], [0.0, -1.0]]),  # reflection
+                # low-magnitude rank-1 input: the squared terms of the
+                # discriminant underflow in float16 unless the input is
+                # rescaled first
+                np.array([[0.005, 0.002], [0.0, 0.0]]),
+                # tiny input: 1 / amax overflows to infinity in float16 for
+                # amax below ~1.5e-5, so the rescaling must divide directly
+                np.array([[1.0e-5, 0.0], [0.0, 5.0e-6]]),
             ],
         ),
         axis=0,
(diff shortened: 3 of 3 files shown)


gdaviet:
This looks valid, thanks for the PR @azrabano23 ! Can you address `adj_svd2` as well?

azrabano23:
@gdaviet Done in e79c8f2 — `adj_svd2` now has a matching repeated-singular-value branch. The generic formula's `1/(s2^2 - s1^2)` factor collapses to the `-1/epsilon` clamp exactly on these inputs (amplifying the skew terms by ~1e6), so the branch instead differentiates the factorization the forward actually returns — `sigma = |A|_F/sqrt(2)`, `U = A/sigma`, `V = Id` — i.e. the exact vector-Jacobian product of the forward code path. `adj_V` is ignored (V is constant in the branch) and `A = 0` contributes nothing.

Verified on CPU (float64):
- gradient of `sigma[0] + sigma[1]` equals `U = A/sigma` **exactly** (0.0 max error) and matches central finite differences to ~2e-10 for rot(90°), 2·rot(0.5), −I, and diag(1,−1) — central differences are meaningful for the sigma sum because the eigenvalue split is antisymmetric in the perturbation and cancels;
- gradient of a weighted reconstruction `sum(W ∘ U·diag(sigma)·V^T)` equals **exactly** `W` (the branch reconstructs A exactly, so the VJP must be the identity — it is, to 0.0);
- new test `test_svd_2D_repeated_grad` covers both invariants across all four degenerate inputs with tape-vs-analytic and tape-vs-central-FD assertions (float32/float64; individual U/V entries are deliberately not differentiated — they are gauge at a repeated singular value). Full `test_mat.py`: 72 tests, OK, CPU.

On the greptile float16 note (false-zero discriminant admitting a not-exactly-orthogonal U): for such inputs the reconstruction `U·diag(sigma)·V^T = A` is still *exact* by construction, and U's deviation from orthogonality is the same order as the rounding that produced the false zero — versus the previous behavior (`U = Id`) which was O(1) wrong in both respects. An exactly-orthogonal U for those inputs would trade away the exact reconstruction; happy to add a normalize-and-complete step if you'd prefer that trade.

warp/tests/matrix/test_mat.py
```diff
@@ -1009,6 +1015,116 @@ def check_mat_svd2(
                 assert_np_equal((plusval - minusval) / (2 * dx), m2grads[ii, jj], tol=fdtol)
 
 
+def test_svd_2D_repeated_grad(test, device, dtype, register_kernels=False):
+    """Gradients on the repeated-singular-value branch (GH-1734).
```
shi-eric: Please remove the GH-<NUM> comment from the test description. The commit message carries the association to the issue.

warp/native/svd.h
```diff
@@ -656,6 +671,26 @@ inline CUDA_CALLABLE void adj_svd2(
 {
     const Type epsilon = _svd_config<Type>::SVD_EPSILON;
 
+    if (sigma[0] == sigma[1]) {
```
shi-eric: There appears to be a CUDA-only gradient issue in the new repeated-singular-value handling.

The float32 and float64 CUDA variants of `test_svd_2D_repeated_grad` fail on the scaled-rotation case, while the corresponding CPU tests pass. This reproduces on CUDA 13.0 with an NVIDIA RTX PRO 6000 Blackwell Server Edition.

On CUDA, `_svd_2` gets a tiny nonzero discriminant and takes the general path, so `V` is not identity. The singular values still both round to `2.0`, and `adj_svd2` uses `sigma[0] == sigma[1]` to select the repeated-value branch. That branch assumes `V = Id`, so the gradient of `sigma[0] + sigma[1]` comes back as `U` instead of `U V^T`.

Could `adj_svd2` use the same discriminant check as `_svd_2` to keep the forward and adjoint on the same path?

warp/native/svd.h
```diff
@@ -656,6 +671,26 @@ inline CUDA_CALLABLE void adj_svd2(
...
+        // Repeated singular value: _svd_2's duplicate-eigenvalue branch returned the
```
shi-eric: Could this comment be shortened? It walks through the full derivative and repeats details expressed by the code below, which makes the implementation harder to scan. The useful context is the branch-specific factorization, why the generic VJP cannot be used, and the zero-input caveat. For example:

```cpp
// _svd_2's duplicate-eigenvalue branch sets sigma = ||A||_F / sqrt(2),
// U = A / sigma, and V = Id. Differentiate that factorization directly
// because the generic VJP is singular for repeated singular values.
// V is constant, and A = 0 has no well-defined direction.
```

changelog/1734.fixed.md
```diff
@@ -0,0 +1,2 @@
+Fix `wp.svd2()` returning identity singular vectors for scaled orthogonal
+inputs such as rotations, reflections, and negative multiples of the identity.
```
shi-eric: Could this be reworded? `identity singular vectors` is not standard terminology and makes it sound like both outputs are wrong. `V = Id` is valid here. The bug is that `U` also comes back as `Id`.

Maybe:

> Fix `wp.svd2()` returning an incorrect `U` matrix for scaled orthogonal inputs, including rotations, reflections, and negative multiples of the identity.

azrabano23:
@shi-eric Thank you for running this on real CUDA hardware and for the precise diagnosis — all four points addressed in 1b3d467:

- **Predicate mismatch (the CUDA failure):** `adj_svd2` now recomputes the discriminant with exactly `_svd_2`'s arithmetic and branches on `discriminant == 0`, so forward and adjoint always agree on the path taken on any device. `sigma[0] == sigma[1]` alone was indeed wrong for general-path inputs whose sigmas round equal (your scaled-rotation case).
- **Test robustness:** the kernel now emits which path ran (`V = Id` with equal sigmas identifies the duplicate-eigenvalue branch), and `test_svd_2D_repeated_grad` asserts the branch-exact invariants only when the branch was taken on the current device; the finite-difference check on the singular-value sum applies on either path (it equals `U` on the branch and `U Vᵀ` on the general path, both matching central differences).
- **Comment:** replaced with your suggested wording (plus one line on why the predicate is recomputed).
- **Changelog:** reworded as suggested; docstring's issue reference removed.

CPU verification after the change: full `test_mat.py` 72 tests OK; the branch invariants still hold exactly (gradient of the sigma sum equals `U` to 0.0, reconstruction VJP equals the weights to 0.0, both matching central FD to ~2e-10, float64). I don't have an NVIDIA GPU locally, so I'd appreciate a re-run of the float32/float64 CUDA variants when convenient — on CUDA the scaled-rotation input should now exercise the general path in both forward and adjoint and pass the FD check.

warp/native/svd.h
```diff
@@ -656,6 +671,30 @@ inline CUDA_CALLABLE void adj_svd2(
...
+        // _svd_2's duplicate-eigenvalue branch sets sigma = |A|_F / sqrt(2),
+        // U = A / sigma, and V = Id. Differentiate that factorization directly
+        // because the generic VJP is singular for repeated singular values.
+        // V is constant, and A = 0 has no well-defined direction.
+        if (sigma[0] != Type(0)) {
+            Type inv_sigma = Type(1) / sigma[0];
+            Type adj_sigma_sum_half = Type(0.5) * (adj_sigma[0] + adj_sigma[1]);
+            Type u_dot_adj_u = U.data[0][0] * adj_U.data[0][0] + U.data[0][1] * adj_U.data[0][1]
+                + U.data[1][0] * adj_U.data[1][0] + U.data[1][1] * adj_U.data[1][1];
+            adj_A = adj_A + (adj_sigma_sum_half - Type(0.5) * inv_sigma * u_dot_adj_u) * U + inv_sigma * adj_U;
```
shi-eric: The new repeated-value adjoint gives the wrong gradient when a loss uses `U` and `V` together. This reproduces on CPU with float32 and float64:

```text
A = [[0, -1], [1, 0]]
loss = (U @ V.T)[0, 1]

tape gradient      = [[0, 0.5], [0.5, 0]]
central difference ≈ [[0, 0],   [0, 0]]
```

`U` and `V` are not individually unique when the singular values repeat, but their changes cancel in `U * transpose(V)`. Holding `V` fixed drops that cancellation.

This comes from the branch-specific adjoint added in this PR, so it should be fixed here. Could the branch account for both `adj_U` and `adj_V` and add a CPU regression test for this case?

warp/native/svd.h
```diff
@@ -520,12 +520,27 @@ inline CUDA_CALLABLE void _svd_2( // input A
...
+        v22 = Type(1);
+        if (s1 == Type(0)) {
+            // A = 0
+            u11 = Type(1);
+            u12 = Type(0);
+            u21 = Type(0);
+            u22 = Type(1);
+        } else {
+            // A / s is exactly orthogonal, so U = A / s, V = Id reconstructs A
+            Type inv_sigma = Type(1) / s1;
```
shi-eric: The discriminant check can send a non-orthogonal float16 matrix through this branch because both squared terms underflow to zero. This reproduces on CPU:

```text
A = [[0.005, 0.002],
     [0,     0]]

sigma   = [0.003813, 0.003813]  # expected ≈ [0.005386, 0]
U.T @ U = [[1.72, 0.69],
           [0.69, 0.28]]
```

The underflow was already present, but this PR changes what the branch does. It now computes `U = A / sigma`, which gives non-orthogonal singular vectors for this input. This case needs to be handled as part of the new branch.

A local test confirmed that scaling `A` by its largest absolute entry before forming `A.T @ A` avoids the underflow. The repeated-value check can then use `diff == 0 && ATA12 == 0`, with `sigma` rescaled afterward. The adjoint needs to use the same normalized check.

That change passed this repro and all 18 existing SVD2 tests on CPU and CUDA. Could you use this approach and add the CPU float16 case as a regression test?

warp/native/svd.h
azrabano23: Fixed in 03990a7 — you're right that holding V fixed broke the U–V cancellation. The branch adjoint now backpropagates through the gauge-invariant combinations instead of the U = A/sigma chart: adj_A += U·G·Vᵀ with G = [[adj_sigma[0], g], [−g, adj_sigma[1]]] and g = (ku − kv)/(4·sigma), where ku/kv are the skew components of UᵀadjU and VᵀadjV. For losses of the polar factor U·Vᵀ this reduces to the exact polar-factor VJP Q·skew(Qᵀ·adjU)/sigma; for sigma losses it is unchanged.

Your reproducer now gives tape gradient exactly 0 (max |entry| = 0.0), matching central differences (~1e-10). A weighted polar loss sum(W ∘ U·Vᵀ) matches the analytic (W − Q·Wᵀ·Q)/(2·sigma) to ≤4e-17 and central FD to ~9e-11 across all four degenerate inputs, and the sigma-sum checks are unchanged. `test_svd_2D_repeated_grad` now uses the polar loss as its second gauge-invariant output with these assertions.

One honest caveat: no single linear adjoint can serve every loss family at a repeated singular value (losses of U·Vᵀ and of U·diag(sigma)·Vᵀ produce identical (adjU, adj_sigma, adjV) tuples for some weight choices yet have different true gradients — the factor map itself is non-differentiable there). This convention privileges the polar factor and the sigmas, which I believe are the combinations simulation code actually differentiates; the pure-gauge response is dropped. Happy to adjust if you'd prefer a different convention.

warp/native/svd.h
azrabano23: Fixed in 03990a7 with the approach you hinted at: `_svd_2` now scales the input by its largest magnitude entry before forming AᵀA (U and V are scale-invariant; the singular values are rescaled at the end), so the squared terms cannot underflow into a false-zero discriminant. `adj_svd2` recomputes the predicate with the identical scaled arithmetic so forward and adjoint still agree on the path.

Your input now takes the general path in float16: sigma = [0.005386, 0] (vs the expected ≈[0.005386, 0]) with |UᵀU − I| = 0.0. The matrix is added to `test_svd_2D`'s edge-case list, where the existing orthogonality and reconstruction assertions cover it across all dtypes and devices. The A = 0 case is handled before scaling (returns U = V = Id, sigma = 0, and contributes no gradient). Full `test_mat.py`: 72 tests, OK, CPU — CUDA re-verification appreciated when you get a chance.

warp/native/svd.h
```diff
@@ -656,6 +685,46 @@ inline CUDA_CALLABLE void adj_svd2(
...
+        // built on U V^T, U diag(sigma) V^T (up to the eigenvalue-splitting
+        // direction, which cancels in central differences), and the sigmas;
+        // the response of the pure-gauge direction is dropped.
+        Type inv_four_sigma = Type(0.25) / sigma[0];
+        Type ku = U.data[0][0] * adj_U.data[0][1] + U.data[1][0] * adj_U.data[1][1] - adj_U.data[0][0] * U.data[0][1]
+            - adj_U.data[1][0] * U.data[1][1];
+        Type kv = V.data[0][0] * adj_V.data[0][1] + V.data[1][0] * adj_V.data[1][1] - adj_V.data[0][0] * V.data[0][1]
+            - adj_V.data[1][0] * V.data[1][1];
+        Type g = (ku - kv) * inv_four_sigma;
+        mat_t<2, 2, Type> G = mat_t<2, 2, Type>(adj_sigma[0], g, -g, adj_sigma[1]);
```
shi-eric: The repeated-value adjoint still drops part of the gradient for the smooth reconstruction `R = U diag(sigma) V^T`. Because this branch reconstructs `R = A` exactly, a loss such as `R[0,1]` must produce `adj_A = [[0, 1], [0, 0]]`. With `A = I`, the current code instead returns `[[0, 0.5], [-0.5, 0]]`. Please preserve the full reconstruction VJP in this branch and add an off-diagonal reconstruction-loss regression test.

warp/tests/matrix/test_mat.py
```diff
@@ -1009,6 +1019,144 @@ def check_mat_svd2(
...
+    M = len(mats)
+    weights = rng.standard_normal((M, 2, 2))
+
+    m2 = wp.array(mats, dtype=mat22, requires_grad=True, device=device)
+    w = wp.array(weights, dtype=mat22, device=device)
+    outcomponents = wp.zeros(3 * M, dtype=wptype, requires_grad=True, device=device)
+    out = wp.zeros(1, dtype=wptype, requires_grad=True, device=device)
+
+    analytic_tol = 1.0e-6 if dtype == np.float64 else 1.0e-5
+    fdtol = 5.0e-4 if dtype == np.float64 else 2.0e-2
```
shi-eric: These tolerances are much looser than the observed error and could let substantial adjoint errors pass. On current CPU and CUDA runs, the maximum analytic error was `4.7e-8` for float32 and `5.6e-17` for float64; the maximum central-difference error was `2.9e-4` and `1.1e-7`, respectively. Could you tighten `analytic_tol` to about `1e-6`/`1e-12` and `fdtol` to about `1e-3`/`1e-6` for float32/float64?

warp/tests/matrix/test_mat.py
```diff
@@ -1009,6 +1019,144 @@ def check_mat_svd2(
                 assert_np_equal((plusval - minusval) / (2 * dx), m2grads[ii, jj], tol=fdtol)
 
 
+def test_svd_2D_repeated_grad(test, device, dtype, register_kernels=False):
+    """Gradients on the repeated-singular-value branch.
```
shi-eric: Could you reduce this docstring to the test contract and leave the branch-detection detail next to `on_branch`? Most of the current text narrates the implementation. For example:

```python
"""Check gauge-invariant gradients at repeated singular values.

Individual ``U`` and ``V`` are not differentiable there, so this test covers
the singular-value sum and polar factor ``U V^T``.
"""
```

A short comment above `on_branch` can explain why branch detection is needed.

warp/native/svd.h
shi-eric: No need to update this separately before fixing the VJP, but when revising this branch, could you also shorten the replacement comment? The current block narrates finite-difference behavior and lists supported loss constructions, and the reconstruction claim is currently incorrect. Please keep the updated comment focused on why the special case is needed and the mathematical convention it implements.

shi-eric:
Once the review feedback is addressed, please rebase this branch onto the current `main` and squash the four incremental commits into one signed-off commit before marking the PR ready to merge. Doing this as the final cleanup will avoid repeated force pushes during review.

warp/native/svd.h
azrabano23: Fixed in b1ffdc9. The branch now recovers the reconstruction cotangent symmetrically from both factor adjoints — C = (adj_U + U·adj_Vᵀ)/(2σ), then removes the part already carried by adj_sigma — instead of differentiating the U = A/σ chart with V held fixed. Your case now returns exactly [[0, 1], [0, 0]] for R[0,1] at A = I (regression added at the identity with those exact weights), the weighted reconstruction VJP equals the weights to ≤7e-18 with central-FD agreement ~1e-10 across all branch inputs, and your earlier (U·Vᵀ)[0,1]-at-rotation case still returns exactly 0.

One thing worth stating openly for the record: reconstruction losses and *general* polar-factor losses cannot both be exact at a repeated singular value — at A = I, the loss <C, U·diag(σ)·Vᵀ> with C = e01 and the loss <C', U·Vᵀ> with C' = σ·e01 hand the adjoint the byte-identical tuple (adj_U = e01, adj_σ = 0, adj_V = e10) while their true gradients differ ([[0,1],[0,0]] vs [[0,0.5],[−0.5,0]]). This convention makes reconstruction and sigma losses exact everywhere and polar losses exact wherever the tuples are distinguishable (which includes your rotation example); the residual polar cases inherit the reconstruction convention. The in-code comment states the convention without narrating it.

Also noting the branch state per your process request: the rebase/squash onto main is in (e719819), and this review response sits as one commit on top so the delta is reviewable — I'll fold it into the single signed-off commit as the final cleanup once you're satisfied with the convention.

warp/tests/matrix/test_mat.py
azrabano23: Tightened in b1ffdc9 to your measured envelopes: analytic_tol 1e-6 / 1e-12 and fdtol 1e-3 / 1e-6 for float32 / float64. Locally (CPU) the observed max analytic error is 7e-18 (float64) and the max central-difference error ~6e-10, so there is comfortable margin under the new bounds.

warp/tests/matrix/test_mat.py
azrabano23: Done in b1ffdc9 — docstring reduced to the contract (your wording, with the reconstruction as the second invariant), and the branch-detection rationale moved to a short comment above on_branch.

warp/native/svd.h
azrabano23: Shortened in b1ffdc9 along with the VJP revision — the comment now states the branch factorization, why a convention is required, how the cotangent is recovered, and nothing else.

warp/native/svd.h
shi-eric: The normalization added in `03990a7` fixes the original squared-term underflow, but it introduces a second float16 failure. For nonzero inputs with `amax` below about `1.53e-5`, `Type(1) / amax` overflows to infinity, so the subsequent multiplications produce non-finite values. After rebuilding the current head, `A = [[1e-5, 0], [0, 5e-6]]` returns all-NaN `U`, `sigma`, and `V` on both CPU and CUDA. `adj_svd2` repeats the same normalization. Please use direct division by `amax` or a wider intermediate in both paths and add a float16 regression case near `1e-5`.

azrabano23: Fixed in 4be6a41 — thank you for catching this. Both `_svd_2` and `adj_svd2` now divide by `amax` directly instead of multiplying by `Type(1) / amax`, so the reciprocal can no longer overflow for tiny inputs. On CPU in float16, `A = [[1e-5, 0], [0, 5e-6]]` now returns `sigma = [1.0014e-5, 5.0068e-6]` (the float16 roundings of the inputs) with `U = V = Id` and exact reconstruction, and the smallest float16 subnormal `[[6e-8, 0], [0, 0]]` also comes through finite. That matrix is added to `test_svd_2D`'s edge cases as a float16 regression next to the underflow one, so the forward orthogonality and reconstruction checks cover both ends.

Per your process note, the branch is now rebased onto current `main` (d4de134) and squashed into a single signed-off commit. CPU `test_mat.py`: 72 tests pass across float16/32/64. I still can't run the CUDA variants locally, so a re-run there would be appreciated.

PR #1838, opened 2026-08-24
Probe Windows USD TLS loading

shi-eric:
## Description

Diagnose the frequent native USD loading crash seen with usd-core 26.8 on
the Windows RTX Pro 6000 runner. This draft intentionally replaces normal
PR validation with a resource-scoped experiment and is not intended to be
merged. It follows the Windows USD failure tracked in #1828.

## Changes

- Build only the Windows CUDA 13 artifact needed by the target runner.
- Serialize four diagnostic scenarios with matrix fail-fast disabled.
- Compare the 25.5.1 modular Sdf load with 26.8 early and late Tf loads.
- Sweep live thread counts after Warp, PyTorch, and CUDA initialization.
- Compare full 26.8 suites with lazy and eager Tf loading.
- Capture module maps, native thread counts, exit statuses, and WER dumps.

## Checklist

- [x] The documentation is up to date with these changes. (Not applicable to this temporary diagnostic.)
- [x] I added a changelog fragment if this change affects users. (No user-facing change.)

## Validation summary

- Verified eight focused unittest cases, including native Windows module
  enumeration, failed-child preservation, eager import ordering, and workflow
  routing.
- Verified both workflows with actionlint and ran pre-commit on all five files.
- Completed a standard Windows CUDA build with Warp 1.18.0.dev0.
- Exercised 25.5.1 late Sdf, 26.8 early Tf, 26.8 late Tf, and a 16-thread
  late-Tf case on a local NVIDIA L4. All completed, as expected from the prior
  local investigation.
- Confirmed the logs distinguish usd_tf.dll then usd_sdf.dll in 25.5.1 from
  usd_ms.dll at Tf import in 26.8. The held-thread case recorded 13 native
  threads after CUDA warmup and 29 immediately before the USD load.

The lazy full-suite scenario may fail by design. Artifacts upload with an
always condition, and fail-fast is disabled so every serial scenario runs.

## Summary by CodeRabbit

* **New Features**
  * Added Windows diagnostics for USD loading across CUDA, Warp, PyTorch, and MSVC runtime configurations.
  * Added eager USD initialization checks with structured failure diagnostics.
  * Added downloadable JUnit, coverage, and diagnostic log artifacts.
  * Added validation for CUDA 13.0.3 and USD 26.8 environments.

* **Tests**
  * Expanded validation for Windows imports, child processes, module filtering, and workflow configuration.

* **Chores**
  * Streamlined pull-request checks around the Windows USD diagnostic workflow.


shi-eric:
### Investigation result: MSVC runtime resolution, not TLS exhaustion

The controlled runs now isolate the failure to an incompatible Visual C++ runtime selected by the Windows loader.

Evidence:

- In [run 32778595982](https://github.com/NVIDIA/warp/actions/runs/32778595982), usd-core 26.8 failed on the first Tf import before Warp, Torch, or CUDA loaded, with 7 native threads. The same wheel passed 16/16 when imported after Warp, Torch, and CUDA, and passed the 0/16/64/128 held-thread sweep. usd-core 25.5.1 failed 16/16 after Torch. This is the opposite of the proposed late static-TLS-exhaustion behavior.
- The lazy full suite reproduced the original first USD failure and downstream 1114 errors. Eager Tf import failed in roughly two seconds before suite startup, proving accumulated suite state is unnecessary.
- In [run 32783233585](https://github.com/NVIDIA/warp/actions/runs/32783233585), plain import torch protected the later Tf import; CUDA initialization was unnecessary. Explicitly loading the environment MSVC runtime also protected it. Importing Blosc or Warp did not.
- The final minimal experiment, [run 32785453083](https://github.com/NVIDIA/warp/actions/runs/32785453083), installed only usd-core 26.8 and Blosc 1.11.4. Direct Tf import reproduced the access violation and DLL initialization failure 2/2. Every intervention that exposed the environment MSVCP140 passed 2/2:
  - preload only msvcp140.dll
  - preload msvcp140_atomic_wait.dll, which also loads MSVCP140
  - add the environment bin directory with os.add_dll_directory
  - add that directory to PXR_USD_WINDOWS_DLL_PATH
  - preload the complete environment runtime set

Runtime versions on the failing runner:

- System32 msvcp140.dll and vcruntime140.dll: 14.00.24215.1
- Environment runtime supplied by Blosc: 14.44.35211.0
- usd-core 26.8 usd_ms.dll linker version: 14.44
- usd-core 25.5.1 libraries linker version: 14.29

Microsoft requires the deployed Redistributable to be at least as new as the newest toolset used by any component: https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170

The usd-core Windows package initializes PXR_USD_WINDOWS_DLL_PATH with only the pxr directory. The Tf import wrapper chooses that variable before PATH and registers only its listed directories. Consequently, the compatible runtime in environment/bin is not considered. Torch happens to load a compatible MSVCP140 first, which explains why late imports can succeed.

Conclusion: moving roughly 279 KB of TLS into usd_ms.dll in 26.8 is a real layout change, but it is not the cause of this failure. ERROR_DLL_INIT_FAILED is the loader symptom, not evidence of static-TLS exhaustion.

Likely remedies:

1. Update the runner Visual C++ Redistributable to a version at least as new as the wheel toolset.
2. In the usd-core wheel/import wrapper, include the directory containing its compatible runtime dependencies in PXR_USD_WINDOWS_DLL_PATH, or package and resolve those dependencies explicitly.
3. In Warp tests, replace the find_spec-only USD availability check with an isolated real native import smoke test so an unusable installation is detected before the suite.

The red diagnostic check is expected: it records the failing direct control while still running and uploading all protected cases. This PR should remain draft and must not be merged because it intentionally replaces normal PR CI.

shi-eric:
Final Windows runner result: [run 32793130981](https://github.com/NVIDIA/warp/actions/runs/32793130981) passed.

- The real `usd-core==26.8` `pxr.Tf` smoke test passed.
- The full RTX Pro 6000 suite passed: 8,748 tests in 700.434 seconds, with 71 skipped.
- The suite log contains zero `_tf` loader failures.

The first implementation repaired the path only in `unittest_utils`. That fixed the main test process, but 19 example subprocesses still failed because each fresh process imported `pxr`, which overwrote the inherited `PXR_USD_WINDOWS_DLL_PATH` again.

The passing variant runs a silent `sitecustomize` hook in every test interpreter. When `pxr` is installed, the hook imports `pxr`, appends the active environment's `bin` and `Library\bin` directories to the overwritten value, and then imports `Tf`. It does nothing in isolated build interpreters where USD is absent and fails fast only when USD is installed but unusable.

This demonstrates that Warp can test against 26.8 on the existing runner without the 25.5.1 pin, provided the repair is applied in every Python process. It remains a workaround for the underlying usd-core loader behavior.

shi-eric:
Closing this diagnostic PR now that it has served its purpose. The runs confirmed the OpenUSD DLL search-path issue and showed that a process-start workaround can make usd-core 26.8 pass, but Warp does not need that added CI complexity. Windows CI can remain pinned to usd-core 25.5.1 while the upstream OpenUSD issue is tracked.

PR #1850, opened 2026-08-25
Extend wp.func to support an inline option

### LINKED ISSUE #1849: Add inlining control to @wp.func
### Description
When calling a function from a kernel, in some cases there are benefits from inlining but in other cases it can be more optimal to not inline (due to e.g. register usage/occupancy and instruction cache behaviour in large kernels). Often this needs manual control to get the best setting. The addition of explicit inlining control to Warp (via "inline=true" and "inline=false" func options) will allow users to specify the best option for their specific case. There is precedent for this via @wp.kernel, which already exposes e.g. launch_bounds. 

### Context
Cases have been observed when using warp for rendering and simulation associated with RL training, where modifying the codegen to explicitly disable inlining has given significant performance benefits. (Given the symmetry, it also makes sense to enable forceinline as part of the change.)

agray3:
## Description

Add `@wp.func(inline=...)` to control whether a Warp function is emitted out of line or inlined
into its call sites. `None` leaves the choice to the backend compiler, `False` keeps the function
out of line, and `True` requires inlining. Inlining a large function into many call sites raises a
kernel's peak register count and duplicates its body, so `inline=False` can improve occupancy and
instruction-cache behavior in large kernels; `inline=True` overrides the backend compiler's
heuristic in the other direction.

The decorator resolves the option to a single hint stored on the `Function`, which codegen maps to
the `WP_NOINLINE` / `WP_FORCEINLINE` macros defined in both module headers. The hint covers the
forward function, its adjoint, and any custom gradient or replay hooks, and feeds
`ModuleHasher.hash_function()` so it forms part of the kernel cache key. Warp compiles device code
with `WP_NO_CRT`, so the toolkit's `host_defines.h` is unavailable and the macros expand to the
underlying attributes rather than `__noinline__` / `__forceinline__`.

Closes #1849.

## Changes

- `warp/_src/context.py` — decorator keyword argument, validation, the `inline_hint` field on
  `Function`, the cache-key update, and builder plumbing.
- `warp/_src/codegen.py` — macros in both module headers, an `{inline_attr}` slot in all four
  function templates (CPU/CUDA × forward/reverse), and the hint-to-macro mapping.
- Tests: new `warp/tests/test_func_inline.py`, plus `TestModuleHasherFunctionOptions` in
  `warp/tests/test_module_hashing.py` next to the existing `TestModuleHasherKernelOptions`.
- Docs: new "Function Settings" section in `docs/user_guide/configuration.rst`.
- Changelog fragment.

## Checklist

## Validation summary

Verified on an L40S (sm_89) with the full suite (`-s autodetect`, 5710 tests), compared against
the same run on unmodified `main`. The two failure sets differ only in tests that need a CPU
compiler backend unavailable on this machine, plus the new tests themselves, which fail on `main`
because the option does not exist there.

- `test_noinline_func_runs` launches a kernel calling a function marked `inline=False` on the GPU
  and checks the results.
- Two PTX tests compile ahead of time and check each direction in the optimization mode where it
  is observable: `inline=False` in a release build, where the compiler would otherwise inline the
  helper by choice, and `inline=True` in a debug build, where it would otherwise inline nothing.
- `TestFuncInline` pins the emitted attribute for both backends, including on the generated
  adjoint and on custom gradient and replay hooks;
  `TestModuleHasherFunctionOptions.test_inline_hint_hashed` pins the cache key.
- Red/green: defining `WP_NOINLINE` as nothing fails the release test, dropping the `inline`
  specifier from `WP_FORCEINLINE` fails the debug test, and dropping the `hash_function()` update
  collapses the three variants to one hash.
- Not exercised locally: CPU compilation and the `_MSC_VER` branch (both are covered by CI).

## New feature / enhancement

```python
import warp as wp


@wp.func(inline=False)
def integrate_contact(x: wp.vec3, v: wp.vec3, n: wp.vec3, dt: float) -> wp.vec3:
    vn = wp.dot(v, n)
    if vn < 0.0:
        v = v - 2.0 * vn * n
    return x + v * dt


@wp.kernel(enable_backward=False)
def step(x: wp.array[wp.vec3], v: wp.array[wp.vec3], dt: float):
    tid = wp.tid()
    p = x[tid]
    p = integrate_contact(p, v[tid], wp.vec3(0.0, 1.0, 0.0), dt)
    p = integrate_contact(p, v[tid], wp.vec3(1.0, 0.0, 0.0), dt)
    p = integrate_contact(p, v[tid], wp.vec3(0.0, 0.0, 1.0), dt)
    x[tid] = p
```

`integrate_contact` is called from three sites in this kernel. Without the hint the backend
compiler is free to inline it at each one, so its register footprint and code size count three
times toward the kernel; with `inline=False` it is emitted once and called. The generated CUDA
source contains a single definition, and the same attribute reaches the adjoint:

```cpp
static WP_NOINLINE CUDA_CALLABLE wp::vec_t<3, wp::float32> integrate_contact_0(...)
static WP_NOINLINE CUDA_CALLABLE void adj_integrate_contact_0(...)
```

`inline=True` applies the opposite hint, for a small helper the compiler chose to outline.

## Summary by CodeRabbit

* **New Features**
  * Replaced separate `noinline` and `forceinline` options with a unified `inline` setting for `@wp.func`.
  * Supports backend defaults, forced inlining, and explicitly disabling inlining.
  * Applies consistently to forward and generated adjoint functions on CPU and CUDA.

* **Documentation**
  * Added guidance covering inline behavior, configuration, examples, and performance considerations.

* **Tests**
  * Added coverage for validation, execution, code generation, derivatives, native functions, and deterministic module hashing.


warp/_src/context.py
```diff
@@ -1282,6 +1284,8 @@ def func(
     *,
     name: str | None = None,
     module: Module | Literal["unique"] | str | None = None,
+    noinline: bool = False,
+    forceinline: bool = False,
```
shi-eric: I think it makes more sense to have a single parameter named `inline` instead of two mutually exclusive booleans. What do you think of `inline: bool | None = None`:

- `None`: Leave choice to backend
- `True`: Require inlining
- `False`: Do not inline

agray3: I guess a slight downside of this is that it moves the syntax further away from that used in CUDA (so less familiar for CUDA programmers), but overall yes I agree it is more elegant and appropriate so I'll go ahead and make that change.

warp/_src/codegen.py
```diff
@@ -6741,6 +6750,12 @@ def get_references(adj) -> tuple[dict[str, Any], dict[Any, Any], dict[warp._src.
 #define WP_ENABLE_SMEM_SPILLING()
 #endif
 
+// Inline control for @wp.func(noinline=True) / @wp.func(forceinline=True). Warp compiles
+// device code with WP_NO_CRT, so host_defines.h, which normally defines __noinline__ and
+// __forceinline__, is not included; spell both out as the attributes NVRTC and NVCC accept.
+#define WP_NOINLINE __attribute__((noinline))
+#define WP_FORCEINLINE __attribute__((always_inline))
```
shi-eric: `WP_FORCEINLINE` is missing the `inline` specifier here. CUDA defines `__forceinline__` as `__inline__ __attribute__((always_inline))`, while this replacement only includes the attribute.

I reproduced this with CUDA 12.9.86 and CUDA 13.0 in debug (`-G -O0`) builds. Both compilers ignored the attribute and emitted the forced helper as a PTX `.func` called through `call.uni`, so `forceinline=True` was not honored.

This is not limited to debug builds. Debug mode is simply the most deterministic way to expose the problem because it disables heuristic inlining. A release build may hide it by choosing to inline the function independently.

Could you change this to:

```cpp
#define WP_FORCEINLINE inline __attribute__((always_inline))
```

You may also want the same definition in the non-MSVC CPU branch for consistency, although Warp's embedded Clang currently honors the bare attribute there.

warp/tests/test_func_inline.py
```diff
@@ -0,0 +1,141 @@
...
+    def test_invalid_inline_hint_rejected(self):
+        """Reject non-bool values for the inline hint."""
+        for value in (1, 1.5, "yes"):
+            with self.subTest(value=value), self.assertRaisesRegex(TypeError, "inline must be a bool"):
+
+                @wp.func(inline=value, module="unique")
+                def invalid(x: float) -> float:
+                    return x
+
+    def test_codegen_emits_inline_attrs(self):
```
shi-eric, line 50: The current tests do not verify that the compiler actually honors these inline hints. Could you add a regression test that compiles to PTX and checks for the expected function definitions and calls?

This can stay self-contained by using `compile_aot_module()` with a temporary `module_dir`. In debug mode, the `noinline` helper should remain a `.func` with a call, while the `forceinline` helper should have neither.

warp/_src/context.py
agray3: Now done.

warp/tests/test_func_inline.py
agray3: I've added 2 such tests: a release test for inline=False and a debug test for inline=True. The problem with a single debug test is that nothing is automatically inlined in debug mode anyway, so inline=False would pass whether or not the attribute is honoured.

warp/_src/codegen.py
agray3: Good spot! Now fixed (including non-MSVC CPU branch).

changelog/1849.added.md
shi-eric, line 1: The fragment should be renamed to reference the issue number https://github.com/NVIDIA/warp/issues/1849

warp/tests/test_module_hashing.py
```diff
@@ -414,6 +416,32 @@ def k(a: wp.array[int]):
         self.assertEqual(make(32).module.hash_module(), make(32).module.hash_module())
 
 
+class TestModuleHasherFunctionOptions(unittest.TestCase):
```
shi-eric: Since there is only one test method, you just add this into the existing `TestModuleHashing` class?

docs/user_guide/configuration.rst
```diff
@@ -264,3 +264,58 @@ Declaring ``cluster_dim`` only forms the cluster; *using* it — distributed
...
+        # inlined even where the backend compiler would not choose to
+        return x * 2.0
+
+Left unset, inlining is decided by the backend compiler, per function and per call site, with
+no way to influence it from Python. Inlining a large function into many call sites raises a
+kernel's peak register count, because register allocation is bounded by the worst path through
+the whole kernel body, and duplicates the body at every call site, pressuring the instruction
+cache. ``inline=False`` keeps one copy out of line at the cost of an ABI call;
+``inline=True`` is the reverse, for small helpers the compiler chose to outline. Either
+direction is workload dependent and worth measuring.
```
shi-eric: Could we simplify this paragraph? The second half packs several separate ideas into two sentences: the default behavior, register pressure, instruction-cache use, and the tradeoffs of both settings. Connecting all of that with several comma clauses and a semicolon makes the main point difficult to pick out.

I suggest separating what the settings do from why someone might choose them. For example:

```rst
By default, the backend compiler decides whether to inline a function at each call site. Set ``inline=False`` to prevent inlining or ``inline=True`` to require it.

Inlining duplicates the function body at each call site. This can increase register pressure and instruction-cache use. Keeping a function out of line avoids that duplication but adds function-call overhead. The best choice depends on the workload, so measure both options.
```

I am not attached to this exact wording. My main suggestion is to use shorter sentences and keep each sentence focused on one part of the tradeoff.

changelog/1849.added.md
agray3: Done

warp/tests/test_module_hashing.py
agray3: Done

docs/user_guide/configuration.rst
agray3: Done - I'm happy with your text, and noted for next time.

shi-eric:
Thanks, @agray3! I think this is good to go, can you open a merge request on the internal GitLab repo?

agray3:
> Thanks, @agray3! I think this is good to go, can you open a merge request on the internal GitLab repo?

Great! I don't have permission to this internal repo ATM, please can you add me as a dev?

shi-eric:
> > Thanks, @agray3! I think this is good to go, can you open a merge request on the internal GitLab repo?
> 
> Great! I don't have permission to this internal repo ATM, please can you add me as a dev?

Done.

PR #1865, opened 2026-08-27
Derive tile warp-scan constants from WP_TILE_WARP_SIZE

### LINKED ISSUE #1936: Kernel array slicing fails to compile under wp.config.llvm_cuda: byte_offset_helper is host-only
# Bug Description

Slicing an array inside a kernel fails to compile when `wp.config.llvm_cuda` is
enabled. The kernel below compiles and runs under the default NVRTC path, but is
rejected by the Clang-based CUDA compiler.

`view()` in `warp/native/array.h` is `CUDA_CALLABLE` and is what a kernel-side
array slice lowers to. It calls `byte_offset_helper()`, which is declared without
`CUDA_CALLABLE` and is therefore host-only:

https://github.com/NVIDIA/warp/blob/main/warp/native/array.h#L831

NVCC tolerates the host-from-device call; Clang does not, so the failure is
specific to the `llvm_cuda` path. Adding `CUDA_CALLABLE` to that one declaration
fixes it. No behavior changes on the NVRTC path — the annotation only widens
where the function may be called from.

## Reproducer

Requires a build that includes `warp-clang.so` (i.e. **not** `--no-standalone`).
`wp.config.llvm_cuda` dispatches to `runtime.llvm.wp_compile_cuda`, so without
that library the setting is silently inert and the bug does not reproduce.

```python
import numpy as np
import warp as wp

wp.config.llvm_cuda = True


@wp.kernel
def slice_kernel(src: wp.array2d[float], out: wp.array[float]):
    i = wp.tid()
    view = src[i : i + 2, 1]
    out[i] = view[0] + view[1]


src = wp.array(np.arange(15, dtype=np.float32).reshape(5, 3), dtype=float, device="cuda:0")
out = wp.zeros(4, dtype=float, device="cuda:0")
wp.launch(slice_kernel, dim=4, inputs=[src, out], device="cuda:0")
print(out.numpy())
```

Expected: `[ 5.  11.  17.  23.]`

Actual, on `main` at `6465652a53e76937bd21cc3e1ea05ce6ddf3fa33` (clean tree):

```
In file included from .../warp/native/builtin.h:2835:
warp/native/array.h:862:21: error: reference to __host__ function 'byte_offset_helper<float, 0ULL, 1ULL>' in __host__ __device__ function
  862 |     size_t offset = byte_offset_helper(src, slices, make_index_sequence<N> {});
      |                     ^
.../wp___main___c5edd41.cu:100:21: note: called by 'slice_kernel_923dd757_cuda_kernel_forward'
  100 |         var_6 = wp::view(var_src, var_3, var_5);
      |                     ^
warp/native/array.h:831:8: note: 'byte_offset_helper<float, 0ULL, 1ULL>' declared here
  831 | size_t byte_offset_helper(array_t<T>& src, const slice_t (&slices)[sizeof...(Idxs)], index_sequence<Idxs...>)
      |        ^
```

The same error repeats for `slice_kernel_..._cuda_kernel_backward`, then:

```
2 errors generated when compiling for sm_75.
Module __main__ c5edd41 load on device 'cuda:0' took 6463.52 ms  (error)
CUDA kernel build failed with error code -1
```

With `CUDA_CALLABLE` added to `byte_offset_helper()` on the same tree, the same
script prints the expected values. The kernel cache was cleared between the two
runs so a stale artifact could not mask either outcome.

# System Information

```
GPU:      NVIDIA H100 80GB HBM3
Driver:   590.48.01
CUDA:     13.1
Warp:     1.18.0.dev3, built from source at 6465652a53e76937bd21cc3e1ea05ce6ddf3fa33
          (unmodified NVIDIA/warp main, `git status --porcelain` empty)
OS:       Linux
```

zhihuidu-amd:
## Description

Two related changes to the tile execution contract in `warp/native/`.

The first is a bug fix. `byte_offset_helper()` in `array.h` was declared without
`CUDA_CALLABLE`, making it host-only, but it is called from `view()`, which is
`CUDA_CALLABLE` and is what a kernel-side array slice lowers to. NVCC tolerates
the host-only call; the Clang-based CUDA compiler selected by
`wp.config.llvm_cuda` rejects it, so slicing an array inside a kernel fails to
compile on that path. Closes #1936.

The second is a refactor with no change in behavior. `tile_radix_sort.h` and
`tile_scan.h` hardcoded the warp width in several forms — `32` as a loop bound
and as a divisor, `0xffffffffu` as a ballot mask, and `lane == 31` to select the
lane holding a scan total — while `WP_TILE_WARP_SIZE` was already defined beside
them and meant exactly the same thing. The divisor is the one that matters on its
own: callers size shared memory as
`(WP_TILE_BLOCK_DIM + WP_TILE_WARP_SIZE - 1) / WP_TILE_WARP_SIZE`, so
`thread_id / 32` is a second, independent definition of the same quantity, and a
disagreement between them is a silent out-of-bounds shared-memory write rather
than a compile error. All values are unchanged: the mask is still 32 bits and the
intrinsics are the ones the call sites already invoked directly.

The coupling became visible while porting Warp to HIP/ROCm, where a wavefront is
64 lanes. That work is not proposed here — this change is CUDA-only.

## Changes

- `array.h`: mark `byte_offset_helper()` as `CUDA_CALLABLE`.
- `tile.h`: define the tile execution contract — `WP_TILE_WARP_SIZE`, the lane
  mask type `wp_tile_lane_mask_bits_t`, `WP_TILE_LANE_MASK_ALL`,
  `WP_TILE_LANE_MASK_BELOW`, `WP_TILE_LANE_MASK_POPC`, `WP_TILE_LANE_MASK_FFS` —
  above the `builtin.h` include, since `tile_reduce.h`, `tile_scan.h`,
  `tile_radix_sort.h` and `sparse.cu` all consume it.
- `tile_scan.h`, `tile_reduce.h`, `sparse.cu`: use those names instead of `32`,
  `0xffffffffu`, `lane == 31`, `__popc` and `__ffs`.
- `tile_radix_sort.h`: name the bitonic shuffle constants
  `WP_TILE_BITONIC_GROUP_SIZE` / `WP_TILE_BITONIC_GROUP_MASK` for their real
  scope. These are deliberately **not** `WP_TILE_WARP_SIZE`: the bitonic
  exchanges are bounded to a fixed 32-lane group by their stride, so tying them
  to the warp width would be wrong on a wider target.
- `warp/tests/cuda/test_clang_cuda.py`: add `test_slice_kernel`, which slices an
  array inside a kernel under `wp.config.llvm_cuda`. No other test in that file
  slices, so without it the annotation could be removed unnoticed.

## Checklist

## Validation summary

Validated on an **NVIDIA H100 80GB HBM3** (driver 590.48.01, CUDA 13.1), built
from source with `warp-clang.so` present, since `wp.config.llvm_cuda` dispatches
to `runtime.llvm.wp_compile_cuda` and is silently unavailable without it.

Red/green for the bug fix, on unmodified `main` at
`6465652a53e76937bd21cc3e1ea05ce6ddf3fa33` (`git status --porcelain` empty):

- Verified the reproducer below **fails** on `main` without this change:

  ```
  warp/native/array.h:862:21: error: reference to __host__ function
      'byte_offset_helper<float, 0ULL, 1ULL>' in __host__ __device__ function
  .../wp___main___c5edd41.cu:100:21: note: called by
      'slice_kernel_923dd757_cuda_kernel_forward'
  2 errors generated when compiling for sm_75.
  CUDA kernel build failed with error code -1
  ```

- Verified it **passes** with the one-line `CUDA_CALLABLE` annotation applied to
  that same tree (`git diff --stat` = 1 file, 1 insertion, 1 deletion), printing
  `[ 5. 11. 17. 23.]`. Both runs used the same build; the kernel cache was
  cleared between them so a stale artifact could not mask either outcome.

Full output is in #1936.

`test_slice_kernel` in `warp/tests/cuda/test_clang_cuda.py` is that reproducer as
a regression test.

For the refactor, this branch was compared against unmodified `main` with an
identical run, selecting `TestTileReduce`, `TestTileSort`, `TestTileScan` and
`TestArray`: both arms report **Ran 127 — ok=125, ERROR=1, FAIL=0**, identical
including one pre-existing failure present in both. Header placement was checked
separately by swapping the `tile_reduce.h` and `tile_scan.h` include lines in
`builtin.h`: on the pre-image that still builds and then fails at first kernel
launch, where NVRTC instantiates the templates —

```
warp/native/tile_scan.h(61): error: identifier "WP_TILE_FULL_WARP_MASK" is undefined
warp/native/tile_scan.h(60): error: identifier "WP_TILE_WARP_SIZE" is undefined
```

— while with the definitions in `tile.h` the same swap builds and passes. Each
phase ran on a cold kernel cache.

The branch also builds and passes tile-sort correctness checks on gfx942.

## Bug fix

Fails on `main`, passes with this PR. Requires a build that includes
`warp-clang.so`.

```python
import numpy as np
import warp as wp

wp.config.llvm_cuda = True


@wp.kernel
def slice_kernel(src: wp.array2d[float], out: wp.array[float]):
    i = wp.tid()
    view = src[i : i + 2, 1]
    out[i] = view[0] + view[1]


src = wp.array(np.arange(15, dtype=np.float32).reshape(5, 3), dtype=float, device="cuda:0")
out = wp.zeros(4, dtype=float, device="cuda:0")
wp.launch(slice_kernel, dim=4, inputs=[src, out], device="cuda:0")
print(out.numpy())
```

### DIFF
--- a/warp/native/array.h
+++ b/warp/native/array.h
@@ -828,7 +828,8 @@ CUDA_CALLABLE inline bool view_arg_is_slice(const slice_t&) { return true; }
 
 
 template <typename T, size_t... Idxs>
-size_t byte_offset_helper(array_t<T>& src, const slice_t (&slices)[sizeof...(Idxs)], index_sequence<Idxs...>)
+inline CUDA_CALLABLE size_t
+byte_offset_helper(array_t<T>& src, const slice_t (&slices)[sizeof...(Idxs)], index_sequence<Idxs...>)
 {
     return byte_offset(src, slices[Idxs].start...);
 }
--- a/warp/native/tile.h
+++ b/warp/native/tile.h
@@ -3,6 +3,20 @@
 
 #pragma once
 
+// Internal lane-participation masks for tile warp collectives, not logical
+// tile masks. Keep this block before builtin.h because its tile headers use it.
+#define WP_TILE_WARP_SIZE 32
+
+#define WP_TILE_LANE_MASK_ALL 0xffffffffu
+#define WP_TILE_LANE_MASK_POPC(mask) __popc(mask)
+#define WP_TILE_LANE_MASK_FFS(mask) __ffs(mask)
+
+using wp_tile_lane_mask_bits_t = decltype(WP_TILE_LANE_MASK_ALL);
+
+// Excludes `lane` itself.
+#define WP_TILE_LANE_MASK_BELOW(lane)                                                                                  \
+    ((((wp_tile_lane_mask_bits_t)1) << (lane)) - ((wp_tile_lane_mask_bits_t)1))
+
 #include "builtin.h"
 
 #include "rand.h"
--- a/warp/native/tile_radix_sort.h
+++ b/warp/native/tile_radix_sort.h
@@ -19,6 +19,13 @@ namespace wp {
 // The radix sort in this file is consistently slower than the bitonic sort
 #define BITONIC_SORT_THRESHOLD 2048
 
+// Bitonic shuffles use fixed 32-lane groups; cross-group exchanges use shared
+// memory, so every shuffle below has a stride of at most 16 and stays inside its
+// group. The overloads below are not full-warp collectives when WP_TILE_WARP_SIZE
+// differs from WP_TILE_BITONIC_GROUP_SIZE.
+#define WP_TILE_BITONIC_GROUP_SIZE 32
+#define WP_TILE_BITONIC_GROUP_MASK 0xffffffffu
+
 struct UintKeyToUint {
     inline CUDA_CALLABLE uint32_t convert(uint32 value) { return value; }
 
@@ -102,7 +109,7 @@ constexpr inline CUDA_CALLABLE int next_higher_pow2(int input)
 inline CUDA_CALLABLE half warp_shuffle_xor(half val, int lane_mask)
 {
     unsigned int bits = static_cast<unsigned int>(val.u);
-    bits = __shfl_xor_sync(0xFFFFFFFFu, bits, lane_mask);
+    bits = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, bits, lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
 
     half result;
     result.u = static_cast<unsigned short>(bits);
@@ -113,7 +120,7 @@ inline CUDA_CALLABLE half warp_shuffle_xor(half val, int lane_mask)
 inline CUDA_CALLABLE bfloat16 warp_shuffle_xor(bfloat16 val, int lane_mask)
 {
     unsigned int bits = static_cast<unsigned int>(val.u);
-    bits = __shfl_xor_sync(0xFFFFFFFFu, bits, lane_mask);
+    bits = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, bits, lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
 
     bfloat16 result;
     result.u = static_cast<unsigned short>(bits);
@@ -141,7 +148,7 @@ template <typename T> inline CUDA_CALLABLE T warp_shuffle_xor(T val, int lane_ma
 
     WP_PRAGMA_UNROLL
     for (int i = 0; i < word_count; ++i) {
-        output[i] = __shfl_xor_sync(0xFFFFFFFFu, input[i], lane_mask);
+        output[i] = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, input[i], lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
     }
 
     return *reinterpret_cast<T*>(output);
@@ -153,7 +160,7 @@ inline CUDA_CALLABLE wp::vec_t<Length, T> warp_shuffle_xor(wp::vec_t<Length, T> 
     wp::vec_t<Length, T> result;
 
     for (unsigned i = 0; i < Length; ++i)
-        result[i] = __shfl_xor_sync(0xFFFFFFFFu, val[i], lane_mask);
+        result[i] = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, val[i], lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
 
     return result;
 }
@@ -189,7 +196,8 @@ inline CUDA_CALLABLE wp::mat_t<Rows, Cols, T> warp_shuffle_xor(wp::mat_t<Rows, C
 
     for (unsigned i = 0; i < Rows; ++i)
         for (unsigned j = 0; j < Cols; ++j)
-            result.data[i][j] = __shfl_xor_sync(0xFFFFFFFFu, val.data[i][j], lane_mask);
+            result.data[i][j]
+                = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, val.data[i][j], lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
 
     return result;
 }
@@ -226,8 +234,8 @@ template <typename T> inline CUDA_CALLABLE T* warp_shuffle_xor(T* val, int lane_
     unsigned long long ptr = reinterpret_cast<unsigned long long>(val);
     unsigned int ptr_lo = static_cast<unsigned int>(ptr);
     unsigned int ptr_hi = static_cast<unsigned int>(ptr >> 32);
-    ptr_lo = __shfl_xor_sync(0xFFFFFFFFu, ptr_lo, lane_mask);
-    ptr_hi = __shfl_xor_sync(0xFFFFFFFFu, ptr_hi, lane_mask);
+    ptr_lo = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, ptr_lo, lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
+    ptr_hi = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, ptr_hi, lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
     ptr = (static_cast<unsigned long long>(ptr_hi) << 32) | static_cast<unsigned long long>(ptr_lo);
     return reinterpret_cast<T*>(ptr);
 }
@@ -237,7 +245,8 @@ inline CUDA_CALLABLE wp::shape_t warp_shuffle_xor(wp::shape_t val, int lane_mask
     wp::shape_t result;
 
     for (int i = 0; i < wp::ARRAY_MAX_DIMS; ++i)
-        result.dims[i] = __shfl_xor_sync(0xFFFFFFFFu, val.dims[i], lane_mask);
+        result.dims[i]
+            = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, val.dims[i], lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
 
     return result;
 }
@@ -250,9 +259,14 @@ template <typename T> inline CUDA_CALLABLE wp::array_t<T> warp_shuffle_xor(wp::a
     result.grad = wp::warp_shuffle_xor(val.grad, lane_mask);
     result.shape = wp::warp_shuffle_xor(val.shape, lane_mask);
     for (int i = 0; i < wp::ARRAY_MAX_DIMS; ++i)
-        result.strides[i] = __shfl_xor_sync(0xFFFFFFFFu, val.strides[i], lane_mask);
-    result.ndim = static_cast<uint16_t>(__shfl_xor_sync(0xFFFFFFFFu, static_cast<unsigned int>(val.ndim), lane_mask));
-    result.flags = static_cast<uint16_t>(__shfl_xor_sync(0xFFFFFFFFu, static_cast<unsigned int>(val.flags), lane_mask));
+        result.strides[i]
+            = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, val.strides[i], lane_mask, WP_TILE_BITONIC_GROUP_SIZE);
+    result.ndim = static_cast<uint16_t>(__shfl_xor_sync(
+        WP_TILE_BITONIC_GROUP_MASK, static_cast<unsigned int>(val.ndim), lane_mask, WP_TILE_BITONIC_GROUP_SIZE
+    ));
+    result.flags = static_cast<uint16_t>(__shfl_xor_sync(
+        WP_TILE_BITONIC_GROUP_MASK, static_cast<unsigned int>(val.flags), lane_mask, WP_TILE_BITONIC_GROUP_SIZE
+    ));
 
     return result;
 }
@@ -329,11 +343,10 @@ inline CUDA_CALLABLE void bitonic_sort_single_stage_full_thread_block(
     __syncthreads();
 }
 
-// stride can be 1, 2, 4, 8, 16
 template <typename K, typename V>
 inline CUDA_CALLABLE void bitonic_sort_single_stage_full_warp(int k, unsigned int thread_id, int stride, K& key, V& val)
 {
-    auto s_key = __shfl_xor_sync(0xFFFFFFFFu, key, stride);
+    auto s_key = __shfl_xor_sync(WP_TILE_BITONIC_GROUP_MASK, key, stride, WP_TILE_BITONIC_GROUP_SIZE);
     auto s_val = warp_shuffle_xor(val, stride);
     auto swap = (((thread_id & stride) != 0 ? key > s_key : key < s_key)) ^ ((thread_id & k) == 0);
     key = swap ? s_key : key;
@@ -341,12 +354,12 @@ inline CUDA_CALLABLE void bitonic_sort_single_stage_full_warp(int k, unsigned in
 }
 
 
-// Sorts 32 elements according to keys
+// Sorts one 32-lane subwarp's worth of elements according to keys
 template <typename K, typename V>
 inline CUDA_CALLABLE void bitonic_sort_single_warp(unsigned int thread_id, K& key, V& val)
 {
 #pragma unroll
-    for (int k = 2; k <= 32; k <<= 1) {
+    for (int k = 2; k <= WP_TILE_BITONIC_GROUP_SIZE; k <<= 1) {
 #pragma unroll
         for (int stride = k / 2; stride > 0; stride >>= 1) {
             bitonic_sort_single_stage_full_warp(k, thread_id, stride, key, val);
@@ -469,9 +482,9 @@ template <int max_num_elements, typename K, typename V, typename KeyToUint>
 inline CUDA_CALLABLE void
 bitonic_sort_thread_block_shared_mem(int thread_id, K* keys_input, V* values_input, int num_elements_to_sort)
 {
-    if constexpr (max_num_elements < 32) {
+    if constexpr (max_num_elements < WP_TILE_BITONIC_GROUP_SIZE) {
         // Fast track - single warp sort
-        if (thread_id < 32)
+        if (thread_id < WP_TILE_BITONIC_GROUP_SIZE)
             bitonic_sort_single_warp<K, V, KeyToUint>(thread_id, keys_input, values_input, num_elements_to_sort);
         __syncthreads();
     } else {
@@ -568,9 +581,9 @@ template <int max_num_elements, typename K, typename V, typename KeyToUint>
 inline CUDA_CALLABLE void
 bitonic_sort_thread_block_direct(int thread_id, K* keys_input, V* values_input, int num_elements_to_sort)
 {
-    if constexpr (max_num_elements < 32) {
+    if constexpr (max_num_elements < WP_TILE_BITONIC_GROUP_SIZE) {
         // Fast track - single warp sort
-        if (thread_id < 32)
+        if (thread_id < WP_TILE_BITONIC_GROUP_SIZE)
             bitonic_sort_single_warp<K, V, KeyToUint>(thread_id, keys_input, values_input, num_elements_to_sort);
         __syncthreads();
     } else {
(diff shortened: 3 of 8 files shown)


yaoliu13:
@coderabbitai help

yaoliu13:
@coderabbitai review

shi-eric:
This PR describes the `byte_offset_helper` change as a user-facing bug fix and includes a `.fixed` changelog fragment. Please restore the applicable PR template sections and checklist, file and link a GitHub issue containing a minimal `warp.config.llvm_cuda` reproducer, rename the fragment to `<issue-number>.fixed.md`, and add a regression test that exercises the reported Clang failure. `CONTRIBUTING.md` asks contributors to report bugs in an issue first, use the PR template, and cover bug fixes with tests.

If this is only internal preparation for future HIP/ROCm work rather than a current user-facing bug, please remove the changelog fragment instead.

warp/native/tile_radix_sort.h
```diff
@@ -676,8 +687,14 @@ inline CUDA_CALLABLE void radix_sort_thread_block_core(
...
     int num_bits_to_sort = 32;  // Sort all bits because that's what the bitonic fast pass does as well
 
-    const int warp_id = thread_id / 32;
-    const int lane_id = thread_id & 31;
+    // WP_TILE_WARP_SIZE, not a hardcoded 32. The callers size shared_mem as
+    // num_warps = (WP_TILE_BLOCK_DIM + WP_TILE_WARP_SIZE - 1) / WP_TILE_WARP_SIZE,
+    // so a literal here is a second, independent definition of the warp width.
+    // If the two ever disagree the warp index can exceed num_warps and write
+    // past the end of shared_mem -- silently, with no compile error and
+    // no launch failure.
```
shi-eric: Please reduce this to the invariant a future reader needs: `shared_mem` is allocated with one row per tile warp, so its index uses `WP_TILE_WARP_SIZE`. The discussion of the former literal and hypothetical failure mode narrates the diff rather than documenting the resulting code.

warp/native/tile_reduce.h
```diff
@@ -13,6 +13,21 @@
 
 #define WP_TILE_WARP_SIZE 32
 
+// Companions to WP_TILE_WARP_SIZE, so that code indexing lanes or building lane
+// masks derives everything from one place rather than repeating the literals.
+// These are the CUDA values and change nothing: the mask is still 32 bits wide
+// and __popc/__ffs are the same intrinsics the call sites used directly.
```
shi-eric: Some of the new comments only make sense while reading the diff. Phrases like `change nothing`, `the same intrinsics the call sites used directly`, `rather than a bare unsigned int`, and `rather than a literal 32` describe what changed, but they will be confusing once this is merged. Please remove them or rewrite them to explain a constraint in the current code. The comparison with the old code belongs in the PR description or commit message.

warp/native/array.h
```diff
@@ -827,8 +827,11 @@ CUDA_CALLABLE inline bool view_arg_is_slice(int) { return false; }
 CUDA_CALLABLE inline bool view_arg_is_slice(const slice_t&) { return true; }
 
 
+// CUDA_CALLABLE: called from byte_offset below, which is device-reachable
+// through array slicing in kernels.
```
shi-eric: This comment has the call direction backwards: `byte_offset_helper()` calls `byte_offset()`, while its device-reachable caller is `view()`. `CUDA_CALLABLE` already expresses the requirement, so please remove the comment rather than leaving an incorrect explanation.

warp/native/tile_radix_sort.h
```diff
@@ -713,15 +730,15 @@ inline CUDA_CALLABLE void radix_sort_thread_block_core(
                 bool contributes = digit == b;
                 int sum_per_warp = warp_scan_inclusive(lane_id, 0xFFFFFFFF, contributes);
```
shi-eric: `0xFFFFFFFF` here is a full-warp mask, so leaving it as a literal bypasses the new `WP_TILE_FULL_WARP_MASK` abstraction. The same literal remains in the second radix pass and in several `__ballot_sync()` calls in `tile_reduce.h`. Please use the named constant for the full-warp masks covered by this refactor. Unrelated uses, such as the maximum 32-bit key value, should remain literals.

warp/native/array.h
```diff
@@ -827,8 +827,11 @@ CUDA_CALLABLE inline bool view_arg_is_slice(int) { return false; }
...
 template <typename T, size_t... Idxs>
-size_t byte_offset_helper(array_t<T>& src, const slice_t (&slices)[sizeof...(Idxs)], index_sequence<Idxs...>)
+inline CUDA_CALLABLE size_t
```
shi-eric, line 831: The current Clang-CUDA tests do not exercise array slicing, so this regression could return unnoticed. I reproduced the failure with `wp.config.llvm_cuda = True` and a kernel containing `view = array[start:stop, 1]`; without `CUDA_CALLABLE`, Clang rejects the call to the host-only `byte_offset_helper()`. Please add a regression test to `warp/tests/cuda/test_clang_cuda.py` that launches a slicing kernel and checks its result. The test should fail without this annotation and pass with it.

warp/native/tile_reduce.h
```diff
@@ -13,6 +13,20 @@
...
 
+// Companions to WP_TILE_WARP_SIZE: one bit per lane of a tile warp, and the
+// intrinsics that operate on such a mask. Defined together so a change to the
+// warp width has a single place to touch.
+#define WP_TILE_FULL_WARP_MASK 0xffffffffu
+#define WP_TILE_POPC(mask) __popc(mask)
+#define WP_TILE_FFS(mask) __ffs(mask)
+
+// Type wide enough to hold one bit per lane.
+using wp_tile_warp_mask_t = decltype(WP_TILE_FULL_WARP_MASK);
```
shi-eric, line 24: The new mask type does not reach tile reductions, so a mask wider than 32 bits would still be truncated. Every `warp_shuffle_down()` overload takes `int mask`, the reduction helpers and ballot locals use `unsigned int`, and the partial-mask checks shift an untyped `1`. The three first-active-lane lookups also call `__ffs()` directly, leaving `WP_TILE_FFS` unused. Please use `wp_tile_warp_mask_t` for these mask parameters and locals, type the lane-bit operations through it, and use `WP_TILE_FFS` for the first-active-lane lookups.

warp/native/tile_radix_sort.h
```diff
@@ -333,7 +337,7 @@ inline CUDA_CALLABLE void bitonic_sort_single_stage_full_thread_block(
 template <typename K, typename V>
 inline CUDA_CALLABLE void bitonic_sort_single_stage_full_warp(int k, unsigned int thread_id, int stride, K& key, V& val)
 {
-    auto s_key = __shfl_xor_sync(0xFFFFFFFFu, key, stride);
+    auto s_key = __shfl_xor_sync(WP_TILE_FULL_WARP_MASK, key, stride);
```
shi-eric: The full-warp mask now follows `WP_TILE_WARP_SIZE`, but the lanes participating in this bitonic path do not. `bitonic_sort_single_warp()` stops at `k <= 32`, and both fast paths require `max_num_elements < 32` and guard the call with `thread_id < 32`. On a 64-lane target, the shuffle mask would name lanes that never enter the collective, and the sort would run only the 32-lane stages. Please derive the stage bound and participation guards from `WP_TILE_WARP_SIZE`, or keep this as an explicit 32-lane subwarp and pass a matching mask.

warp/native/tile_reduce.h
```diff
@@ -13,6 +13,20 @@
 
 #define WP_TILE_WARP_SIZE 32
 
+// Companions to WP_TILE_WARP_SIZE: one bit per lane of a tile warp, and the
+// intrinsics that operate on such a mask. Defined together so a change to the
+// warp width has a single place to touch.
+#define WP_TILE_FULL_WARP_MASK 0xffffffffu
+#define WP_TILE_POPC(mask) __popc(mask)
```
shi-eric, line 20: `bsr_compress_select_sorted_runs()` would use different widths for lane indexing and its ballot mask after `WP_TILE_WARP_SIZE` changes. The function already derives its lane, warp, terminal lane, and shuffle width from `WP_TILE_WARP_SIZE`, but it keeps a local `unsigned int full_warp_mask = 0xffffffffu`, stores the ballot in `unsigned int`, calls `__popc()` directly, and computes the prefix with `(1u << lane) - 1u`. Please use `wp_tile_warp_mask_t`, `WP_TILE_FULL_WARP_MASK`, `WP_TILE_POPC`, and `WP_TILE_LANES_BELOW` here so the indexing and mask operations use the same width.

warp/native/tile_radix_sort.h
```diff
@@ -333,20 +353,20 @@ inline CUDA_CALLABLE void bitonic_sort_single_stage_full_thread_block(
 template <typename K, typename V>
 inline CUDA_CALLABLE void bitonic_sort_single_stage_full_warp(int k, unsigned int thread_id, int stride, K& key, V& val)
 {
-    auto s_key = __shfl_xor_sync(0xFFFFFFFFu, key, stride);
+    auto s_key = __shfl_xor_sync(WP_TILE_BITONIC_SUBWARP_MASK, key, stride, WP_TILE_BITONIC_SUBWARP_SIZE);
```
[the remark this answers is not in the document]
zhihuidu-amd: Thanks — this found a real problem, though not the one described. Two corrections, then what I changed.

**The premise is off.** Line 356 does not use `WP_TILE_FULL_WARP_MASK`. This PR changed it to a dedicated 32-lane constant with an explicit width argument; the review reads as though it were against the pre-image. The mask *value* at this call site is also unchanged by this PR — `main` already passes `0xffffffffu` here, so nothing was narrowed.

**The caller analysis is right, and I had it wrong.** `bitonic_sort_single_stage_full_warp()` has two callers, not one:

- `bitonic_sort_single_warp()`, guarded by `thread_id < 32`
- `bitonic_sort_pow2_length()` (line 446), which enters from every thread in the block

So my comment on the constants — "only threads below lane 32 enter it" — was false, and the `SUBWARP` name overclaimed.

**What bounds the exchange is the stride, not the lane id.** Both callers issue `stride <= 16`: the single-group loop runs `k` up to 32, so `stride = k/2 <= 16`, and `bitonic_sort_pow2_length()` gates this call on `if (stride <= 16)` and routes anything larger through shared memory. Since `lane ^ stride` with `stride <= 16` never crosses a 32-aligned boundary, the width-32 shuffle selects the same source lane a full-width shuffle would — for both callers. On CUDA it is a no-op regardless, as 32 == `warpSize`.

Pushed 8d079b8, which renames the constants to `WP_TILE_BITONIC_GROUP_{SIZE,MASK}` and replaces the comment with that invariant, stated at the shuffle where a reader would look for it. Macro values unchanged.

On wave64 more broadly: a 32-bit participation mask on a 64-lane wave is a real gap, but it is pre-existing throughout this file and out of scope here. This PR only names the existing constants at their uses, per @shi-eric's review; it is not a wave64 enablement change and should not be read as one.

Verified on an H100 (sm_90) built from this branch at 8d079b8: `test_tile_sort`, `test_tile_reduce`, `test_tile`, `test_tile_load`, `test_sparse` — 152 tests, all pass.

daedalus5:
A design concern here: the new helpers are internal execution/lane masks for warp collectives, not logical tile masks. Since Warp may eventually want user-facing tile masks, I think we should avoid names that read like a tile-mask data type.

In particular, `wp_tile_warp_mask_t` looks a bit too much like a future `wp_tile_mask_t` / `tile_mask` abstraction. I would prefer names that emphasize lane-bit/sync-mask semantics, for example:

- `wp_tile_lane_mask_bits_t`
- `WP_TILE_LANE_MASK_ALL`
- `WP_TILE_LANE_MASK_BELOW(lane)`
- `WP_TILE_LANE_MASK_POPC(mask)`
- `WP_TILE_LANE_MASK_FFS(mask)`

Also, these helpers are now used outside reductions (`tile_scan.h`, `tile_radix_sort.h`, and `sparse.cu`), so defining them in `tile_reduce.h` makes the dependency rely on include order. I think they belong in `tile.h`, next to `WP_TILE_WARP_SIZE`, as part of the common tile execution contract.

The goal is to keep these clearly scoped as warp-lane participation masks for collectives, while reserving `tile_mask`-style names for any future user-facing, shape-aware logical tile mask API.

zhihuidu-amd:
Both changes are in 34c13c0. Thanks — the second one turned out to be a runtime bug, not only a tidiness issue.

**Naming.** Renamed exactly as proposed: `wp_tile_lane_mask_bits_t`, `WP_TILE_LANE_MASK_ALL`, `WP_TILE_LANE_MASK_BELOW`, `WP_TILE_LANE_MASK_POPC`, `WP_TILE_LANE_MASK_FFS`. Agreed on reserving `tile_mask`-style names for a shape-aware API; the comment on the block now says in as many words that these are lane-participation masks for collectives, not logical tile masks.

**Location, with one correction.** `WP_TILE_WARP_SIZE` was not in `tile.h` — it was at `tile_reduce.h:14`, immediately above the helpers you flagged, so it carries the identical dependency. Moving only the helpers would have left it behind. I moved the whole block, `WP_TILE_WARP_SIZE` included.

**The dependency is worse than relying on include order.** Swapping `builtin.h`'s `tile_reduce.h` and `tile_scan.h` lines on the pre-image still *builds*: the ahead-of-time compile only parses those templates, so nothing is diagnosed. It then fails at the first kernel launch, where NVRTC instantiates them:

```
Warp NVRTC compilation error 6: NVRTC_ERROR_COMPILATION
warp/native/tile_scan.h(61): error: identifier "WP_TILE_FULL_WARP_MASK" is undefined
warp/native/tile_scan.h(60): error: identifier "WP_TILE_WARP_SIZE" is undefined
6 errors detected
```

With the block in `tile.h`, the same swap builds and passes.

**One placement detail.** The block sits *above* `tile.h`'s own `#include "builtin.h"`. `builtin.h` includes `tile.h`, so the two form a cycle: a translation unit entering through `tile.h` suspends it at line 6, runs all of `builtin.h` — `tile_reduce.h` among it — and only then resumes `tile.h`'s body. Placed below that include, the definitions would arrive too late for exactly the headers this is meant to fix.

Verified on an H100 (sm_90) built from this branch at 34c13c0:

- `test_tile_sort`, `test_tile_reduce`, `test_tile`, `test_tile_load`, `test_sparse` — 152 tests, all pass.
- Same tree with the two include lines swapped: builds, and both tile modules pass.
- 8d079b8 with the same swap: the NVRTC failure quoted above. Each phase ran on a cold kernel cache, so a stale cubin could not mask it.

Macro values are unchanged.

warp/native/tile.h
```diff
@@ -3,6 +3,30 @@
 
 #pragma once
 
+// The tile execution contract. WP_TILE_WARP_SIZE is the lane width of a tile warp, and
```
shi-eric: This block makes a small set of definitions harder to scan by repeating what their names already say. The useful constraints are that these are internal lane-participation masks, not logical tile masks, and that the definitions must appear before `builtin.h` because tile headers included from it depend on them. Please reduce the block and the per-definition comments to those constraints. For example:

```cpp
// Internal lane-participation masks for tile warp collectives, not logical
// tile masks. Keep this block before builtin.h because its tile headers use it.
```

The only per-definition note that may still help is a concise statement that `WP_TILE_LANE_MASK_BELOW(lane)` excludes `lane` itself.

warp/native/tile_radix_sort.h
```diff
@@ -19,6 +19,16 @@ namespace wp {
 // The radix sort in this file is consistently slower than the bitonic sort
 #define BITONIC_SORT_THRESHOLD 2048
 
+// The bitonic sort exchanges elements only within a fixed 32-lane group, independently
```
shi-eric: Repeating the fixed 32-lane-group invariant in three places makes this code harder to scan and gives future changes three explanations to keep synchronized. The blocks at lines 22-28, 109-111, and 353-356 all describe the same group boundary and shuffle constraint. Please consolidate that into one short comment beside `WP_TILE_BITONIC_GROUP_SIZE`, then remove the repeated overload-level and function-level explanations. For example:

```cpp
// Bitonic shuffles use fixed 32-lane groups; cross-group exchanges use shared
// memory. The overloads below are not full-warp collectives when
// WP_TILE_WARP_SIZE differs from WP_TILE_BITONIC_GROUP_SIZE.
```

shi-eric:
Thanks, @zhihuidu-amd. I've resolved my earlier inline review threads now that the requested code changes are present. I left two minor follow-up comments about keeping the new header comments concise.

Separately, this PR is still blocked because the repository contribution requirements called out above have not been addressed. Following these guidelines keeps review and integration work manageable for maintainers, which helps us continue accepting community contributions. Please complete the following before requesting another review:

- File and link a GitHub issue for the user-facing `byte_offset_helper` bug, with a minimal `warp.config.llvm_cuda` reproducer.
- Rewrite the PR description using the current template, including the applicable checklist, validation summary, and bug-fix reproducer.
- Rename `changelog/+tile-warp-width-constants.fixed.md` to `<issue-number>.fixed.md`.
- Clean up the commit history as required by `CONTRIBUTING.md`. The current five commits mostly record incremental review fixes, so squashing them into one clean commit is appropriate. Reference the issue as `(GH-<number>)` in the commit message.
- Keep the branch rebased onto the current `main` until it is ready to merge.

zhihuidu-amd:
Thanks @shi-eric. All five items are addressed.

- **Issue filed:** #1936, with a minimal `wp.config.llvm_cuda` reproducer. It is
  backed by a red/green run on an H100 against unmodified `main` at
  `6465652a53e76937bd21cc3e1ea05ce6ddf3fa33`, not by reading the source: the
  issue quotes the actual Clang diagnostic, and the same tree with only the
  `CUDA_CALLABLE` annotation applied compiles and returns the expected values.
  Note the build must include `warp-clang.so` — `llvm_cuda` dispatches to
  `runtime.llvm.wp_compile_cuda`, so with `--no-standalone` the setting is
  silently inert and the bug does not reproduce.
- **PR description:** rewritten against the current template, with the checklist,
  validation summary and bug-fix reproducer.
- **Changelog:** `+tile-warp-width-constants.fixed.md` → `1936.fixed.md`.
- **History:** the five commits are squashed into one, `Fix host-only
  byte_offset_helper call (GH-1936)`, signed off.
- **Rebase:** onto current `main`; the branch is 1 ahead, 0 behind.

Both comment-concision suggestions are applied. One deviation worth flagging: in
`tile_radix_sort.h` I kept the "stride of at most 16" clause in the consolidated
comment. That bound is the actual correctness argument for why a 32-lane bitonic
group is safe when a block-wide caller enters with thread ids ≥ 32, so dropping
it would have removed the reason and left only the claim. The three repeated
blocks are now one, beside `WP_TILE_BITONIC_GROUP_SIZE`. Happy to trim it
further if you would rather it went.

shi-eric:
Looks good to me @zhihuidu-amd, someone else got a merge into `main`  before I was able to merge your changes, do you mind rebasing with `main` again? Sorry for the trouble

zhihuidu-amd:
No trouble at all — rebased onto `ee46f9913`. No conflicts, and the diff is unchanged at 8 files, +140 / -77; new head is `0a155515f`. DCO passes.

The force-push cleared your approval, so it will need one more whenever you get a chance.

shi-eric:
Great, this is now merged @zhihuidu-amd. Thanks for the fix!

PR #1877, opened 2026-08-30
Fix Python slicing of rectangular matrices

Om-singhaI:
## Description

Slicing a rectangular matrix type at Python scope gives the wrong answer. In `warp/_src/types.py`, `__getitem__` and `__setitem__` resolve a row slice against the column count and a column slice against the row count, so every slice bound is clamped against the wrong axis.

For a `wp.types.matrix(shape=(2, 3), ...)` value, `m[0, :]` returns a `vec2` holding the first two entries of the row, when that row is a `vec3`. Nothing raises, so the dropped element is easy to miss. `m[:, 0]` and `m[:, :]` raise `IndexError`. Slice assignment rejects operands that are in fact the right size, reporting the expected shape transposed.

Square matrix types have the same count on both axes, so the swap cancels out and they behave correctly, which is probably why this went unnoticed. Slicing the same matrices inside a kernel is also correct, since code generation already looks up `_shape_` by dimension.

I did not find an existing issue for this, so there is nothing to close.

## Changes

- `warp/_src/types.py`: resolve each slice against the extent of the axis it indexes. Eight sites, covering the row vector, column vector, and submatrix branches of both `__getitem__` and `__setitem__`.
- `warp/tests/matrix/test_mat_basics.py`: add `test_slicing`, exercising reads, writes, and submatrices on a 2 by 3 and a 3 by 2 matrix across the scalar types the file already parametrizes over.
- `changelog/+fix-rectangular-matrix-slicing.fixed.md`: add a fragment.

## Checklist

## Validation summary

- Reverted `warp/_src/types.py` alone to the parent commit and ran `test_slicing`. All eleven scalar type variants fail, each on the first assertion, reporting `vec2f([1.0, 2.0]) != vec3f([1.0, 2.0, 3.0])` and its equivalent for the other dtypes.
- Restored the file and ran `test_slicing` again. All eleven pass.
- Ran `test_components` in the same file before and after the change. It passes either way, so the neighboring Python paths are undisturbed.
- Compared `__getitem__` and `__setitem__` against NumPy across eight shapes and every pairing of the slice forms `:`, `1:`, `:2`, `0:1`, `::2`, `::-1`, `-2:` and `:-1` on both axes, with positive and negative integer indices. That is 1056 cases in each direction. Before the change 654 reads and 654 writes disagree with NumPy, and after it none do.
- Repeated that comparison restricted to the square shapes `(1, 1)`, `(2, 2)`, `(3, 3)` and `(4, 4)`. Zero mismatches both before and after, which confirms square types keep exactly the behavior they have today.

One caveat on how I ran these. I have no local native build, so instead of `python -m warp.tests` I drove the device free tests through `unittest` directly. `test_slicing` registers with `devices=None` and only touches the Python paths, so it runs in full, but I have not exercised the device backed tests in that file.

## Bug fix

```python
import warp as wp

mat23 = wp.types.matrix(shape=(2, 3), dtype=wp.float32)
m = mat23(1, 2, 3, 4, 5, 6)

print(m[0, :])  # [1.0, 2.0]                    expected [1.0, 2.0, 3.0]
print(m[:, 0])  # IndexError: invalid index     expected [1.0, 4.0]
print(m[:, :])  # IndexError: Invalid row index expected the matrix itself

m[0, :] = wp.vec3f(7, 8, 9)
# RuntimeError: The length of the provided vector (3) isn't compatible
# with the given slice (expected 2)
```

One consequence worth flagging. Assigning a nested Python list to a two dimensional slice, as in `m[:, :] = [[1, 2, 3], [4, 5, 6]]`, raises `TypeError: list indices must be integers or slices, not tuple`. That is not new, a square matrix raises the same thing today, but a rectangular matrix used to fail earlier with the transposed shape message, so this change is what makes the existing error reachable there. NumPy accepts that form. It looks like a separate fix in the assignment loop rather than part of this one, and I am happy to open an issue for it.

The same thing happens with the axes swapped. On a `wp.types.matrix(shape=(3, 2), ...)` it is `n[:, 0]` that silently comes back as a `vec2` instead of a `vec3`, and `n[0, :]` that raises.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected slicing and assignment for rectangular matrices.
  * Row and column slices now return the expected dimensions and values.
  * Fixed errors when slicing columns and truncation when slicing rows.
  * 2D matrix slicing and assignment now use the correct row and column bounds.
  * Existing slicing behavior for square matrices remains unchanged.

* **Tests**
  * Added coverage for slicing and assignment across vector and rectangular matrix types.


Om-singhaI:
`test_components` just above is registered the same way, `devices=None` with the dtype in the name, so I followed the file rather than the `subTest` form used elsewhere in the suite. Happy to switch if you would rather this file moved over.

sylvesterkaczmarek (commented):
Checked the read and write paths together. The swapped extents are the same bug in all eight sites: row slices need `_shape_[1]`, column slices need `_shape_[0]`, and 2D slices need each axis against itself. The rectangular 2x3/3x2 tests catch what square matrices were masking. Looks right.

sylvesterkaczmarek (commented):
Re-reviewed current `d32f5640` after the merge from `main`. The PR-specific diff remains the focused rectangular-matrix slicing fix in `types.py` with matching 2x3/3x2 read/write coverage, and the upstream merge does not introduce a new issue in that change. No blocker from me.

PR #1890, opened 2026-09-02
Add block-Jacobi preconditioner to warp.optim.linear (GH-1891)

### LINKED ISSUE #1234: [REQ] Extend texture interop capabilities

### LINKED ISSUE #1891: Add Block Jacobi Preconditioner for Warp CG Solver
### Description
Warp generates only diagonal (Jacobi) Preconditioners for the conjugate gradient solver. For BSR matrices, using the entire diagonal block in the preconditioner can improve performance

### Context
Preconditioned conjugate gradient is a useful solver for finite element type simulations on the GPU and block preconditioning will improve convergence.

dilevin:
Diagonal (Jacobi) preconditioning ignores intra-block coupling, so block-structured SPD systems (BsrMatrix with block size > 1) converge slower than necessary under CG/CR. Add a "block_jacobi" preconditioner type that inverts each diagonal block via Cholesky factorization (wp.tile_cholesky / tile_cholesky_solve_inplace), generic over block size, falling back to standard Jacobi for 1x1-block (CSR) matrices.

## Checklist

## Bug fix

```python
import warp as wp
# Code that demonstrates the bug
```

## New feature / enhancement

```python
import warp as wp
# Code that demonstrates the new capability
```

## Summary by CodeRabbit

- **New Features**
  - Added Block Jacobi strategies using QR, LDLᵀ, tile-based Cholesky, or automatic selection.
  - Added support for additional floating-point types, square-block BSR matrices, and safe fallback handling for singular or non-positive-definite blocks.
  - Added automatic block-size-based strategy selection, including 1×1 blocks and CPU/GPU tile execution.

- **Documentation**
  - Documented supported strategies, data types, fallback behavior, and square-block BSR matrix usage.

- **Breaking Changes**
  - Removed the legacy `block_jacobi` alias. Use `block_jacobi_auto` or another explicit strategy.

- **Tests**
  - Expanded coverage for strategy selection, validation, data types, concurrency, and rank-deficient edge cases.


changelog/+block-jacobi-preconditioner.added.md
shi-eric, line 1: Hi @dilevin, thanks for the contribution! Can you please file a GitHub issue for this contribution and link to the fragment by renaming it per `changelog/README.md` instructions?

Also, note that NVIDIA employees should open the merge request internally on GitLab at `omniverse/warp`.

But I'm fine leaving this pull request open here as it gives the community insight into what's being reviewed.

dilevin: Thanks Eric, I filed the issue and added the changelog md to this commit. Let me know what else I need to patch up.

warp/_src/optim/linear.py
```diff
@@ -242,17 +242,136 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+    L_flat = wp.empty(shape=A_flat.shape, dtype=scalar_type, device=device)
+    scratch = wp.empty(shape=(dim * block_size,), dtype=scalar_type, device=device)
+
+    factorize_kernel, solve_kernel = _create_block_jacobi_kernels(block_size)
+    wp.launch_tiled(factorize_kernel, dim=[dim], inputs=[A_flat, L_flat], block_dim=32, device=device)
+
+    def block_jacobi_mv(x, y, z, alpha, beta):
+        """Compute ``z = alpha * (blockdiag(A)^-1 @ x) + beta * y``, the block-Jacobi matvec."""
+        xf = _as_scalar_array(x)
+        wp.launch_tiled(solve_kernel, dim=[dim], inputs=[L_flat, xf, scratch], block_dim=32, device=device)
```
shi-eric, line 305: Reusing one block-Jacobi preconditioner across CUDA streams can silently corrupt both outputs. `scratch` is allocated once when the preconditioner is created, so every `matvec` call writes to the same array. One stream can overwrite it before another stream's `_affine_combine_kernel` reads the result.

A focused repro using a 50,000-block 12x12 identity BSR matrix on streams with priorities 0 and -1 produced 438,108 incorrect entries in one output and 19,368 in the other. Both outputs should have matched their inputs exactly.

<details>
<summary>Reproduction</summary>

```python
import numpy as np
import warp as wp
from warp.optim.linear import preconditioner
from warp.sparse import bsr_identity

wp.init()

device = wp.get_device("cuda:0")
num_blocks = 50_000
block_size = 12
num_values = num_blocks * block_size

mat12f = wp.types.matrix(
    shape=(block_size, block_size),
    dtype=wp.float32,
)
A = bsr_identity(
    num_blocks,
    block_type=mat12f,
    device=device,
)
M = preconditioner(A, "block_jacobi")

x0 = wp.full(num_values, 1.0, dtype=wp.float32, device=device)
x1 = wp.full(num_values, 2.0, dtype=wp.float32, device=device)
z0 = wp.zeros_like(x0)
z1 = wp.zeros_like(x1)

stream0 = wp.Stream(device, priority=0)
stream1 = wp.Stream(device, priority=-1)

# Warm up JIT compilation and complete setup work.
M.matvec(x0, z0, z0, alpha=1.0, beta=0.0)
wp.synchronize_device(device)

with wp.ScopedStream(stream0, sync_enter=False, sync_exit=False):
    M.matvec(x0, z0, z0, alpha=1.0, beta=0.0)

with wp.ScopedStream(stream1, sync_enter=False, sync_exit=False):
    M.matvec(x1, z1, z1, alpha=1.0, beta=0.0)

wp.synchronize_device(device)

print(np.count_nonzero(z0.numpy() != 1.0))
print(np.count_nonzero(z1.numpy() != 2.0))
```

Observed on an RTX PRO 6000 Blackwell MIG instance:

```text
438108
19368
```

</details>

This is a plausible use case when an application caches the preconditioner for a fixed matrix and processes independent right-hand sides concurrently. Please make `matvec` safe when the same preconditioner is invoked concurrently from multiple CUDA streams.

warp/_src/optim/linear.py
```diff
@@ -242,17 +242,136 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+        wp.launch(
+            _affine_combine_kernel,
```
shi-eric, line 307: Each block-Jacobi application launches the tiled solve into `scratch`, then launches this kernel to read the result back and write `z`. This adds an extra kernel and a full-vector memory pass both during ordinary execution and CUDA graph replay. Graph capture only avoids the separate host launch overhead.

A focused benchmark of 10,000 warmed applications with 8,000 and 10,000 12x12 blocks took 2.07x and 1.73x as long as launching the tiled solve directly into `z`, respectively.

Could the `alpha`/`beta` combination be folded into `solve_kernel` so that `matvec` uses one application kernel for the full `LinearOperator` contract? This would avoid adding a separate fast path and remove the intermediate buffer entirely.

warp/_src/optim/linear.py
```diff
@@ -242,17 +242,136 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+        (dim * block_size, dim * block_size),
+        A.dtype,
+        device,
+        matvec=block_jacobi_mv,
+    )
+
+
+@functools.cache
+def _create_block_jacobi_kernels(block_size: int):
+    """Build (and cache, per block size) the tile-Cholesky factorize/solve kernel pair
```
shi-eric, line 323: nit: Could you please check the new function docstrings and keep each opening summary to one physical line, in imperative mood, ending with a period? Several currently wrap a single summary sentence across multiple lines, including this one, `_affine_combine_kernel`, and some of the new tests. If more explanation is needed, please end the summary and continue after a blank line. This follows the [Google Python Style Guide's summary-line format](https://google.github.io/styleguide/pyguide.html#s3.8.1-comments-in-doc-strings), while keeping the mood consistent with the other new helper docstrings.

warp/_src/optim/linear.py
```diff
@@ -242,17 +242,136 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+    used by the block-Jacobi preconditioner."""
+
+    @wp.kernel(module="unique")
+    def factorize_kernel(
+        A_flat: wp.array2d(dtype=Any),
```
shi-eric, line 328: nit: Please use subscript syntax for the new Warp array annotations, such as `wp.array2d[Any]` and `wp.array[Any]`. This file still contains legacy factory-style annotations, but new code should not extend that pattern.

changelog/1891.added.md
```diff
@@ -0,0 +1,5 @@
+Block Jacobi preconditioning for the Warp Conjugate Gradient Solver uses the entire diagonal
```
shi-eric, line 1: The entry does not name the API users call, and "the Warp Conjugate Gradient Solver" makes block-Jacobi sound specific to CG. `warp.optim.linear.preconditioner(A, "block_jacobi")` returns a `LinearOperator`, so that description is too narrow. The rest of the entry focuses on the tile implementation but never shows readers how to select the new option.

Could we lead with the public API and use imperative present tense, as required by the [changelog instructions](https://github.com/NVIDIA/warp/blob/main/changelog/README.md#choose-a-category)? Suggested wording:

> Add block-Jacobi preconditioning for `warp.sparse.BsrMatrix` inputs with square blocks using `warp.optim.linear.preconditioner(A, "block_jacobi")`. The preconditioner applies the inverse of each symmetric positive-definite diagonal block instead of treating its scalar diagonal entries independently. This accounts for coupling within each block.

gdaviet (commented):
Thanks @dlevin!  
Currently this uses a tile implementation unconditionally, which is great for largish blocks, but likely unoptimal for small blocks. I wonder if we could have a selection strategy, block size 2-6 could use inverse, 6-12 sequential llt, and >= 12 llt (see e.g. this commit: [fa8b08f1d969eb98c2dec0e9d87824d88837a828](https://github.com/gdaviet/warp/commit/fa8b08f1d969eb98c2dec0e9d87824d88837a828)) for an implementation of he former approaches). The choice could be made explicit (add another argument to `preconditioner`) or be kept heuristical (`"auto"`)

warp/_src/optim/linear.py
dilevin: Thanks Eric, this should be fixed and tested in the most recent commits

warp/_src/optim/linear.py
dilevin: fixed, the code now folds alpha/beta into solve_kernel

dilevin:
@gdaviet  I implemented the other flavors of block Jacobi preconditioning. In tests on my somewhat aged workstation tiled was quite fast for all sizes, but the options are now there for anyone to try out.

dilevin:
@coderabbitai resume

changelog/1891.added.md
shi-eric, line 1: This is way too detailed to work as a scannable changelog entry. Please reduce this to a concise summary of the user-facing capability and the primary way to enable it, using the entries in `CHANGELOG.md` as examples of the intended level of detail.

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,506 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+         - ``"block_jacobi_auto"``: Dispatches to ``"block_jacobi_direct"`` for block sizes 2-6,
+           ``"block_jacobi_sequential"`` for 7-11, or ``"block_jacobi_tile"`` for 12 and up --
+           except when ``A``'s scalar type isn't ``float32``/``float64`` (e.g. ``float16``), in
+           which case block sizes 7 and up use ``"block_jacobi_sequential"`` instead of
+           ``"block_jacobi_tile"``, since the tile strategy can't accept that dtype. These
+           thresholds pick a strategy that supports the input scalar type and stays zero-safe by
+           default; they are not a performance recommendation (see ``"block_jacobi_tile"``
+           above) and may change in the future.
+         - ``"block_jacobi"``: Alias for ``"block_jacobi_auto"``, kept for backward
+           compatibility.
```
shi-eric: This alias unnecessarily makes two spellings part of the public API. Both "block_jacobi" and "block_jacobi_auto" are introduced by this PR, so there is no released behavior requiring backward compatibility between them. Please choose one public spelling and remove the alias.

warp/_src/optim/linear.py
shi-eric, line 237: The docstring has become too long and detailed for generated API documentation and IDE hovers. It mixes the basic information users need to know with benchmark results, compile-time measurements, internal dispatch thresholds and rationale, and repeated descriptions of the same constraints.

Please restructure it around what callers need to use the function correctly: a focused summary, concise argument descriptions, one or two sentences per public strategy covering supported inputs and behavior, followed by `Returns:` and `Raises:` sections.

I don't recommend including benchmark data as they are very machine-specific. If you think the extended strategy tradeoffs is worth documenting, I would suggest putting them in the Sphinx docs on this page https://nvidia.github.io/warp/stable/domain_modules/sparse.html

shi-eric, line 31: Note that kernels decorated with `module="unique"` do not inherit this setting, so you should probably make sure to disable backward code generation on the new kernels.

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,506 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+            out += wp.tile_load(y, shape=(block_size,), offset=(off,)) * beta
+        wp.tile_store(z, out, offset=(off,))
+
+    return factorize_kernel, solve_kernel
+
+
+# --- Direct (QR) and sequential (LDL^T) block-Jacobi helpers -----------------
+#
+# Local re-implementations of QR-based dense inversion and LDL^T factorization for small,
+# fixed-size blocks, adapted from a prototype by a Warp maintainer (gdaviet). Kept as plain
```
shi-eric: Appreciate the attempt to give @gdaviet credit but this is not useful implementation context and does not belong in a source comment. The GitHub pull request discussion and Git history already preserve authorship and provenance. Please keep this comment focused on the implementation constraints that future maintainers need to understand.

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,506 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+           ``float16`` support.
+         - ``"block_jacobi_sequential"``: Block-Jacobi preconditioner that factorizes each
+           diagonal block via a scalar LDL^T factorization. Requires the block to be symmetric
+           positive-definite, but is zero-safe like ``"block_jacobi_direct"``: a non-SPD block
+           falls back to the identity rather than producing NaNs. Supports any Warp floating
+           scalar type, including ``float16``. As with ``"block_jacobi_direct"``, benchmarking
+           showed ``"block_jacobi_tile"`` applies faster at every block size measured, so prefer
+           this strategy for ``float16``/zero-safe-fallback support rather than for speed.
+         - ``"block_jacobi_tile"``: Block-Jacobi preconditioner that factorizes each diagonal
+           block via GPU-tile-parallel Cholesky. Applied fastest of the three strategies at every
```
shi-eric: About "via GPU-tile-parallel Cholesky": This path also compiles and runs on CPU. We should probably use backend-neutral language here unless an actual backend restriction applies.

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,506 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+    "block_jacobi": "auto",
+    "block_jacobi_auto": "auto",
+    "block_jacobi_direct": "direct",
+    "block_jacobi_sequential": "sequential",
+    "block_jacobi_tile": "tile",
+}
+
+# Block-size thresholds used by the "auto" block-Jacobi strategy: blocks of size
+# [2, _BLOCK_JACOBI_AUTO_DIRECT_MAX] use "direct", (..., _BLOCK_JACOBI_AUTO_SEQUENTIAL_MAX] use
+# "sequential", and anything larger uses "tile". Matches the ranges suggested in review, adjusted
```
shi-eric: "Matches the ranges suggested in review" is not durable implementation context, and it does not explain why the thresholds are 6 and 11. Please replace it with the technical rationale for those values. If there is no rationale beyond the review suggestion, remove the sentence and simply document the dispatch mapping.

warp/tests/test_linear_solvers.py
```diff
@@ -940,6 +949,367 @@ def test_functor_compat_errors(test, device):
...
+    z_np = z.numpy()
+    test.assertTrue(np.all(np.isfinite(z_np)), msg="sequential: non-finite output on non-SPD block")
+    assert_np_equal(z_np[:3], x.numpy()[:3], tol=0.0)
+
+
+def test_block_jacobi_preconditioner_concurrent_streams(test, device):
+    """Verify a preconditioner reused across concurrent CUDA streams never corrupts an output.
+
+    Regression test adapted from the exact repro that caught the data race in the tile solve
+    kernel: two priority streams apply the same preconditioner to different inputs with no host
```
shi-eric: Nitpick: "Regression test adapted from the exact repro that caught the data race in the tile solve kernel: " and "Before the fix"  narrate this patch's development history that future maintainers probably don't need to know about.

Please just stick to making clear what the test is checking: reusing a preconditioner across concurrent CUDA streams must not share mutable scratch state or corrupt either output.

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,506 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+    diagonal), mirroring the zero-safe convention of the scalar Jacobi preconditioner, so the
+    result stays well-defined for non-invertible blocks.
+    """
+    Q, R = _qr_decomposition(A)
+    row = type(A[0])()
+    zero = A.dtype(0)
+    one = A.dtype(1)
+    singular = wp.bool(False)
+    for j in range(row.length):
+        singular = singular or (R[j, j] == zero)
```
shi-eric: Codex: This only catches pivots that remain exactly zero after QR. For a rank-deficient block like `[[1, 2], [2, 4]]`, roundoff leaves a tiny nonzero pivot, so the code inverts it instead of taking the documented identity fallback. A CUDA repro with `x = [1, 1]` returns `[nan, nan]` for `float16` and values on the order of 1e7 and 1e15 for `float32` and float64, respectively, instead of `[1, 1]`. Since block_jacobi selects this path for 2x2 blocks, the default strategy is affected too.

Consider using a tolerance relative to the block scale, treat non-finite pivots as singular, and return the identity before the triangular solves.

Add this rank-one case as a regression test across the supported dtypes.

warp/_src/optim/linear.py
dilevin: Thanks Eric, Ok I've shortened it and removed the benchmark comments while retaining the important functional information.

changelog/1891.added.md
dilevin: ok I've significantly shortened the changelog entry while retaining the high level information about the preconditioner change.

warp/_src/optim/linear.py
shi-eric: Thanks!

warp/_src/optim/linear.py
dilevin: Removed the alias, going with block_jacobi_auto so its clear to the user what's going on.

warp/_src/optim/linear.py
```diff
@@ -298,6 +771,30 @@ def _as_scalar_array(x: wp.array):
     return arr
 
 
+def _as_vector_array(x: wp.array, length: int):
+    """View a 1-D scalar array as a 1-D array of fixed-length vectors.
+
+    The underlying storage is shared (no copy); ``x`` must be contiguous along its last
+    dimension, and that dimension's length must be divisible by ``length``.
```
shi-eric, line 814: Can you please enforce the last-dimension contiguity requirement, e.g. checking `x.strides[-1] == type_size_in_bytes(x.dtype)`

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,489 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+        """Apply one block-diagonal inverse per tile and write the result directly into ``z``.
+
+        Computes ``z = alpha * (A_diag^-1 x) + beta * y`` via forward/backward substitution on
+        ``L_flat``, writing straight into the caller-provided ``z`` tile with no shared scratch
+        buffer, so concurrent calls on different streams cannot race on intermediate state.
+        """
+        i = wp.tid()
+        off = i * block_size
+        L_tile = wp.tile_load(L_flat, shape=(block_size, block_size), offset=(off, 0))
+        rhs = wp.tile_load(x, shape=(block_size,), offset=(off,))
```
shi-eric: Please skip loading and solving x when `alpha == 0`. Otherwise NaNs in an unused x propagate through the solve and multiplication by zero, so the result is not `beta * y`.

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,489 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+    row = type(A[0])()
+    zero = A.dtype(0.0)
+    M = type(A)(zero)
+    spd = wp.bool(True)
+    for j in range(row.length):
+        d = A[j, j]
+        for k in range(j):
+            d -= M[j, k] * M[j, k] * M[k, k]
+
+        if d <= zero:
```
shi-eric: Codex: This check can miss a singular block when LDLᵀ roundoff leaves a small positive pivot. For example, the rank-one block `[[9, 21], [21, 49]]` produces a positive residual pivot in `float32` and is solved instead of taking the documented identity fallback.

This also affects the default `block_jacobi` strategy at block size 7. Consider using a scale-relative pivot tolerance, rejecting non-finite pivots, and adding a regression with a rank-deficient block whose factorization requires an inexact division.

shi-eric:
Thanks @dilevin, before bringing this over to GitLab I suggest setting up your Git identity properly so your commits are not authored as `Horde` and signed off with `Signed-off-by: Horde <horde@nvidia.com>`. Setting up your `.gitconfig` + then squashing is probably an easy way to clean up the branch + deal with the authorship issue.

Then please follow the instructions to open a GitLab merge request as I mentioned in this comment https://github.com/NVIDIA/warp/pull/1890#discussion_r3909730815

warp/_src/optim/linear.py
dilevin: updated to say the choice is heuristic

warp/_src/optim/linear.py
dilevin: fixed, I set out to zero initially if alpha = 0

warp/_src/optim/linear.py
dilevin: fixed using scale relative tolerance as well as appropriate tolerances for dtypes.

dilevin:
commits are now signed properly, once automated tests pass I will move this over to gitlab as instructed. Thanks for all the help.

dilevin:
@coderabbit resume

warp/_src/optim/linear.py
```diff
@@ -242,17 +243,522 @@ def preconditioner(A: _Matrix, ptype: str = "diag") -> LinearOperator:
...
+    should use :func:`_make_block_jacobi_preconditioner`, which caps ``block_size`` for this
+    strategy and falls back to ``"tile"`` above the cap.
+    """
+    device = A.device
+    scalar_type = A.scalar_type
+
+    A_diag = sparse.bsr_get_diag(A)
+    dim = A_diag.shape[0]
+    inv_diag = wp.empty_like(A_diag)
+    rel_tol = scalar_type(_get_dtype_epsilon(scalar_type) ** 0.75)
```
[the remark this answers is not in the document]
dilevin: this is because epsilon for float16 ~ 1e-3. Added comments and updated test for float16 to address this.

PR #1896, opened 2026-09-02
Raise on array_scan device failures

### LINKED ISSUE #1894: [BUG] `array_scan` silently corrupts its output on device OOM: unchecked `wp_alloc_device` return launches CUB on a NULL buffer
## Summary

`wp.utils.array_scan` can return normally having written garbage into its output array, with no exception raised and no return value to check.

On a device OOM, `scan_device` passes an unchecked NULL temporary buffer to `cub::DeviceScan`, which fails with `cudaErrorInvalidDevice (101)`. Every layer between that failure and the caller discards the status, so the caller observes a successful call and a corrupt result.

This is silent data corruption rather than a crash, which is what makes it expensive: downstream code that validates its inputs raises a *plausible domain error* on the garbage, so the fault is undiagnosable from the application side. In our case a corrupted prefix-sum reached a neighbor-list bounds check and surfaced as a legitimate-looking overflow error, and we spent a while blaming our own physics before finding the Warp error lines in raw container logs.

## Where the status is dropped

Three layers, all on current `main`:

1. **`warp/native/scan.cu:100`** — the allocation is never tested:
   ```c
   void* temp_buffer = wp_alloc_device(WP_CURRENT_CONTEXT, scan_temp_size, "(native:scan)");
   ```
2. **`warp/native/scan.cu:109-119`** — `check_cuda(cub::DeviceScan::{Inclusive,Exclusive}Sum(temp_buffer, ...))`. `check_cuda` returns `bool` and, per `warp/native/cuda_util.cpp:301`, only calls `wp::set_error_string` and returns `false`; it never throws. `scan_device` discards that `bool`.
3. **`warp/_src/utils.py`** — `array_scan` invokes `runtime.core.wp_array_scan_int_device(...)` as a bare statement, discarding the FFI return, and never consults `runtime.get_error_string()`.

`scan_device` returns `void`, so even a diligent `array_scan` has no channel to check. The failure is unreachable from Python by construction.

One aggravating detail: the per-component loop introduced by `ba982aa7` ("Extend scan and sort", GH-1538) launches CUB on the NULL buffer once per component, so a single OOM now produces `type_length` CUDA errors instead of one.

## Observed

Production logs, H100, long-running worker process:

```
Warp CUDA error 2:   out of memory          (wp_alloc_device_async, warp.cu:817)
Warp CUDA error 101: invalid device ordinal (scan_device, scan.cu:115)
```

No exception was raised. The scan's output array held garbage; the value that reached our downstream bounds check was `18446744073709551615`.

**Second-order effect: the process never recovers.** Warp caches its device and context handles in a process-global runtime singleton with no rebuild path — `init()` in `warp/_src/context.py` is a one-way latch (`if runtime is None`), there is no public `shutdown`/`reinit`, and `wp_cuda_context_destroy` is bound in the ctypes layer but never called. So every subsequent Warp CUDA call in that process fails the same way. We reconstructed one worker that took 44 sequential jobs: job 1 hit the OOM, and jobs 2-43 all died immediately without performing any work. Our current mitigation is to run one job per process and pay a cold start every time, because there is no in-process recovery available.

## Repro sketch

Needs a GPU you can nearly fill; the point is to make CUB's temp allocation fail while the scan inputs themselves fit.

```python
import numpy as np
import warp as wp

wp.init()
device = wp.get_device("cuda:0")

n = 1 << 24
values = wp.array(np.ones(n, dtype=np.int32), device=device)
out = wp.zeros(n, dtype=wp.int32, device=device)

# Leave less headroom than CUB's temp storage requires.
hog = wp.zeros(device.free_memory // 4 - n, dtype=wp.int32, device=device)

wp.utils.array_scan(values, out, inclusive=True)   # returns normally

print(out.numpy()[:8])                             # expected: 1 2 3 4 5 6 7 8
print(wp.context.runtime.get_error_string())       # shows the two CUDA errors above
```

Expected behaviour: `array_scan` raises, as the docs imply ("if running out of memory you should see an exception from Warp that the memory allocation failed"). Actual behaviour: it returns, `out` holds garbage, and the only evidence is a string on stderr.

## Affected versions

Confirmed present in **v1.15.0, v1.16.0, and v1.17.0** (current).

Not a regression — the pre-GH-1538 40-line `scan.cu` (≤ v1.14.0) has the same defect at the old coordinates: unchecked `wp_alloc_device` at `:27`, discarded `check_cuda` on the CUB calls at `:31`/`:33`. GH-1538 relocated and multiplied it rather than introducing it.

## Suggested fix

Guard the allocation and give the failure a path to Python:

```c
void* temp_buffer = wp_alloc_device(WP_CURRENT_CONTEXT, scan_temp_size, "(native:scan)");
if (scan_temp_size > 0 && !temp_buffer)
    return false;   // error string already set by the failed allocation
```

...with `scan_device` returning `bool` and `array_scan` checking it, then raising `RuntimeError(runtime.get_error_string())` — the pattern already used at roughly fifteen sites in `warp/_src/context.py`.

More broadly: any `wp.utils` primitive whose native half returns `void` has this shape and cannot report failure. `radix_sort_pairs` looks like the same case (cf. GH-1373). A sweep for `check_cuda(` whose result is discarded would probably find more.

Happy to open a PR for `scan.cu` if that's welcome.

## Related but distinct

- **GH-960** — same `101 invalid device ordinal` symptom, different root cause (`wp.Mesh` constructor)
- **GH-1430** — `101` from `wp.Mesh(bvh_constructor="cubql")` on devices without mempool support
- **GH-1373** — `radix_sort_pairs` causing CUDA errors under graph capture

### System Information

```
GPU              NVIDIA H100 (data-center Hopper), single device, no MPS
OS               Debian 12 (bookworm) container, x86_64
Python           3.12  (base image ghcr.io/astral-sh/uv:python3.12-bookworm)
torch            CUDA 13.0 build, installed from https://download.pytorch.org/whl/cu130
Warp             >= 1.15.0
Caller           nvalchemiops (pure-Python Warp consumer; calls wp.utils.array_scan
                 from its cell-list build)
Workload         long-lived worker process, many sequential jobs per process,
                 wp.init() once at import
```

rickpr:
## Description

`wp.utils.array_scan()` can return normally having left its output array unwritten on CUDA devices. `scan_device` discards the result of every `check_cuda` call and returns `void`, so a CUDA failure while sizing CUB's temporary storage, allocating it, or running the scan is recorded in Warp's error string and then dropped. The caller sees a
successful call and a wrong result.

This is silent data corruption rather than a crash, which is what makes it hard to diagnose from the caller's side. Downstream code that validates its inputs raises a *plausible domain error* on the bad data, so the failure presents as a bug in the caller rather than a CUDA fault, and the only evidence is a line on stderr that nothing correlates with the result. Note this is not limited to out-of-memory: any CUDA error reached inside `scan_device`, including one already pending on the context from earlier work, is swallowed the same way.

We hit this in a molecular-simulation workload where a corrupted prefix sum inside a neighbor-list build surfaced as a legitimate-looking overflow error.

The asymmetry this fixes: Warp's *Python* allocators already check and raise (`Failed to allocate {size_in_bytes} bytes...` in `warp/_src/context.py`), so `wp.zeros()` fails loudly on OOM. Only the native internal path was silent.

Closes #1894

## Changes

**`warp/native/scan.cu`** — `scan_device` returns `bool` and checks all three failure sites rather than only the allocation:

- The CUB sizing call is the first domino. If it fails, `scan_temp_size` stays at its `0` initializer, so the allocation then asks for zero bytes and *succeeds*, and the scan runs with no temporary storage. Guarding only the allocation does not catch this — it is the path our reproducer actually takes.
- The allocation is guarded only when a nonzero size was requested, and the error string `wp_alloc_device` already recorded is left intact rather than overwritten.
- The per-component loop accumulates status and exits early; `temp_buffer` is still freed on every path.

**Propagation** — `warp/native/scan.h`, `warp/native/warp.h`, and the four `wp_array_scan_*_device` definitions in `warp/native/warp.cu`. The `WP_ENABLE_CUDA=0` stubs in `warp/native/warp.cpp` return `bool`/`false` so CPU-only builds still compile.

**`warp/_src/context.py`** — declares `restype = ctypes.c_bool` for the four device functions. It was previously unset, so ctypes was reading the return as `c_int`.

**`warp/_src/utils.py`** — `array_scan()` raises `RuntimeError(runtime.get_error_string())` on device failure, matching the existing convention used elsewhere in `warp/_src/context.py`.

**Not changed:** `scan_host` still returns `void`. `warp/native/scan.cpp` allocates nothing and has no CUDA status to report, so a `bool` there would always be `true`. Happy to make it symmetric if you would prefer that for the shared Python call site.

**Not in scope:** `radix_sort_pairs` looks like the same shape — native half returns `void`, status dropped. Left out deliberately; glad to follow up.

## Checklist

## Validation summary

Built from source and tested on an NVIDIA H100 80GB HBM3 (CUDA Toolkit 12.9, driver 13.0), via `build_lib.py --quick --no-use-libmathdx`.

1. **Reproduced the bug first, on 1.17.0 from PyPI.** With the device filled so CUB's temporary storage cannot be sized, `array_scan()` returned normally and the output array was left as zeros instead of `[1, 2, 3, ...]`. Warp's own stderr showed `Warp CUDA error 101: invalid device ordinal (in function scan_device, scan.cu:95)` followed by `Warp CUDA error 2: out of memory (... scan.cu:109)` — with no failure at the allocation in between, which is what redirected the fix from a null check to checking every status.

2. **Red/green on the new test.** `test_array_scan_error_device_failure_is_reported` stubs the FFI entry point to report failure and asserts `array_scan()` raises. Reverting only the Python-side check makes it fail with `AssertionError: RuntimeError not raised`; with the change it passes.

3. **No regression.** All 16 `array_scan` tests in `warp/tests/test_utils.py` pass on `cuda:0` and `cpu`. An ordinary scan of 65536 ones still yields `[1..8]` at the head and `65536` at the tail.

4. **The failure now surfaces.** The same filled-device scenario raises `RuntimeError: Warp CUDA error 101: invalid device ordinal (in function scan_device, scan.cu:98)` instead of returning a zeroed buffer, raised by the guard on the sizing call.

5. **CPU-only build.** macOS aarch64 with `WP_ENABLE_CUDA=0` and `-Werror` compiles `warp.cpp` and links `libwarp.dylib` with all four device scan symbols present.

**Coverage limitation.** The new test covers the propagation path deterministically and needs no GPU, but the native guards themselves are only exercised by the hardware reproducer below, which fills an 80 GB device and is not CI material. I could not find an existing way to inject a failure into a *native internal* allocation: `set_device_allocator` is a Python-layer hook, and `wp_alloc_device_default` calls `cudaMalloc` directly with only `g_alloc_tracker` observing. If you would like an injectable allocation-failure path so this class of bug becomes testable in CI, I am glad to do that as a separate PR — it seemed too much of an API decision to fold in here.

## Bug fix

Minimal reproducer **without** this PR applied. Requires a GPU you can fill; the point is to make CUB's temporary storage unsizeable while the scan operands are already resident.

```python
import numpy as np
import warp as wp

wp.init()
device = wp.get_device("cuda:0")

# Plain cudaMalloc fails cleanly when the device is full; the async mempool can
# satisfy a small late request out of reserve.
if wp.is_mempool_supported(device):
    wp.set_mempool_enabled(device, False)

n = 1 << 20
values = wp.array(np.ones(n, dtype=np.int32), device=device)
out = wp.zeros(n, dtype=wp.int32, device=device)

# Fill the device, halving the request whenever one fails.
hogs = []
size_bytes = 1 << 30
while size_bytes >= 4096:
    try:
        hogs.append(wp.zeros(size_bytes // 4, dtype=wp.int32, device=device))
    except RuntimeError:
        size_bytes //= 2

wp.utils.array_scan(values, out, inclusive=True)  # returns normally, no exception

hogs.clear()
print(out.numpy()[:8])  # [0 0 0 0 0 0 0 0], expected [1 2 3 4 5 6 7 8]
```

With this PR, `array_scan()` raises `RuntimeError` carrying the CUDA reason.

## Summary by CodeRabbit

* **Bug Fixes**
  * CUDA array scans now report device-side failures instead of silently returning incomplete output.
  * Clear `RuntimeError` messages are raised when temporary storage sizing, allocation, or scan execution fails.
  * CUDA scan replay now detects and preserves scan failures.

* **Tests**
  * Added coverage to verify that CUDA array-scan errors are surfaced correctly.

### DIFF
--- a/changelog/1894.fixed.md
+++ b/changelog/1894.fixed.md
@@ -1,0 +1,1 @@
+Make `wp.utils.array_scan()` raise a `RuntimeError` with the CUDA error when a device scan fails, instead of returning with unwritten or partially written output.
--- a/warp/native/warp.cu
+++ b/warp/native/warp.cu
@@ -2403,36 +2403,36 @@ static void apic_capture_array_scan_device(
     );
 }
 
-void wp_array_scan_int_device(
+bool wp_array_scan_int_device(
     uint64_t in, uint64_t out, int len, int in_stride, int out_stride, int type_len, bool inclusive
 )
 {
     apic_capture_array_scan_device(in, out, len, in_stride, out_stride, type_len, APIC_TYPE_INT32, inclusive);
-    scan_device((const int*)in, (int*)out, len, in_stride, out_stride, type_len, inclusive);
+    return scan_device((const int*)in, (int*)out, len, in_stride, out_stride, type_len, inclusive);
 }
 
-void wp_array_scan_int64_device(
+bool wp_array_scan_int64_device(
     uint64_t in, uint64_t out, int len, int in_stride, int out_stride, int type_len, bool inclusive
 )
 {
     apic_capture_array_scan_device(in, out, len, in_stride, out_stride, type_len, APIC_TYPE_INT64, inclusive);
-    scan_device((const int64_t*)in, (int64_t*)out, len, in_stride, out_stride, type_len, inclusive);
+    return scan_device((const int64_t*)in, (int64_t*)out, len, in_stride, out_stride, type_len, inclusive);
 }
 
-void wp_array_scan_float_device(
+bool wp_array_scan_float_device(
     uint64_t in, uint64_t out, int len, int in_stride, int out_stride, int type_len, bool inclusive
 )
 {
     apic_capture_array_scan_device(in, out, len, in_stride, out_stride, type_len, APIC_TYPE_FLOAT32, inclusive);
-    scan_device((const float*)in, (float*)out, len, in_stride, out_stride, type_len, inclusive);
+    return scan_device((const float*)in, (float*)out, len, in_stride, out_stride, type_len, inclusive);
 }
 
-void wp_array_scan_double_device(
+bool wp_array_scan_double_device(
     uint64_t in, uint64_t out, int len, int in_stride, int out_stride, int type_len, bool inclusive
 )
 {
     apic_capture_array_scan_device(in, out, len, in_stride, out_stride, type_len, APIC_TYPE_FLOAT64, inclusive);
-    scan_device((const double*)in, (double*)out, len, in_stride, out_stride, type_len, inclusive);
+    return scan_device((const double*)in, (double*)out, len, in_stride, out_stride, type_len, inclusive);
 }
 
 int wp_cuda_driver_version()
--- a/warp/_src/context.py
+++ b/warp/_src/context.py
@@ -6865,6 +6865,7 @@ class Runtime:
                 ctypes.c_int,
                 ctypes.c_bool,
             ]
+            self.core.wp_array_scan_int_device.restype = ctypes.c_bool
             self.core.wp_array_scan_int64_device.argtypes = [
                 ctypes.c_uint64,
                 ctypes.c_uint64,
@@ -6874,6 +6875,7 @@ class Runtime:
                 ctypes.c_int,
                 ctypes.c_bool,
             ]
+            self.core.wp_array_scan_int64_device.restype = ctypes.c_bool
             self.core.wp_array_scan_float_device.argtypes = [
                 ctypes.c_uint64,
                 ctypes.c_uint64,
@@ -6883,6 +6885,7 @@ class Runtime:
                 ctypes.c_int,
                 ctypes.c_bool,
             ]
+            self.core.wp_array_scan_float_device.restype = ctypes.c_bool
             self.core.wp_array_scan_double_device.argtypes = [
                 ctypes.c_uint64,
                 ctypes.c_uint64,
@@ -6892,6 +6895,7 @@ class Runtime:
                 ctypes.c_int,
                 ctypes.c_bool,
             ]
+            self.core.wp_array_scan_double_device.restype = ctypes.c_bool
 
             self.core.wp_radix_sort_pairs_int_host.argtypes = [
                 ctypes.c_uint64,
--- a/warp/_src/utils.py
+++ b/warp/_src/utils.py
@@ -138,7 +138,11 @@ def array_scan(in_array: wp.array, out_array: wp.array, inclusive: bool = True) 
         else:
             raise RuntimeError(f"Unsupported data type: {type_repr(in_array.dtype)}")
 
-    native_func(in_array.ptr, out_array.ptr, in_array.size, in_stride, out_stride, type_length, inclusive)
+    status = native_func(in_array.ptr, out_array.ptr, in_array.size, in_stride, out_stride, type_length, inclusive)
+
+    # Only the CUDA implementations return a status.
+    if in_array.device.is_cuda and not status:
+        raise RuntimeError(runtime.get_error_string())
 
 
 def radix_sort_pairs(
--- a/warp/native/apic.cu
+++ b/warp/native/apic.cu
@@ -544,26 +544,31 @@ static bool apic_replay_ops_into_cuda_capture(
             // g_apic_state is null during rebuild, so these execute (not record)
             // on the capture stream set as current above. Same entry points
             // wp.utils.array_scan() dispatches to.
+            bool scan_ok;
             if (rec->dtype == APIC_TYPE_INT32)
-                wp_array_scan_int_device(
+                scan_ok = wp_array_scan_int_device(
                     reinterpret_cast<uint64_t>(src), reinterpret_cast<uint64_t>(dst), static_cast<int>(rec->length),
                     rec->in_stride, rec->out_stride, rec->type_len, rec->inclusive != 0
                 );
             else if (rec->dtype == APIC_TYPE_INT64)
-                wp_array_scan_int64_device(
+                scan_ok = wp_array_scan_int64_device(
                     reinterpret_cast<uint64_t>(src), reinterpret_cast<uint64_t>(dst), static_cast<int>(rec->length),
                     rec->in_stride, rec->out_stride, rec->type_len, rec->inclusive != 0
                 );
             else if (rec->dtype == APIC_TYPE_FLOAT32)
-                wp_array_scan_float_device(
+                scan_ok = wp_array_scan_float_device(
                     reinterpret_cast<uint64_t>(src), reinterpret_cast<uint64_t>(dst), static_cast<int>(rec->length),
                     rec->in_stride, rec->out_stride, rec->type_len, rec->inclusive != 0
                 );
             else
-                wp_array_scan_double_device(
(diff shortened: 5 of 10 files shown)


warp/native/warp.cu
```diff
@@ -2403,36 +2403,36 @@ static void apic_capture_array_scan_device(
...
 }
 
-void wp_array_scan_int_device(
+bool wp_array_scan_int_device(
     uint64_t in, uint64_t out, int len, int in_stride, int out_stride, int type_len, bool inclusive
 )
 {
     apic_capture_array_scan_device(in, out, len, in_stride, out_stride, type_len, APIC_TYPE_INT32, inclusive);
-    scan_device((const int*)in, (int*)out, len, in_stride, out_stride, type_len, inclusive);
+    return scan_device((const int*)in, (int*)out, len, in_stride, out_stride, type_len, inclusive);
```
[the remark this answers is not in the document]
rickpr: Great catch, fixed. `APIC_OP_SCAN` in apic.cu now sets `success = false` when a replayed scan fails, using the same flag that case already uses for region-resolution failure, so a rebuild can't proceed around an unwritten output.

changelog/1894.fixed.md
```diff
@@ -0,0 +1 @@
+Fix `wp.utils.array_scan()` silently corrupting its output on CUDA devices. `scan_device` discarded every CUDA status and returned no result, so a failure while sizing the temporary storage, allocating it, or running the scan left the output array unwritten while `array_scan()` returned normally. The device scan now propagates failure and `array_scan()` raises a `RuntimeError` carrying the CUDA reason.
```
shi-eric: Thanks @rickpr! Changes look good, thanks for following the contribution guidelines flawlessly.

I just have a minor nit on the changelog entry, can you please focus on the user-visible behavior without mentioning `scan_device`? The `corrupting` term is also perhaps more alarming than necessary since a failure can leave the output unwritten or partially written.

You don't need to use my exact words, but something like this is what I had in mind:

Make `wp.utils.array_scan()` raise a `RuntimeError` with the CUDA error when a device scan fails, instead of returning with unwritten or partially written output.

Please continue keeping the branch up-to-date with `main` and down to a single commit, that makes it easy for me to merge.

rickpr: Thanks! I like your wording so I've taken it verbatim.

Agreed on `corrupting`, I removed it from the commit and changed to `Raise on array_scan device failures` which is more faithful to what this actually does.

shi-eric:
@rickpr Thank you, this is now merged into `main` and will be included in the next Warp release (1.18).

PR #1909, opened 2026-09-04
Reject keyword arguments from incompatible built-in overloads [GH-1841]

### LINKED ISSUE #1841: Python built-ins accept keywords from the wrong overload
### Bug Description

Some overloaded built-ins accept keyword names from another overload.

For example, the matrix version of `transform_point()` takes `mat`, but calling it with `xform` also works. Likewise, the `vec3f` version of `curlnoise()` takes `xyz`, but also accepts `xy`.

```python
import warp as wp

point = wp.vec3f(1.0, 2.0, 3.0)
matrix = wp.mat44f()

print(wp.transform_point(xform=matrix, point=point))

state = wp.rand_init(42)
print(wp.curlnoise(state=state, xy=wp.vec3f(0.5, 0.5, 0.5)))
```

Both calls succeed:

```text
[0.0, 0.0, 0.0]
[0.6576697826385498, 0.048152029514312744, 0.8914632797241211]
```

Both calls should fail because those keyword names are not part of the overload being used. The calls should instead use `mat` and `xyz`:

```python
wp.transform_point(mat=matrix, point=point)
wp.curlnoise(state=state, xyz=wp.vec3f(0.5, 0.5, 0.5))
```

Talha-Dmr:
## Description

In Python-scope built-in overload resolution, passing keyword arguments belonging to another overload was silently accepted rather than raising an error. For example, `wp.transform_point(xform=matrix, point=point)` succeeded even though the `mat44f` overload expects parameter `mat`, not `xform`. Likewise, `wp.curlnoise(state=state, xy=wp.vec3f(0.5, 0.5, 0.5))` succeeded even though the 3D overload expects `xyz`.

This occurred because `get_builtin` used the primary signature as a fast path and then checked all other overloads in `self.overloads` solely by bound argument types, treating the primary signature's parameter names as aliases for all overloads.

Closes #1841.

## Changes

- In `warp/_src/context.py` (`get_builtin`): when keyword arguments are supplied, skip overloads whose `_call_shape` differs from `self._call_shape` during the primary overload loop. Overloads with different parameter names fall through to the call-shape fallback loop, which validates arguments against each overload's signature.
- In `warp/tests/test_builtins_resolution.py`: update overload tests to verify that distinct parameter names are accepted for their corresponding overloads and rejected when mismatched. Added regression test asserting `wp.curlnoise(xy=...)` is rejected for 3D vector arguments.
- Added changelog fragment `changelog/1841.fixed.md`.

## Checklist

## Validation summary

- Verified that without this fix, `wp.transform_point(xform=matrix, point=point)` and `wp.curlnoise(state=state, xy=wp.vec3f(...))` succeed silently.
- Verified that with this fix, both calls properly raise `RuntimeError: Couldn't find a function ... compatible with the arguments ...`.
- Verified that valid calls using proper keyword arguments (`mat=matrix`, `xyz=vec3f`) and positional calls continue to work as expected.
- Ran all 57 tests in `warp/tests/test_builtins_resolution.py` and 23 tests in `warp/tests/test_func.py`: all passed.
- Verified `ruff check` and `ruff format --check` pass on changed files.
- Verified changelog renders properly via `towncrier build --draft`.

## Bug fix

```python
import warp as wp

point = wp.vec3f(1.0, 2.0, 3.0)
matrix = wp.mat44f()

# Succeeded before this PR, now correctly raises RuntimeError:
wp.transform_point(xform=matrix, point=point)

state = wp.rand_init(42)
# Succeeded before this PR, now correctly raises RuntimeError:
wp.curlnoise(state=state, xy=wp.vec3f(0.5, 0.5, 0.5))
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Built-in overload resolution now rejects keyword arguments that don’t match the selected overload’s parameter names.
  * Keyword-based calls no longer interpret incompatible parameter names as aliases.
  * Valid overload-specific default values continue to work as expected.
  * Positional overload resolution remains unchanged.

* **Tests**
  * Expanded coverage for invalid and missing keyword arguments across built-in overloads, including multiplication, transforms, and noise functions.


Talha-Dmr:
Rebased onto latest main and resolved the conflict in warp/tests/test_builtins_resolution.py (kept both the new upstream tile tests and this PR's kwargs regression test). ruff check + format clean; full test run left to CI since I don't have a CUDA build env handy.

PR #1917, opened 2026-09-04
Fix APIC handling of PyTorch tensor views

### LINKED ISSUE #1908: APIC capture_save may snapshot past allocation bounds for offset torch views
## Description

`wp.from_torch()` reports an incorrect capacity for some offset or strided Torch views. APIC treats the view pointer as the region base and snapshots `capacity` bytes from that pointer, which can extend beyond the underlying CUDA allocation.

## Minimal reproduction
Run with the Torch CUDA caching allocator disabled so the invalid copy fails deterministically:
Without it it will seem to work but the capture will potentially capture some corrupted data.
```bash
PYTORCH_NO_CUDA_MEMORY_CACHING=1
```

```python
wp.init()
base = torch.zeros((1, 6), dtype=torch.float32, device="cuda:0")
view = wp.from_torch(base[:, 3:])

with wp.ScopedCapture(device="cuda:0", apic=True) as capture:
    wp.launch(my_kernel, dim=1, inputs=[view], device="cuda:0")

wp.capture_save(capture.graph, "offset_view", inputs={"view": view})
```

`wp.capture_save()` fails with:

```text
Warp CUDA error 1: invalid argument
RuntimeError: APIC: region 1 could not be snapshotted for capture_save
(device-to-host copy failed for a non-transient region)
```

While investigating the snapshot failure, I found that the Warp array reports a capacity of 24 bytes:

```python
print(view.shape, view.strides, view.capacity)
# (1, 3) (24, 4) 24
```
The allocation is 24 bytes and `view.data_ptr()` starts 12 bytes into it. A 24-byte region beginning at the view pointer therefore extends 12 bytes past the allocation.

## Workaround

Materializing a new zero-offset allocation before conversion avoids the invalid region:

```python
view = view.clone(memory_format=torch.contiguous_format)
```

An explicit clone is preferable to only calling `.contiguous()`, because Torch may consider an offset view with a size-one dimension contiguous and return the original view.

### System Information

Observed with Warp 1.16.0 and also present on `main` at `dd571813a`.

Output of `python -c "import warp; warp.print_diagnostics()"`:

```text
Software
  Warp (package):   1.16.0
  Warp (core):      1.16.0
  warp-clang:       1.16.0 (LLVM 18.1.3)
  NumPy:            2.4.6
  PyTorch:          2.9.1
  Python:           3.12.11 (main, Oct  7 2025, 15:34:39) [Clang 20.1.4 ]
  Platform:         Linux-7.0.0-28-generic-x86_64-with-glibc2.39

CUDA
  Enabled:          True
  Toolkit:          12.9
  Driver:           13.2
  NVRTC:            12.9
  Forward compat:   True

Libraries
  MathDx:           0.3.1
  cuBQL:            enabled
  NanoVDB:          32.8.0

Build
  Compiler:         GCC 14.2.1
  Debug:            False
  Verify FP:        False
  Fast math:        False
  Sanitizer:        none

CPU
  Model:            alderlake
  ISA features:     54
                      64bit, adx, aes, avx, avx2, avxvnni, bmi, bmi2, ...

Devices
  cpu
    Name:             x86_64
  cuda:0
    Name:             NVIDIA GeForce RTX 4090
    Memory:           23.5 GiB
    Arch:             sm_89
    SMs:              128
    PCI:              00000000:01:00
    Mempool:          enabled
    GPU->CPU mem:     True
    CPU->GPU mem:     False
    CPU/GPU atomics:  False

### PR DESCRIPTION
**Pull request by shi-eric**
## Description

`wp.from_torch()` previously treated a tensor view's data pointer as the start of its allocation. For offset or strided views, this could cause APIC to snapshot past the underlying PyTorch storage. Separately converted views of the same storage were also treated as unrelated regions, which broke their aliasing after saving and loading a graph.

This change records the PyTorch storage base and size on the Warp array. APIC uses that metadata to calculate view offsets and recognize arrays backed by the same storage. The array's `capacity` now reflects the bytes available from its `ptr` to the end of the storage.

Closes #1908.

## Changes

- Derive a PyTorch-backed array's capacity from the tensor's underlying storage.
- Use the storage base pointer to identify shared external regions in APIC.
- Preserve aliasing between separately converted PyTorch views after an APIC save and load.
- Clarify the distinction between `ptr` and `capacity` and the corresponding full-storage metadata.
- Add CPU and CUDA coverage for view capacity and APIC storage aliasing.

## Checklist

## Validation summary

The first version of `test_from_torch_view_capacity` passed unchanged on `main`, so it did not establish a red test. I extended it to cover the reported offset view as well as interior and strided views. The final test fails for all three cases on `main`, on both CPU and CUDA, and passes on this branch.

The APIC aliasing test also fails on `main` on CPU and CUDA, returning `2.0` instead of the expected `99.0`. It passes on this branch.

I reproduced #1908 with `PYTORCH_NO_CUDA_MEMORY_CACHING=1`. On `main`, `wp.capture_save()` fails with the reported invalid device-to-host copy. On this branch, the offset view reports a capacity of 12 bytes and the graph saves successfully.

A fresh run of the `TestTorch` and `TestApic` selections completed 224 tests across CPU and CUDA without failures. Targeted pre-commit checks also passed.

## Bug fix

Run this example with `PYTORCH_NO_CUDA_MEMORY_CACHING=1` to reproduce the out-of-bounds snapshot on `main`:

```python
import torch
import warp as wp


@wp.kernel
def write_value(view: wp.array2d[float]):
    view[0, 0] = 1.0


wp.init()

base = torch.zeros((1, 6), dtype=torch.float32, device="cuda:0")
view = wp.from_torch(base[:, 3:])

with wp.ScopedCapture(device="cuda:0", apic=True) as capture:
    wp.launch(write_value, dim=1, inputs=[view], device="cuda:0")

wp.capture_save(capture.graph, "offset_view", inputs={"view": view})
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PyTorch interoperability by preserving storage bounds and aliasing for tensor views.
  - APIC capture and replay now correctly maintain shared storage relationships between converted tensor views.
  - Added validation for inconsistent external storage capacities.

- **Documentation**
  - Clarified array capacity, pointer views, external allocation aliasing, and backing-storage behavior.

### DIFF
--- a/changelog/1908.fixed.md
+++ b/changelog/1908.fixed.md
@@ -1,0 +1,1 @@
+Preserve PyTorch storage bounds and aliasing when serializing APIC graphs that use tensor views.
--- a/warp/_src/apic/capture.py
+++ b/warp/_src/apic/capture.py
@@ -38,11 +38,11 @@ class APICapture:
         # C++ APICState*
         self.apic_state: ctypes.c_void_p = runtime.core.wp_apic_create_state()
 
-        # Region tracking: id(base_array) -> (region_id, base_ptr, capacity, base_array).
+        # Region tracking: region key -> (region_id, base_ptr, capacity, base_array).
         # The base array is retained so Python cannot recycle its id() (which would
         # otherwise alias unrelated allocations to the same region_id) and so that
         # capture_save still has a valid base_ptr to read from.
-        self._regions: dict[int, tuple[int, int, int, object]] = {}
+        self._regions: dict[object, tuple[int, int, int, object]] = {}
 
         # Region ids whose backing was allocated *during* this capture (graph-scoped
         # on a memory-pool device). Their memory is gone once the capture ends, so
@@ -81,17 +81,28 @@ class APICapture:
         """Walk the _ref chain to find the base allocation.
 
         Returns:
-            (base_array, byte_offset) where base_array is the root allocation
-            and byte_offset is the offset of arr.ptr from base_array.ptr.
+            (base_array, base_ptr, base_capacity, byte_offset) where base_array
+            is the root array, and byte_offset is the offset of arr.ptr within
+            the backing allocation.
         """
         base = arr
         while hasattr(base, "_ref") and base._ref is not None and hasattr(base._ref, "ptr"):
             base = base._ref
+
+        base_ptr = base._storage_base_ptr if base._storage_base_ptr is not None else base.ptr
+        base_capacity = base._storage_nbytes if base._storage_nbytes is not None else base.capacity
+
         # Empty arrays (shape=(0,)) have arr.ptr == None; treat as zero offset.
-        if arr.ptr is None or base.ptr is None:
-            return base, 0
-        byte_offset = arr.ptr - base.ptr
-        return base, byte_offset
+        if arr.ptr is None or base_ptr is None:
+            return base, base_ptr, base_capacity, 0
+        byte_offset = arr.ptr - base_ptr
+        return base, base_ptr, base_capacity, byte_offset
+
+    @staticmethod
+    def _region_key(base, base_ptr):
+        if base._storage_base_ptr is not None:
+            return "external", base_ptr
+        return id(base)
 
     def _find_handle_offsets(self, dtype, base_offset: int = 0) -> list[int]:
         """Recursively find byte offsets of ``wp.handle`` fields in a type.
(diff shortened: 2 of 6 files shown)


frlai:
This branch seems to solve the previous problems i was facing. Thanks for getting on this so fast.

Issue #186, opened 2024-02-19
warp 0.13.0 License clarification


traversaro:
Hello warp authors, thanks a lot for working on warp and changing license to allow commercial usage (see https://github.com/NVIDIA/warp/releases/tag/v0.13.0). 

Together with @flferretti we are mantaing the conda-forge package of warp, and we wanted to understand how to properly document this new license, in particular using [`SPDX`](https://en.wikipedia.org/wiki/Software_Package_Data_Exchange)

I see that the new license is reference as "NVIDIA Software License" (see for example https://github.com/NVIDIA/warp/blob/9b2a57f7268e8523c24eec1625c6d08548e5c60c/pyproject.toml#L12). However, I am not finding any reference to this "NVIDIA Software License". Is that a "generic"  license used by NVIDIA, or it is something warp-specific? If it is warp specific, could it make sense to call it with a warp-specific name, to avoid ambiguity wit the software license of other NVIDIA libraries?

Thanks in advance for the answer!

shi-eric:
Hi @traversaro, looking at our internal guidance, I think `SPDX-License-Identifier: LicenseRef-NvidiaProprietary` is the appropriate identifier to use here.

traversaro:
Thanks a lot @shi-eric ! I just have a comment on this.

> Hi @traversaro, looking at our internal guidance, I think SPDX-License-Identifier: LicenseRef-NvidiaProprietary is the appropriate identifier to use here.

Based on https://spdx.github.io/spdx-spec/v2.3/other-licensing-information-detected/, ""LicenseRef-"[idstring] where [idstring] is a unique string containing letters, numbers, . and/or -." it seems that the license used by warp is quite warp-specific, and Nvidia employies for its source available software many prorietary licenses, that are all different. Using `LicenseRef-NvidiaProprietary` seems a bit risky to me, as it seems to be highly probably that people will use it to refer to other (but different) licenses used for proprietary Nvidia software. What do you think on this? Could it make sense to use a more warp-specific `idstring`?

shi-eric:
Hi @traversaro, I just double-checked with Legal after your last question. I'm told that `LicenseRef-NvidiaProprietary` is used as a "catch-all" and is indeed the correct one to use for Warp.

traversaro:
[...]

PR #1919, opened 2026-09-04
Floor integer division toward negative infinity [GH-1918]

### LINKED ISSUE #1918: Integer floor division (`//`) truncates toward zero instead of flooring
### Bug Description

Integer `//` in a Warp kernel rounds toward zero rather than toward negative infinity, so it returns the wrong value whenever exactly one operand is negative. There is no error or warning.

`ast.FloorDiv` maps to the `floordiv` built-in (`warp/_src/codegen.py`), which is documented as "Divide two scalars using floor division" (`warp/_src/builtins.py`). The floating-point overloads implement that as `floorf(a / b)` (`warp/native/builtin.h`), but the integer overload declared in `DECLARE_INT_OPS` is plain C division, which truncates.

### Reproducer

```python
import warp as wp
import numpy as np

@wp.kernel
def floor_div(a: wp.array(dtype=wp.int32), b: wp.array(dtype=wp.int32), out: wp.array(dtype=wp.int32)):
    i = wp.tid()
    out[i] = a[i] // b[i]

a = wp.array([-7, 7, -1], dtype=wp.int32)
b = wp.array([2, -2, 2], dtype=wp.int32)
out = wp.zeros(3, dtype=wp.int32)
wp.launch(floor_div, dim=3, inputs=[a, b], outputs=[out])

print(out.numpy())                            # [-3 -3  0]
print(np.floor_divide(a.numpy(), b.numpy()))  # [-4 -4 -1]  <- expected
```

| expression | Warp integer | Python / NumPy / Warp float |
| ---------- | ------------ | --------------------------- |
| `-7 // 2`  | `-3`         | `-4`                        |
| `7 // -2`  | `-3`         | `-4`                        |
| `-1 // 2`  | `0`          | `-1`                        |

The same kernel with `wp.float32` returns the correct `-4.0`, so integer and floating-point `//` disagree within Warp itself.

### Impact

Any kernel computing something like `index // stride`, a centered-stencil offset, or a grid bin where the dividend can go negative silently gets a value that is one too high.

Existing coverage misses this because `randvals()` in `warp/tests/test_arithmetic.py` only generates positive integers (`rng.integers(1, high=5)`), so the existing `wp.floordiv` assertions never see a negative operand.

Note that `mod` appears to be intentionally different -- it is documented as "Modulo operation using truncated division" -- which suggests the truncation in `floordiv` is an oversight rather than a deliberate choice.

I have a patch and a regression test ready (verified failing before the change and passing after, across `int8`/`int16`/`int32`/`int64`) and will open a PR against this issue.

### System Information

Warp 1.18.0.dev2, built from source at commit d4de134 with `python build_lib.py --no-cuda`.
Python 3.10.12, Ubuntu 22.04 (x86_64), GCC 11.4.0, NumPy 2.2.6.
Reproduced on the `cpu` device. The affected code path is shared with CUDA,
since `DECLARE_INT_OPS` is `CUDA_CALLABLE`.

elbourne12345:
## Description

Closes #1918.

Integer floor division (`//`) truncates toward zero instead of rounding toward negative infinity, so kernels
silently compute the wrong value whenever exactly one operand is negative.

`ast.FloorDiv` maps to the `floordiv` builtin (`warp/_src/codegen.py`), which is documented as "Divide two scalars
using floor division" (`warp/_src/builtins.py`). The floating-point overloads honor that contract with
`floorf(a / b)` (`warp/native/builtin.h`), but the integer overload declared in `DECLARE_INT_OPS` is plain C
division, which truncates. There is no error or warning, so an expression such as `index // stride` with a
negative index quietly produces a wrong result.

Note that `mod` is intentionally different — it is documented as "Modulo operation using truncated division" — so
it is left unchanged.

This is a behavior change for code that depended on the previous truncating result, but that result contradicted
the documented semantics, Python's `//`, and Warp's own floating-point overloads.

## Changes

- `warp/native/builtin.h`: integer `floordiv()` moves out of `DECLARE_INT_OPS`. Signed types call a
  `floordiv_signed()` helper that lowers the truncated quotient when the remainder is non-zero and the operands have
  opposite signs; the unsigned overloads remain plain `a / b`, so no signed-only comparison is compiled for unsigned
  types.
- `warp/tests/test_arithmetic.py`: added `test_floordiv_negative`, registered for every signed integer type, which
  exercises the `//` operator over mixed-sign operands, including exact quotients with a negative divisor and a zero
  dividend, against `np.floor_divide`.
- Added changelog fragment `changelog/1918.fixed.md`.

## Checklist

## Validation summary

Built with CUDA 12.9 (`python build_lib.py --cuda-path /usr/local/cuda`, Ubuntu 24.04, GCC 13.3) and CPU-only
(`--no-cuda`), and ran the arithmetic suite on `cpu` and `cuda:0` (NVIDIA L4).

- Verified `test_floordiv_negative` fails on `main` without this change, for `int8`, `int16`, `int32` and `int64`,
  on both devices. The int16 case reports:

  ```
  Mismatched elements: 3 / 12 (25%)
  ACTUAL:  array([-3, -3,  3,  3,  0, -2,  1,  1,  0, -2, -9, -9], dtype=int16)
  DESIRED: array([-4, -4,  3,  3, -1, -2,  1,  1,  0, -2, -9, -9], dtype=int16)
  ```

- Verified the new test passes with the fix applied, and that the rest of `test_arithmetic` (134 tests) still passes,
  including the existing `floordiv` coverage in `test_binary_ops` and the unsigned-integer cases.
- Warning check for the restructured dispatch:
  - nvcc build log: 0 occurrences of `warning #186-D` (the previous revision emitted 1352 on the same machine); the
    only remaining nvcc output is the pre-existing deprecated-architecture notice.
  - Runtime kernel compilation with `warp.config.log_level = warp.LOG_DEBUG`: the NVRTC program log is empty (the
    previous revision printed eight `#186-D` lines per CUDA module, with and without precompiled headers).
  - `g++ -std=c++17 -Wtype-limits -Werror -fsyntax-only` on a translation unit including `builtin.h`: clean (the
    previous revision failed with `comparison of unsigned expression in '< 0' is always false`).
- Confirmed the same kernel using `float32` returned the correct `-4.0` both before and after, which is what
  identified the integer/float inconsistency.
- `pre-commit run` (ruff, clang-format, typos, generated-files check) and `towncrier build --draft` pass locally.

## Bug fix

```python
import warp as wp
import numpy as np

@wp.kernel
def floor_div(a: wp.array(dtype=wp.int32), b: wp.array(dtype=wp.int32), out: wp.array(dtype=wp.int32)):
    i = wp.tid()
    out[i] = a[i] // b[i]

a = wp.array([-7, 7, -1], dtype=wp.int32)
b = wp.array([2, -2, 2], dtype=wp.int32)
out = wp.zeros(3, dtype=wp.int32)
wp.launch(floor_div, dim=3, inputs=[a, b], outputs=[out])

print(out.numpy())                                    # [-3 -3  0]  (truncated)
print(np.floor_divide(a.numpy(), b.numpy()))          # [-4 -4 -1]  (expected)
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected signed-integer floor division to round toward negative infinity, matching Python and NumPy behavior.
  - Ensured `//` and `wp.floordiv()` handle negative operands consistently across supported integer types.

- **Tests**
  - Added coverage for negative signed-integer floor division across supported signed integer types.

- **Documentation**
  - Documented the corrected floor division behavior.

### DIFF
--- a/changelog/1918.fixed.md
+++ b/changelog/1918.fixed.md
@@ -1,0 +1,2 @@
+Fix floor division for signed integers so `//` and `wp.floordiv()` round toward negative infinity,
+matching Python and NumPy.
--- a/warp/native/builtin.h
+++ b/warp/native/builtin.h
@@ -587,7 +587,6 @@ inline CUDA_CALLABLE T mod(T a, T b) { return a%b; } \
 inline CUDA_CALLABLE T min(T a, T b) { return a<b?a:b; } \
 inline CUDA_CALLABLE T max(T a, T b) { return a>b?a:b; } \
 inline CUDA_CALLABLE T clamp(T x, T a, T b) { return min(max(a, x), b); } \
-inline CUDA_CALLABLE T floordiv(T a, T b) { return a/b; } \
 inline CUDA_CALLABLE T nonzero(T x) { return x == T(0) ? T(0) : T(1); } \
 inline CUDA_CALLABLE T bit_and(T a, T b) { return a&b; } \
 inline CUDA_CALLABLE T bit_or(T a, T b) { return a|b; } \
@@ -633,6 +632,25 @@ DECLARE_INT_OPS(uint8)
 DECLARE_INT_OPS(uint16)
 DECLARE_INT_OPS(uint32)
 DECLARE_INT_OPS(uint64)
+
+/* C++ integer division truncates toward zero; adjust signed, non-exact
+   results with opposite signs to round toward negative infinity. */
+template <typename T> inline CUDA_CALLABLE T floordiv_signed(T a, T b)
+{
+    T q = a / b;
+    T r = a % b;
+    if (r != T(0) && ((r < T(0)) != (b < T(0))))
+        q -= T(1);
+    return q;
+}
+inline CUDA_CALLABLE int8 floordiv(int8 a, int8 b) { return floordiv_signed(a, b); }
+inline CUDA_CALLABLE int16 floordiv(int16 a, int16 b) { return floordiv_signed(a, b); }
+inline CUDA_CALLABLE int32 floordiv(int32 a, int32 b) { return floordiv_signed(a, b); }
+inline CUDA_CALLABLE int64 floordiv(int64 a, int64 b) { return floordiv_signed(a, b); }
+inline CUDA_CALLABLE uint8 floordiv(uint8 a, uint8 b) { return a / b; }
+inline CUDA_CALLABLE uint16 floordiv(uint16 a, uint16 b) { return a / b; }
+inline CUDA_CALLABLE uint32 floordiv(uint32 a, uint32 b) { return a / b; }
+inline CUDA_CALLABLE uint64 floordiv(uint64 a, uint64 b) { return a / b; }
 
 
 inline CUDA_CALLABLE int8 step(int8 x) { return x < 0 ? 1 : 0; }
--- a/warp/tests/test_arithmetic.py
+++ b/warp/tests/test_arithmetic.py
@@ -1094,6 +1094,37 @@ for dtype in np_float_types:
         TestArithmetic, f"test_float_to_int_{dtype.__name__}", test_float_to_int, devices=devices, dtype=dtype
     )
 
+
+def test_floordiv_negative(test, device, dtype, register_kernels=False):
+    # Warp's `//` maps to the floordiv builtin, which is documented as floor
+    # division and must round toward negative infinity for signed operands,
+    # matching Python/NumPy and the floating-point overloads.
+    wptype = wp.dtype_from_numpy(np.dtype(dtype))
+
+    def check_floordiv(
+        a: wp.array[wptype],
+        b: wp.array[wptype],
+        outputs: wp.array[wptype],
+    ):
+        for i in range(12):
+            outputs[i] = a[i] // b[i]
+
+    kernel = getkernel(check_floordiv, suffix=dtype.__name__, enable_backward=False)
+
+    if register_kernels:
+        return
+
+    a_np = np.array([-7, 7, -7, 7, -1, -8, 5, -5, 0, 6, 9, -9], dtype=dtype)
+    b_np = np.array([2, -2, -2, 2, 2, 4, 3, -3, -3, -3, -1, 1], dtype=dtype)
+
+    a = wp.array(a_np, dtype=wptype, device=device)
+    b = wp.array(b_np, dtype=wptype, device=device)
+    outputs = wp.zeros(len(a_np), dtype=wptype, device=device)
+
+    wp.launch(kernel, dim=1, inputs=[a, b], outputs=[outputs], device=device)
+    assert_np_equal(outputs.numpy(), np.floor_divide(a_np, b_np))
+
+
 for dtype in np_scalar_types:
     add_function_test_register_kernel(
         TestArithmetic, f"test_clamp_{dtype.__name__}", test_clamp, devices=devices, dtype=dtype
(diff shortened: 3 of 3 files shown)


shi-eric:
The CUDA-enabled builds are flooding their logs with warning #186-D because this non-template `if constexpr` leaves signed-only comparisons visible in the unsigned overloads. The [current CI run](https://github.com/NVIDIA/warp/actions/runs/33930121250) emitted 8,736 copies across all ten CUDA build jobs, covering CUDA 12 and 13 on Linux and Windows, x86-64 and ARM, and both `build_lib.py` and CMake builds. CPU-only builds emit none. Each diagnostic also prints the expanded macro line, producing artifact-build logs between 5 and 8 MB, but GitHub creates no annotations for the warnings, so every affected job remains green. The unsigned runtime results are unchanged, but the same implementation also fails under GCC with `-Wall -Wextra -Werror`. Please restructure the signed/unsigned dispatch so these comparisons are not compiled for unsigned overloads rather than suppressing warning 186.

shi-eric:
`changelog/README.md` says to use the GitHub issue number when an issue exists; the `+` form is reserved for orphan fragments. Since this change fixes #1918, please rename the file to `changelog/1918.fixed.md` so Towncrier adds the issue link.

The current entry reads more like an issue summary than a release-note bullet. The examples and implementation details are already covered in #1918, so please condense it to the user-visible change. For example:

```markdown
Fix floor division for signed integers so `//` and `wp.floordiv()` round toward negative infinity,
matching Python and NumPy.
```

warp/native/builtin.h
```diff
@@ -587,7 +587,20 @@ inline CUDA_CALLABLE T mod(T a, T b) { return a%b; } \
 inline CUDA_CALLABLE T min(T a, T b) { return a<b?a:b; } \
 inline CUDA_CALLABLE T max(T a, T b) { return a>b?a:b; } \
 inline CUDA_CALLABLE T clamp(T x, T a, T b) { return min(max(a, x), b); } \
-inline CUDA_CALLABLE T floordiv(T a, T b) { return a/b; } \
+inline CUDA_CALLABLE T floordiv(T a, T b) \
+{ \
+    /* Integer division truncates toward zero, so for signed types bias the \
+       quotient down when exactly one operand is negative. This matches Python's \
+       floor division and the floating-point overloads below. */ \
+    T q = a/b; \
```
shi-eric: The current wording is slightly imprecise: opposite signs require adjusting the quotient only when the division has a nonzero remainder. Could you shorten it to focus on that rule?

```suggestion
    /* C++ integer division truncates toward zero; adjust signed, non-exact \
       results with opposite signs to round toward negative infinity. */ \
```

The comparison with Python and the floating-point overloads is already covered by the test and changelog.

elbourne12345:
Thanks @shi-eric for reviewing, I've addressed your three points in 01767816

- **Dispatch restructure:** `floordiv` is now out of `DECLARE_INT_OPS`. Signed types go through a small `floordiv_signed<T>` helper; the unsigned overloads are plain `a / b`, so no signed-only comparison is compiled for unsigned types. Rebuilt with CUDA 12.9: 0 occurrences of `#186-D` in the nvcc log (1352 before on the same machine), and `g++ -std=c++17 -Wtype-limits -Werror` on a TU including `builtin.h` is clean. The runtime NVRTC log under `log_level = LOG_DEBUG` is also empty now.
- **Changelog:** renamed to `changelog/1918.fixed.md` and condensed to your suggested wording; `towncrier build --draft` renders it with the issue link.
- **Comment:** applied your wording. It now sits above the helper rather than inside the macro, so the line continuations were dropped.

I also added four vectors to the test (`0 // -3`, `6 // -3`, `9 // -1`, `-9 // 1`) so the "non-exact" rule is actually exercised with a negative divisor. The full `test_arithmetic` suite passes on CPU and an L4. Details are in the updated Validation summary.

warp/native/builtin.h
elbourne12345: Applied in 01767816, moved above the new `floordiv_signed` helper (outside the macro, so without the trailing backslashes).

changelog/1918.fixed.md
```diff
@@ -0,0 +1,2 @@
+Fix floor division for signed integers so `//` and `wp.floordiv()` round toward negative infinity,
+matching Python and NumPy.
```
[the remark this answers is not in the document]
elbourne12345: Keeping this as is. `changelog/README.md` asks for imperative present tense and no leading `-`

shi-eric:
Thanks, @elbourne12345. Can you please rebase your branch with the latest `main`? Then we can get it merged.

elbourne12345:
@shi-eric Rebased onto current main (7ec2fafa) — single commit, now 7784b3e3, no conflicts. Thanks.

PR #1939, opened 2026-09-09
docs: fix docstring argument names that don't match signatures

IshaanPotle:
## Description

Four public functions have `Args:` entries naming parameters they do not accept. Because these docstrings are rendered into the API reference, the published documentation advertises keyword arguments that raise `TypeError` if a user tries them, and in one case the real first parameter is left undocumented.

No issue is linked; I found these by comparing each documented `Args:` name against the actual signature.

## Changes

**`warp.fem`**

- `ExplicitGeometryPartition.rebuild()` documented `geometry`, `max_cell_count` and `max_side_count`. Its signature is `(cell_mask, temporary_store=None)`. The three names match `__init__`, so they appear to have been copied when `rebuild()` was split out. Removed them and documented `temporary_store`, which was missing.
- `make_trial()` documented `device`. Its signature is `(space, space_restriction, space_partition, domain)`. Removed it. (`make_test()` immediately above has the same parameter list and correctly does not document `device`.)
- `make_collocated_function_space()` documented `geo: the Geometry on which to build the space`. The first parameter is `basis_space`, which was otherwise undocumented. Replaced the entry rather than deleting it.

**`warp.render`**

- `UsdRenderer.render_line_list()` documented `time`. Its signature is `(name, vertices, indices, color=None, radius=0.01, visible=True)`. Removed it.

**Deliberately not changed**

`fem.operator.lookup()` and `fem.operator.partition_lookup()` also document names absent from their Python signatures (`max_dist`, `guess`, `env_index`, `filter_array`, `filter_target`). These are operator stubs whose `(domain, x)` signature is a placeholder and whose docstrings intentionally describe the variadic call forms resolved at kernel-compile time. Changing them would make the docs worse, so I left them alone. Same for a few docstrings where wrapped prose such as ``internally managed :class:`Event` `` merely looks like an `Args:` entry.

## Checklist

- [x] New or existing tests cover these changes. *(Docstring text only; no test exercises these strings.)*
- [x] The documentation is up to date with these changes. *(This PR is the documentation fix.)*

## Validation summary

Each of the six removed names was checked against the live signature rather than read off the diff:

- Imported this branch and called `inspect.signature()` on all four functions. Confirmed that none of `geometry`, `max_cell_count`, `max_side_count`, `device`, `geo` or `time` appears in the signature of the function whose docstring named it, and that `temporary_store` and `basis_space` are real parameters. All 12 assertions passed.
- Re-ran the signature-vs-docstring comparison across all 463 Python modules under `warp/`. Before: 16 mismatches. After: 10, all of which are the operator stubs and wrapped-prose cases described above, i.e. every genuine mismatch is resolved and nothing new was introduced.
- `ruff check` and `ruff format --check` pass on the four modified files under the repository's own configuration.

This change touches only text inside docstrings, so there is no runtime behavior to test; the risk is limited to the rendered documentation, which is what the change is correcting.

## Bug fix

Documentation-only, so there is no runtime reproducer for a crash. The mismatch is directly observable by trying the argument the current docs advertise:

```python
import warp.fem as fem

fem.make_trial(space=None, device="cpu")
# TypeError: make_trial() got an unexpected keyword argument 'device'
```

---

*Disclosure: I used an AI assistant to help scan for signature/docstring mismatches. I reviewed every hit myself, verified each one against the live signatures, and rejected the 10 cases where the tool was wrong — those are listed above.*

## Summary by CodeRabbit

- **Documentation**
  - Corrected documented argument names for geometry partitioning, finite element, and USD rendering APIs.
  - Updated `rebuild()` documentation to describe `temporary_store` as storage for intermediate arrays and remove obsolete constructor-only parameters.
  - Removed outdated `device`, `geo`, and `time` argument references from API documentation.
  - Clarified the documented `basis_space` parameter for collocated function spaces.
  - No runtime behavior or public API declarations changed.

### DIFF
--- a/warp/_src/fem/geometry/partition.py
+++ b/warp/_src/fem/geometry/partition.py
@@ -257,7 +257,7 @@ class CellBasedGeometryPartition(GeometryPartition):
             cell_inclusion_test_func: Device function deciding whether a cell is in the partition.
             device: Warp device to run the computation on.
             max_side_count: Optional cap on the number of sides to allocate.
-            temporary_store: Optional temporary storage for intermediate arrays.
+            temporary_store: Temporary storage for intermediate arrays.
         """
         self.side_arg_value.invalidate(self)
 
@@ -387,7 +387,7 @@ class LinearGeometryPartition(CellBasedGeometryPartition):
             partition_rank: The index of the partition being created (0 to partition_count-1).
             partition_count: The total number of partitions over the geometry.
             device: Warp device on which to perform and store computations.
-            temporary_store: Optional temporary storage for intermediate arrays.
+            temporary_store: Temporary storage for intermediate arrays.
         """
         super().__init__(geometry)
 
@@ -465,7 +465,7 @@ class ExplicitGeometryPartition(CellBasedGeometryPartition):
                 synchronization.
             max_side_count: If positive, limits the number of sides to avoid device/host
                 synchronization.
-            temporary_store: Optional temporary storage for intermediate arrays.
+            temporary_store: Temporary storage for intermediate arrays.
         """
 
         super().__init__(geometry)
@@ -486,10 +486,8 @@ class ExplicitGeometryPartition(CellBasedGeometryPartition):
         """Rebuild the geometry partition from a new active cell mask.
 
         Args:
-            geometry: the geometry to partition
             cell_mask: warp array of length ``geometry.cell_count()`` indicating which cells are selected. Array values must be either ``1`` (selected) or ``0`` (not selected).
-            max_cell_count: if positive, will be used to limit the number of cells to avoid device/host synchronization
-            max_side_count: if positive, will be used to limit the number of sides to avoid device/host synchronization
+            temporary_store: Temporary storage for intermediate arrays.
         """
         self.cell_arg_value.invalidate(self)
 
--- a/changelog/+fix-docstring-arg-names.documentation.md
+++ b/changelog/+fix-docstring-arg-names.documentation.md
@@ -1,0 +1,3 @@
+Correct the documented argument names for `ExplicitGeometryPartition.rebuild()`, `fem.make_trial()`,
+`fem.make_collocated_function_space()`, and `UsdRenderer.render_line_list()`, which listed parameters that those
+functions do not accept.
--- a/warp/_src/fem/field/__init__.py
+++ b/warp/_src/fem/field/__init__.py
@@ -84,7 +84,6 @@ def make_trial(
         space_restriction: restriction of the space topology to a domain
         space_partition: if `space_restriction` is ``None``, the optional subset of node indices to consider
         domain: if `space_restriction` is ``None``, optional subset of elements to consider
-        device: Warp device on which to perform and store computations
 
     Returns:
         the trial field
--- a/warp/_src/fem/space/__init__.py
+++ b/warp/_src/fem/space/__init__.py
@@ -162,7 +162,7 @@ def make_collocated_function_space(
     Constructs a function space from a scalar-valued basis space and a value type, such that all degrees of freedom of the value type are stored at each of the basis nodes.
 
     Args:
-        geo: the Geometry on which to build the space
+        basis_space: the scalar-valued basis space on which to build the function space
         dtype: value type the function space. If ``dof_mapper`` is provided, the value type from the DofMapper will be used instead.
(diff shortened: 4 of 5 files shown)


warp/_src/fem/geometry/partition.py
```diff
@@ -486,10 +486,8 @@ def rebuild(
         """Rebuild the geometry partition from a new active cell mask.
 
         Args:
-            geometry: the geometry to partition
             cell_mask: warp array of length ``geometry.cell_count()`` indicating which cells are selected. Array values must be either ``1`` (selected) or ``0`` (not selected).
-            max_cell_count: if positive, will be used to limit the number of cells to avoid device/host synchronization
-            max_side_count: if positive, will be used to limit the number of sides to avoid device/host synchronization
+            temporary_store: Optional temporary storage for intermediate arrays.
```
christophercrouzet: This parameter shows up as `temporary_store=None` in the generated doc's signature (see https://nvidia.github.io/warp/latest/api_reference/_generated/warp.fem.ExplicitGeometryPartition.html#warp.fem.ExplicitGeometryPartition.rebuild), so “Optional” is redundant and unnecessary.

christophercrouzet (changes requested):
Hi @IshaanPotle, thanks for this.

warp/_src/fem/geometry/partition.py
IshaanPotle: Good catch, thanks — fixed in aaa1d16. It now reads `Temporary storage for intermediate arrays.`

For what it's worth, I took that wording from `__init__` just above (line 468), which has the same redundant `Optional`. Happy to fix that one too if you'd like it in this PR, or leave it as out of scope.

christophercrouzet: Thanks. Yes, please feel free fold that change in as well.

IshaanPotle: Done in 6bfca79. I folded it into all three remaining `temporary_store` entries in this module (lines 260, 390 and 468) rather than just `__init__`, since they share the same wording and the same `= None` default — leaving two of them inconsistent seemed worse. Happy to narrow it back to just `__init__` if you'd prefer to keep the diff tighter.

I left `max_side_count: Optional cap on the number of sides to allocate.` alone, since that one defaults to `-1` rather than `None` and "Optional" there describes the behaviour rather than the signature.

warp/_src/fem/geometry/partition.py
```diff
@@ -486,10 +486,8 @@ def rebuild(
         """Rebuild the geometry partition from a new active cell mask.
 
         Args:
-            geometry: the geometry to partition
             cell_mask: warp array of length ``geometry.cell_count()`` indicating which cells are selected. Array values must be either ``1`` (selected) or ``0`` (not selected).
-            max_cell_count: if positive, will be used to limit the number of cells to avoid device/host synchronization
-            max_side_count: if positive, will be used to limit the number of sides to avoid device/host synchronization
+            temporary_store: Temporary storage for intermediate arrays.
```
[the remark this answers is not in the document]
IshaanPotle: Not taking this one - a maintainer requested the opposite in this thread. Sphinx already renders the default into the generated signature as `temporary_store=None`, so "Optional" in the description restates what the reader can already see.

warp/_src/fem/geometry/partition.py
christophercrouzet: Thanks, one more thing, can you please squash all the commits into one once you've rebased on `main`?

IshaanPotle: Done - rebased on `main` (67e40c5) and squashed down to a single commit, d35483e. The diff is unchanged: 5 files, +8/-9.

christophercrouzet:
Thanks @IshaanPotle, it's been merged as commit [d559d70](https://github.com/NVIDIA/warp/commit/d559d703dc94de0bc2fdc05da3c92ce2b1d7c229).

Of course I did not follow the process correctly and this ended up assigning the commit to my name instead of yours (the commit message still says it's from you). Apologies...

christophercrouzet:
Edit: I'll revert and re-merge it properly.

christophercrouzet:
Edit: Done in [0e76b6a](https://github.com/NVIDIA/warp/commit/0e76b6a0e21ce0b06e1dab3b98efd8b5afa9f4f9) :heavy_check_mark:

PR #1958, opened 2026-09-14
Skip atomic CAS spinlock tests on pre-Volta devices

zhihuidu-amd:
### What this changes

Exclude CUDA devices below `sm_70` from `warp/tests/test_atomic_cas.py`:

```python
devices = [d for d in get_test_devices() if not d.is_cuda or d.arch >= 70]
```

Filtering the device list once follows the existing idiom in `test_atomic.py` and avoids
guarding each of the 24 registrations. CPU devices and `sm_70+` CUDA devices are unaffected.

### The pre-Volta CUDA risk (documented, not measured here)

Each GPU test in this module runs a spinlock: threads contend for a single lock, and the
thread that wins the CAS must leave the loop and reach the release while its siblings are
still spinning. That relies on threads of a warp making independent forward progress, which
CUDA provides from Volta (`sm_70`) onward — see
[Independent Thread Scheduling](https://docs.nvidia.com/cuda/cuda-programming-guide/03-advanced/advanced-kernel-programming.html#independent-thread-scheduling).
Under the older reconvergence model this pattern **can** deadlock depending on warp
scheduling. Warp still emits `sm_52`, `sm_60` and `sm_61` targets in `build_dll.py`.

**I have not reproduced this on pre-Volta hardware — I do not have any.** The guard rests on
the documented architectural requirement, not on a measurement.

### The ROCm failure (measured) — and why this PR does not fix it

I found this while running the suite on a downstream HIP/ROCm port of Warp, where
`test_cas_2d_float32` hangs on gfx942 (MI300X). The host sits idle waiting on a GPU kernel
that has stopped retiring; the suite never reports, it just stops. A CDNA wavefront executes
in lockstep in the same way, so the same spinlock stalls.

That is what led me to the pattern, but it is a separate case and **this patch does not
address it**: the filter gates on `d.arch`, a CUDA compute capability, which never applies to
a HIP device. No HIP code is involved here and nothing in this change depends on that port.

### Why exclude rather than let it fail

The failure mode is a hang, not a failed assertion. A test that hangs takes out the whole
suite rather than reporting one failure, which is what makes it worth filtering up front.

### Notes

* **No changelog fragment.** `changelog/README.md` lists test-only changes under internal
  maintenance. Happy to add one if you would rather have it.
* Non-spinning CAS coverage for older architectures could be added separately.
* Commits are signed off (DCO).

## Summary by CodeRabbit

* **Tests**
  * Atomic compare-and-swap spinlock tests are now skipped on CUDA devices below compute capability `sm_70`.
  * These tests continue to run on CPU and `sm_70+` CUDA devices.


shi-eric:
The `sm_70` filter looks right to me, and it still leaves the CPU variants enabled. Could you clean up the explanation before we merge this?

The current comment describes every test variant as a contended GPU spinlock. That isn't true for CPU, where Warp executes these launches serially. It also says the pre-Volta kernel will deadlock, while NVIDIA's documentation says this pattern can deadlock depending on warp scheduling. The "zero CPU" wording is confusing too: the CPU backend isn't stuck; the host is waiting on a GPU kernel that has stopped making progress.

I'd replace it with:

```python
# Pre-Volta CUDA does not provide independent thread scheduling, so a
# spinlock contended by lanes in the same warp can deadlock. CPU launches
# remain covered because Warp executes these kernels serially by default.
devices = [d for d in get_test_devices() if not d.is_cuda or d.arch >= 70]
```

Could you also update the PR description to separate the ROCm failure you reproduced from the documented pre-Volta CUDA risk and link to NVIDIA's [Independent Thread Scheduling documentation](https://docs.nvidia.com/cuda/cuda-programming-guide/03-advanced/advanced-kernel-programming.html#independent-thread-scheduling)?

I don't think we need a pre-Volta hardware reproduction to justify this guard. We can add non-spinning CAS coverage separately.

shi-eric:
Thanks @zhihuidu-amd as always, please ensure the branch is updated with `main` and your commits are squashed down to a single one.

zhihuidu-amd:
Rebased on `main` and squashed to a single commit as requested.

PR #1959, opened 2026-09-14
Unwind capture state when cudaStreamEndCapture fails

### LINKED ISSUE #1957: Failed cudaStreamEndCapture skips clean_up(), disabling graph capture for the process
### Summary

`wp_cuda_graph_end_capture()` in `warp/native/warp.cu` defines a `clean_up()` lambda that
unwinds the capture bookkeeping: the `g_captures` entry, the graph allocation table, and the
terminating `EndCapture`. Every failure return below that definition calls it — except the
`cudaStreamEndCapture()` check itself, which returns `false` directly.

```cpp
// end the capture
if (!check_cuda(cudaStreamEndCapture(cuda_stream, &graph)))
    return false;        // <-- clean_up() never runs
```

### Impact

When that call fails, the capture is never unwound:

* the stream stays marked as capturing,
* the outstanding graph allocations are left with no owner,
* and there is no path back — no later `capture_begin()` on that stream can succeed.

So a single failed `EndCapture` makes the process unusable for graph capture from then on.
The blast radius is process-wide and permanent, which is presumably why every sibling return
already guards against exactly this.

### How this was found

By reading the error-handling paths, not from a user hitting it. I would rather be clear about
that than imply a field report.

On CUDA the path is hard to reach: a failed operation inside a capture generally leaves the
capture recoverable, as `test_async` notes in its own comment — "capture can succeed despite
some errors during capture (e.g. `cudaInvalidValue` during copy)".

It became visible while porting Warp to a HIP/ROCm backend, where `EndCapture` is stricter
about a capture that has already recorded an error, so the failing branch is actually taken.
The proposed fix is backend-neutral and touches no HIP code — it makes the one return that
skips `clean_up()` behave like every other failure return in the same function.

### Proposed fix

Call `clean_up()` before returning `false`. I have a one-line change ready and will open a pull
request referencing this issue.

### Version

`main` at 6931005e1 ("Harden test subprocesses").

zhihuidu-amd:
Fixes #1957

### What this changes

`wp_cuda_graph_end_capture()` defines a `clean_up()` lambda that unwinds the capture
bookkeeping — the `g_captures` entry, the graph allocation table, and the terminating
`EndCapture`. Every failure return below that definition calls it, except the
`cudaStreamEndCapture()` check itself, which returned `false` directly.

This calls `clean_up()` on that path too, so it behaves like every sibling return in the same
function.

### Why it matters

When `EndCapture` fails and the state is not unwound, the stream stays marked as capturing and
the outstanding graph allocations are left with no owner. There is no path back: no later
`capture_begin()` on that stream can succeed, so the process is unusable for graph capture from
then on.

### Honesty about provenance

This came from reading the error-handling paths, not from a user hitting it — see the issue for
the full reasoning. On CUDA the branch is hard to reach, because a failed operation inside a
capture generally leaves the capture recoverable; `test_async` says as much in its own comment.
It became visible while porting Warp to a HIP/ROCm backend, where `EndCapture` is stricter
about a capture that has already recorded an error.

The change is backend-neutral and contains no HIP code.

### Notes

* Changelog fragment `changelog/1957.fixed.md` included, per `changelog/README.md`.
* Commit is signed off (DCO).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed failures during capture completion that could leave streams incorrectly marked as capturing.
  * Capture state and related resources are now properly cleaned up when capture completion fails.
  * Improved recovery after failed capture completion, helping prevent subsequent operations from being affected by stale capture state or unreleased capture resources.


zhihuidu-amd:
Closing this. The HIP measurement @shi-eric asked for in #1957 refutes the
premise of the PR, and it refutes it on HIP, not only on CUDA.

Short version, full table and method in
https://github.com/NVIDIA/warp/issues/1957#issuecomment-5706516898 :

* The claim this PR rests on — a failed `EndCapture` leaves the stream
  permanently unusable — is false on MI300X/ROCm 7.2 in 4 of the 5 ways I could
  find to invalidate a capture. A fresh capture on the same stream succeeds with
  no code change.
* In the 5th (`hipErrorStreamCaptureUnjoined`, capture still `Active`
  afterwards), `clean_up()`'s second `hipStreamEndCapture` **segfaults**. The
  change makes the only genuinely broken case worse.
* The resource-leak concern in your review is confirmed: the retry `EndCapture`
  returns a non-null graph in every recovering scenario and `clean_up()` never
  destroys it.

So there is no revision of this PR worth making, and it should not sit open. The
narrow HIP behaviour that is left is a ROCm-side question and stays in the fork,
as you suggested.

Thanks for pushing for the measurement rather than taking the reasoning — the
reasoning was wrong.

PR #1969, opened 2026-09-17
Fix Clang CUDA launch bounds

### LINKED ISSUE #1968: @wp.kernel(launch_bounds=...) fails under wp.config.llvm_cuda
### Bug Description

Kernels using `@wp.kernel(launch_bounds=...)` compile and run through NVRTC,
but fail when `wp.config.llvm_cuda` selects the bundled Clang/LLVM CUDA compiler.
I encountered this while evaluating the experimental Clang CUDA path with
kernels using the documented launch-bounds option.

This prevents comparing the two compiler paths on these kernels without
removing the launch-bounds option or working around the missing macro. Removing
the option changes the compilation constraints, so it is not an equivalent
compiler comparison.

Save the following as a Python file and run it with a CUDA-enabled Warp build
that includes `warp-clang`:

```python
import numpy as np
import warp as wp

wp.config.llvm_cuda = True


@wp.kernel(launch_bounds=(64, 1), enable_backward=False, module="unique")
def bounded_double(values: wp.array[float]):
    i = wp.tid()
    values[i] *= 2.0


values = wp.array(np.arange(8, dtype=np.float32), device="cuda:0")
wp.launch(bounded_double, dim=8, inputs=[values], block_dim=64, device="cuda:0")
print(values.numpy())
```

With `wp.config.llvm_cuda = False`, this prints:

```text
[ 0.  2.  4.  6.  8. 10. 12. 14.]
```

With `True`, the first diagnostic is:

```text
error: no template named '__launch_bounds__'; did you mean 'wp::launch_bounds_t'?
extern "C" __launch_bounds__(64, 1) __global__ void ...
```

The generated kernel uses CUDA's `__launch_bounds__` spelling.
`warp/native/cuda_crt.h` provides replacement CUDA attribute macros under the
comment `Attributes otherwise defined by CUDA's crt/host_defines.h`. It defines
`__global__`, `__device__`, and other attributes, but omits `__launch_bounds__`.
Injecting the equivalent definition into the generated translation unit makes
the kernel compile and return the expected values:

```cpp
#define __launch_bounds__(...) __attribute__((launch_bounds(__VA_ARGS__)))
```

The resulting PTX contains `.maxntid 64` and `.minnctapersm 1`. The
single-argument form also compiles with this definition and emits `.maxntid 64`.
Separate fresh kernel caches were used for the backend comparison.

Expected: Clang CUDA compiles the kernel and preserves the requested launch
bounds. Actual: it rejects the entry-point declaration before emitting PTX.

Related: #1026 tracks broader Clang CUDA compatibility work; its linked
analysis does not mention this launch-bounds failure. I can submit a focused
fix in `cuda_crt.h` with a Clang CUDA regression test.

### System Information

- Warp: 1.18.0.dev5, built from source at `1bf652059b9e`
- Bundled LLVM: 22.1.8
- CUDA Toolkit / NVRTC: 13.0; NVIDIA driver: 595.84
- GPU: NVIDIA GeForce RTX 4050 Laptop GPU
- Ubuntu 24.04, Python 3.12.3

and2049:

Kernels using `@wp.kernel(launch_bounds=...)` compile through NVRTC but fail
under `wp.config.llvm_cuda` because `__launch_bounds__` is undefined. I found
this while evaluating the Clang CUDA path; removing the option prevents
comparing the compilers with the same launch-bound constraints.

Closes #1968.

## What does this change?

Define `__launch_bounds__` in the minimal CUDA runtime using CUDA's attribute
expansion. Add Clang CUDA regression coverage for integer and tuple bounds,
checking kernel results and that the tuple form emits `.maxntid 64` and
`.minnctapersm 2` in PTX.

## Existing pull requests

## Validation

- Regression validation: The test fails to compile without the macro and
  passes with it. The PTX assertion also fails when the macro expands to
  nothing or forwards only the first argument. All 9 Clang CUDA tests pass
  with the revised test.
- NVIDIA GPU: GeForce RTX 4050 Laptop GPU on Ubuntu 24.04, CUDA 13.0,
  bundled LLVM 22.1.8.
- Other validation: Pre-commit hooks pass. Before this test-only revision,
  the 9 NVRTC launch-bounds tests and the launch-bounds occupancy test passed;
  a broader targeted run covering CUDA launch options, codegen, and module
  handling reported 480 passed and 8 skipped. The production fix is unchanged.

## Checklist

- [x] I included any required [changelog fragments](https://github.com/NVIDIA/warp/blob/main/changelog/README.md) and used a `+` prefix only for changes that do not require a GitHub issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Clang CUDA compilation failures when using `launch_bounds` with Warp kernels.
  - Added support for integer and tuple forms of `launch_bounds`, including minimum blocks per multiprocessor settings.

- **Tests**
  - Added CUDA coverage verifying kernels with integer and tuple launch bounds compile and execute correctly.
  - Added PTX validation to confirm the configured thread and occupancy limits are emitted correctly.


changelog/1968.fixed.md
```diff
@@ -0,0 +1 @@
+Fix Clang CUDA compilation failures in kernels that set `launch_bounds` in `@wp.kernel`.
```
shi-eric: Minor wording suggestion: showing the decorator syntax directly may make the affected case easier for users to recognize.

```suggestion
Fix CUDA kernel compilation with Warp's Clang backend when using `@wp.kernel(launch_bounds=...)`.
```

Please keep subsequent updates squashed into the existing commit, and rebase onto the latest `main` before merging.

and2049: Thanks for the review, I've applied the changelog suggestions and rebased, with the branch remaining as a single commit.

warp/native/cuda_crt.h
```diff
@@ -10,6 +10,7 @@
 #define __host__ __attribute__((host))
 #define __shared__ __attribute__((shared))
 #define __forceinline__ __attribute__((always_inline))
+#define __launch_bounds__(...) __attribute__((launch_bounds(__VA_ARGS__)))
```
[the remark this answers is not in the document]
shi-eric: @coderabbitai The test already covers the second tuple bound. `test_launch_bounds_kernel` uses `launch_bounds=(64, 2)` and asserts that the Clang-generated PTX contains both `.maxntid 64` and `.minnctapersm 2`. If the macro dropped the second argument, the latter directive would be absent and the test would fail. I ran the targeted Clang CUDA test on a GPU; it passed. Checking PTX is deterministic here, while runtime occupancy depends on the device and kernel resources.

PR #1973, opened 2026-09-20
Annotate tile_matmul entry points CUDA_CALLABLE

zhihuidu-amd:

`tile_matmul()`, `tile_matmul_acc()`, `adj_tile_matmul_acc()` and `adj_tile_matmul()` in
`warp/native/tile_matmul.h` are called only from generated device code, and they call
`WP_TILE_SYNC()` — but none of them carries an execution-space annotation.

They compile today only because NVRTC is invoked with
`--device-as-default-execution-space` (`warp/native/warp.cu:4800`), which makes every
unannotated function implicitly `__device__`. The header's execution space is therefore
decided by a compiler command-line option rather than by the source.

This is inconsistent with the rest of the same file: `wp::index()`, `wp::matmul()` and
`wp::scalar_matmul()` are all already `CUDA_CALLABLE`.

We hit this while building Warp for a non-NVIDIA backend whose runtime compiler does not
accept that NVRTC option. The fix is vendor-neutral, so we are proposing it on its own
rather than as part of any backend work.

## What does this change?

Adds `CUDA_CALLABLE` to those four function definitions. Nothing else.

`CUDA_CALLABLE` expands to `__host__ __device__` under nvcc and to nothing on the CPU
build (`warp/native/builtin.h:16-22`), so every supported configuration compiles and runs
exactly as before. The header becomes self-describing: it no longer depends on the
command-line default for its execution space.

Note the file still has three unannotated device-side helpers (`partition_size`,
`partition_coord`, `partition_load`) that rely on the same option. I left them out to
keep this to the four public entry points, but I am happy to extend the change if you
would prefer the file be consistent throughout.

## Existing pull requests

## Validation

- Regression validation: N/A — this is not a bug fix on CUDA; it is a no-op there by
  construction. The risk being validated is the opposite direction: `__host__ __device__`
  on a function that calls `__syncthreads()` could produce a new nvcc or NVRTC
  diagnostic. It does not — see below.

- NVIDIA GPU: **H100 80GB HBM3**, CUDA 13.1 (V13.1.80), Linux.

- Other validation: A/B against unpatched `main` @ `015e5a178`, two trees built from the
  same source, each with its own `WARP_CACHE_PATH` so neither arm can replay the other's
  cubins (Warp's module hash is derived from generated source, which a native-header edit
  does not change).

  | | base | with this change |
  |---|---|---|
  | distinct nvcc warning/error lines, AOT `warp.so` build | 3 | 3, identical |
  | distinct NVRTC warning/error lines, JIT | 5 | 5, identical |
  | kernels JIT-compiled | 64 | 64 |
  | tile + volume suites | Ran 273, errors=3, skipped=53 | Ran 273, errors=3, skipped=53 |

  Both stages are checked because `builtin.h:2862` includes this header, so nvcc *parses*
  it during the AOT build, but the templates are only *instantiated* by NVRTC at JIT time
  — a diagnostic could appear at either.

  Modules run: `test_tile`, `test_tile_matmul_no_mathdx`, `test_tile_matmul_strides`,
  `test_tile_fused_ops`, `test_tile_reduce`, `test_volume`, `test_volume_write`,
  `test_volume_validation`.

  The 3 errors are identical in both arms and are an artifact of this environment, not of
  the change: `tools/packman/packman` is unavailable on the test node, so `warp-clang.so`
  is not built and three `*_cpu` volume tests fail on a missing CPU compiler.

## Checklist

- [ ] Changelog fragment: not included. There is no user-visible change in behavior,
      performance, compatibility or diagnostics on any configuration Warp supports, so
      this reads as internal maintenance under `changelog/README.md`. Happy to add a
      fragment if you disagree.

## Summary by CodeRabbit

- **Improvements**
  - Updated tile matrix multiplication operations to support calls from CUDA device code.
  - Applies to standard, accumulating, and adjoint matrix multiplication operations.
  - Runtime behavior and numerical results remain unchanged.


shi-eric:
This looks good to me, please rebase with the latest `main` and I'll get it merged

shi-eric (approved):

shi-eric:
This is still out of date, btw.

zhihuidu-amd:
Sorry about that — `main` moved between my fetch and my push, so the first attempt landed
one commit behind. Rebased again and force-pushed: head is now `07a9189a7`, parented
directly on `beb7f297f`, which is the current tip of `main`.

Still a single signed-off commit, and the diff is unchanged by the rebase — I diffed
`git diff upstream/main...HEAD` before and after to confirm — so it is still
`warp/native/tile_matmul.h`, +4/-4.

One thing worth flagging so it does not stall: the force-push reset the review state, so
this reads `REVIEW_REQUIRED` even though your approval is still listed on the PR. It will
need a re-approval before it can merge. Happy to rebase again if `main` moves before you
get to it.

PR #1974, opened 2026-09-20
Cast __ffsll argument in rebuild_find_lowest_on

zhihuidu-amd:

`rebuild_find_lowest_on()` in `warp/native/volume_builder.h:250` passes a `uint64_t`
straight to `__ffsll()`:

```c
return uint32_t(__ffsll(word) - 1);
```

`uint64_t` matches no declared `__ffsll` overload exactly. The call therefore compiles
only as long as exactly one integer overload is visible and an implicit conversion is
available — it is a dependency on the surrounding toolchain's declarations rather than on
anything in this file.

We hit this building Warp for a non-NVIDIA backend whose headers declare more than one
overload, which makes the call ambiguous. The fix is vendor-neutral, so we are proposing
it on its own rather than as part of any backend work.

## What does this change?

Makes the conversion explicit:

```c
return uint32_t(__ffsll(static_cast<unsigned long long>(word)) - 1);
```

`static_cast<unsigned long long>` is a no-op on every 64-bit target and generates
identical code. Two lines of comment explain why the cast is there so it does not look
redundant later.

## Existing pull requests

## Validation

- Regression validation: N/A on CUDA — the generated code is unchanged. The property
  being validated is that the cast introduces no new diagnostic and does not alter volume
  results.

- NVIDIA GPU: **H100 80GB HBM3**, CUDA 13.1 (V13.1.80), Linux.

- Other validation: A/B against unpatched `main` @ `015e5a178`, two trees built from the
  same source, each with its own `WARP_CACHE_PATH` so neither arm can replay the other's
  cubins. `warp.so` md5 differed between arms, confirming both were genuinely rebuilt.

  | | base | with this change |
  |---|---|---|
  | distinct nvcc warning/error lines, AOT `warp.so` build | 3 | 3, identical |
  | distinct NVRTC warning/error lines, JIT | 5 | 5, identical |
  | volume + tile suites | Ran 273, errors=3, skipped=53 | Ran 273, errors=3, skipped=53 |

  Modules run include `test_volume`, `test_volume_write` and `test_volume_validation`,
  which exercise the NanoVDB rebuild path this function belongs to.

  The 3 errors are identical in both arms and are an artifact of this environment, not of
  the change: `tools/packman/packman` is unavailable on the test node, so `warp-clang.so`
  is not built and three `*_cpu` volume tests fail on a missing CPU compiler. Note the
  `#if defined(__CUDA_ARCH__)` branch this change touches is the device branch; the CPU
  branch below it is a portable loop and is untouched.

## Checklist

- [ ] Changelog fragment: not included. No user-visible change in behavior, performance,
      compatibility or diagnostics on any configuration Warp supports. Happy to add one
      if you disagree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA build compatibility when identifying the lowest set bit in 64-bit values.


warp/native/volume_builder.h
```diff
@@ -247,7 +247,9 @@ CUDA_CALLABLE inline uint32_t rebuild_find_lowest_on(uint64_t word)
         return 64u;
     }
 #if defined(__CUDA_ARCH__)
-    return uint32_t(__ffsll(word) - 1);
+    // Cast explicitly: uint64_t matches no __ffsll overload exactly, so the
+    // bare call relies on exactly one integer overload being visible.
+    return uint32_t(__ffsll(static_cast<unsigned long long>(word)) - 1);
```
shi-eric, line 252: For the HIP build that exposed this overload ambiguity, could you add the exact ROCm/HIP SDK and HIP compiler versions (for example, from `hipcc --version`), plus the host compiler version if relevant, to the pull request description? This is for the PR record; no version details are needed in source comments. I also noticed two Warp-owned `uint64_t` calls to `__clzll` without explicit casts: `rebuild_find_highest_on()` in this file and `compute_key_deltas()` in `warp/native/bvh.cu`. Did those compile unambiguously with the same HIP toolchain, or is there a reason the cast is needed only for `__ffsll`?

zhihuidu-amd:
Thanks — both done.

**Toolchain versions.** Added to the PR description under "Toolchain that exposed it", and
kept out of the source comments as you asked. The build runs in the
`rocm/dev-ubuntu-24.04:7.2-complete` container on an MI300X (gfx942):

| | |
|---|---|
| ROCm | 7.2.0 |
| HIP | 7.2.26015-fc0010cf6a |
| `hipcc --version` | AMD clang version 22.0.0git (`roc-7.2.0` 26014, LLVM `7b800a19466229b8479a78de19143dc33c3ab9b5`) |
| `g++ --version` | g++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0 |

**On `__clzll`.** Good catch to ask — it is the right question, since if `__clzll` were
equally ambiguous the premise of this PR would be inconsistent. It is not, and the reason
is in the SDK headers rather than in anything Warp does differently at the two call sites.
From `/opt/rocm/include/hip/amd_detail/amd_device_functions.h` on the 7.2 SDK above:

```
:59  __device__ static inline int __clzll(long long int input)
:67  __device__ static inline int __ffsll(unsigned long long int input)
:75  __device__ static inline int __ffsll(long long int input)
```

HIP declares `__ffsll` **twice** and `__clzll` **once**. A `uint64_t` argument is an exact
match for neither `__ffsll` candidate, and the two conversions are equally good, so
overload resolution is ambiguous and the call is a hard error. `__clzll` has a single
candidate, so the same `uint64_t` simply converts to it and the call resolves.

So yes — `rebuild_find_highest_on()` in this file and `compute_key_deltas()` in
`warp/native/bvh.cu` both compile cleanly on this toolchain, and I left them alone
deliberately. To be precise about what that does and does not claim: it means those two
calls are *unambiguous*, not that a `long long` parameter is the ideal spelling for an
unsigned bit pattern. I would rather not widen the PR on that basis — the ambiguity is a
real build break and the signedness is a style point, and only the first one is blocking
anyone.

If you would prefer the two `__clzll` sites cast as well for symmetry, I am happy to add
them here or in a follow-up, whichever you find easier to review.

zhihuidu-amd:
Also rebased this one onto the current tip of `main` (`beb7f297f`) so it is not sitting
stale while you look at it. Head is now `6523b18e7`, still a single signed-off commit and
the diff is unchanged by the rebase — `warp/native/volume_builder.h`, +3/-1.

shi-eric:
@zhihuidu-amd Thanks for checking the HIP headers and toolchain. That answers my questions. We can leave the PR as-is.

One request for future reviews: please lead with the conclusion and the evidence needed to support it, and edit down AI-assisted text before posting. The relevant details in your last reply were helpful, but the extra framing made the answer harder to find.

Issue #1361, opened 2026-04-13
Optimize tiled CUDA launches by reading blockIdx.x directly


adenzler-nvidia:
For tiled CUDA kernel launches, `blockIdx.x` already encodes the tile index and `threadIdx.x` the intra-tile position. The current codegen computes the linear thread index and then divides by `block_dim()` to recover the tile coordinate, which is redundant on the GPU side.

A dedicated codegen path for tiled CUDA kernels could read `blockIdx.x` directly instead of performing the `linear /= block_dim()` division and the multiply+add to reconstruct the linear index. The CPU path would continue using the current approach since it does not have an equivalent of `blockIdx`.

This came up during review of [GH-1270](https://github.com/NVIDIA/warp/issues/1270) (templated `launch_bounds_t`), where the tiled launch coordinate computation was preserved as-is to limit scope.

aneangel:
Hey @adenzler-nvidia, if this is an open issue and no one is actively working on it can I be assigned this issue?

0z5a:
@aneangel @shi-eric I'd be happy to work on this if no one is already implementing it.

One detail I'd like to preserve is the logical tile-index semantics in both the normal and grid-stride launch paths. In particular, I don't think this should be a mechanical replacement with a fixed blockIdx.x, since a grid-stride block may process more than one logical tile, and the multidimensional launch path also needs to preserve the existing coordinate mapping.

PR #1976, opened 2026-09-20
Recover the folded tile axis from the block index in tiled CUDA launches

0z5a:
## Withdrawn by the author

I am closing this. The screening result contradicts the premise I opened it on, and I
would rather withdraw it than ask a reviewer to spend time on a change that cannot pay
off.

### What I got wrong

I filed this as a codegen simplification: `wp.launch_tiled()` folds the block dimension
into `coord_mult`, and I claimed `wp.tid()` was paying a redundant
`linear /= coord_mult` to undo that fold.

That is not what the helper does. For `coord_mult > 1` it is the algebraic identity

```
(linear / coord_mult) * coord_mult + linear % coord_mult == linear
```

so it computes the coord `launch_coord()` would have produced from the same input, and
the linear index it hands over is the one it received. It relocates the coordinate
computation rather than removing any of it. Both the review bot and my own check agree:
0 mismatches over 9354 (`coord_mult`, thread) pairs across 1D/2D/3D/4D launch shapes.

Paired benchmarking agrees with the algebra. Every ratio lands between 0.9996x and
1.0016x, on both architectures and in every build state — i.e. inside the measurement
spread:

| host | E2E geometric mean | micro geometric mean |
|---|---|---|
| L20 / sm_89 / CUDA 12.8 | 1.0002x | 1.0000x |
| RTX 5090 / sm_120 / CUDA 13.0 | 1.0005x | 0.9999x |
| RTX 5090, clean worktree + clean build | 1.0008x | 1.0000x |

There is no speedup to be had here, so a new native helper plus a macro indirection is
not justified.

### What is in the branch, if any of it is separately useful

- `warp/tests/tile/test_tile_launch_coord.py` — asserts the expected coord from the
  launch shape alone, independently of Warp's own unraveling, and covers 1D/2D/3D tile
  grids, uneven tile counts, `max_blocks` grid-stride, an early return in the loop body,
  `wp.tid()` unpack arity above the launch rank, a tile kernel using no `wp.tid()`, and
  capture/replay. It fails 8/8 against the plausible-but-wrong version that adds the
  lane to `coord.j` (`coord_mult` overflows `shape[1]` and aliases tile rows).
- `docs_tile_launch_coord_measurements.md` — the full measurement record: harnesses,
  method and exclusion policy, per-workload medians, arm spreads, correctness matrices
  on sm_89 and sm_120, and the one grid-stride configuration that needed an exclusively
  held device to be measurable at all.

If the tests are considered worth landing on their own they can be taken from this
branch without the `builtin.h` / `codegen.py` changes; otherwise the whole branch can be
discarded.

Apologies for the noise, and thanks to the reviewers for the speedy catch on the
identity.

Refs #1361

Signed-off-by: 0z5a <Dezhen.lu@student.uni-tuebingen.de>


0z5a:
Withdrawn by the author: `launch_coord_tile` is the algebraic identity `(linear / coord_mult) * coord_mult + linear % coord_mult == linear` for `coord_mult > 1`, and paired screening on sm_89 and sm_120 puts every ratio between 0.9996x and 1.0016x. There is no speedup here, so the helper is not justified. Details and the full measurement record are in the updated description and `docs_tile_launch_coord_measurements.md`. Happy to split out just the tests if anyone wants them.

